PR #23, opened 2017-03-20
Upgrade yarn

noisecapella:
None

#### What's this PR do?
Upgrades yarn to latest stable version

The build should pass


noisecapella:
Just updated this with the changes from https://github.com/mitodl/micromasters/pull/2920

aliceriot:
I just realized that the `travis/update_docker_hub.sh` script is using the same tags that we use for micromasters (`mm_web_travis` and `mm_watch_travis`). We should make new tags to use just for bootcamp - I can open a PR to do this. But we should be ok to merge this without doing that.

PR #28, opened 2017-03-23
Add admins for models

### LINKED ISSUE #24: Add Django admin for profile

noisecapella:
Fixes #24

#### What's this PR do?
Adds Django admin interfaces to models

Go to `/admin` and verify they exist. The order audit table should not be modifiable


ecommerce/admin.py
```diff
@@ -0,0 +1,75 @@
...
+    def has_delete_permission(self, request, obj=None):
+        return False
+
+
+class OrderAdmin(admin.ModelAdmin):
+    """Admin for Order"""
+    model = Order
+
+    readonly_fields = [name for name in get_field_names(Order) if name != 'status']
+
```
pdpinch: I've found it more useful to break out the fields in the list view, instead of having them in one string. It would also be useful to filter on status. Something like:

```
    list_display = ('id', 'user', 'status',)
    list_filter = ('status',)
```

It would also save extra steps if we could display some data from the line item here too, but I'm not sure if that's the klass title or the line description.

ecommerce/admin.py
```diff
@@ -0,0 +1,75 @@
...
+)
+from bootcamp.utils import get_field_names
+
+
+class LineAdmin(admin.ModelAdmin):
+    """Admin for Line"""
+    model = Line
+
+    readonly_fields = get_field_names(Line)
+
```
pdpinch: Adding a list display would help. Even better if we could display the klass title

```
    list_display = ('description', 'klass_id', 'price', 'order', )
```

noisecapella: Sure

ecommerce/admin.py
noisecapella: I'll put both just in case

PR #30, opened 2017-03-24
Add klasses app and models

### LINKED ISSUE #29: Add models for bootcamp and klasses

noisecapella:
Fixes #29 

#### What's this PR do?
Adds `Bootcamp` and `Klass` models to store information about them, including the price and recommended installment.

No manual testing


klasses/migrations/0001_initial.py
```diff
@@ -0,0 +1,34 @@
...
+                ('title', models.TextField()),
+            ],
+        ),
+        migrations.CreateModel(
+            name='Klass',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('klass_id', models.IntegerField()),
+                ('price', models.DecimalField(decimal_places=2, max_digits=20)),
+                ('recommended_installment', models.DecimalField(decimal_places=2, max_digits=20)),
```
Ferdi: What is this for ?   I think they want each class to have a somewhat customizable nr of installments.  
```
class_1
     Installment_1 
          deadline:  03/15/2017
          min_amout:  $2000
     Installment_2 
          ...
     etc.
```

noisecapella: Ah, I was thinking that was something they wanted later on. I'll update the schema to add an extra table for that

Ferdi: ok, it was just a fyi .. you can do it in a separate PR. Maybe you need a start_date field ?

noisecapella: I'm not sure if we need a start_date. From @giocalitri's proposed API we know what klasses the user is eligible to pay for. If they are able to pay for a klass we can choose the installment with the earliest deadline which hasn't passed yet, or a later one if they've already paid for the earlier installment.

klasses/models.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class Klass(Model):
+    """
+    A class within a bootcamp
+    """
+    bootcamp = ForeignKey(Bootcamp)
+    klass_id = IntegerField()
+    price = DecimalField(max_digits=20, decimal_places=2)
+
```
Ferdi: We are also building a dashboard for them in redash where ordering by class_start_date would be useful.

noisecapella: Sure, I'll add it and an end date

klasses/models.py
```diff
@@ -0,0 +1,40 @@
+"""Models for klasses"""
+from django.db.models.base import Model
+from django.db.models.fields import (
+    DateTimeField,
+    DecimalField,
+    TextField,
+    IntegerField,
+)
+from django.db.models.fields.related import ForeignKey
```
giocalitri: nit: I am pretty sure that you can simply `from django.db import models` and then access everything from there.
`models. Model`, and all the fields are available under `models`

klasses/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+class Klass(Model):
+    """
+    A class within a bootcamp
+    """
+    bootcamp = ForeignKey(Bootcamp)
+    klass_id = IntegerField()
+    price = DecimalField(max_digits=20, decimal_places=2)
+    start_date = DateTimeField(null=True)
+    end_date = DateTimeField(null=True)
```
giocalitri: there is no title (not sure there should be) 
@Ferdi ?

klasses/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Installment(Model):
+    """
+    A payment installment
```
giocalitri: nit: I would explain what an `Installment` is in practice.

klasses/models.py
```diff
@@ -0,0 +1,40 @@
...
+    title = TextField()
+
+
+class Klass(Model):
+    """
+    A class within a bootcamp
+    """
+    bootcamp = ForeignKey(Bootcamp)
+    klass_id = IntegerField()
+    price = DecimalField(max_digits=20, decimal_places=2)
```
giocalitri: do we really need this? isn't this the sum of all the installments?

klasses/models.py
noisecapella: Sure

klasses/models.py
noisecapella: I'll remove it

klasses/models.py
noisecapella: Sure, but I'm not sure this is the best place to do that. Is there some place I can put a description which Django Admin can use?

klasses/models.py
Ferdi: I think a title would help actually.  We'll display lists of them (maybe in redash) an title helps distinguish between them

klasses/models.py
Ferdi: I'm looking at this from a different perspective maybe (like redash) - I'd keep the price.

giocalitri: If we keep the price we need to keep it in sync with the sum of the installments and this needs to be done either automatically (but I do not how right now) or manually (with the risk of have inconsistent data)

Ferdi: ok, up to you guys.

noisecapella:
I think that should do it

klasses/models.py
```diff
@@ -0,0 +1,56 @@
...
+    def price(self):
+        """
+        Get price, the sum of all installments
+        """
+        return self.installment_set.aggregate(price=models.Sum('amount'))['price']
+
+    def __str__(self):
+        return "Klass {title} of {bootcamp}".format(
+            title=self.title,
+            bootcamp=self.bootcamp,
```
giocalitri: nit: you may want to specify the title here

noisecapella: The title's included in the `Bootcamp` `__str__`

Issue #4, opened 2017-02-27
US1: As a learner, I want to pay for my Bootcamp, the total amount or in a set of installments.


pdpinch:
Installments (i.e. payment price) are recommendations only. We don't need to enforce them. 

The Bootcamp team would like to set the size and number of installments per bootcamp, but to start with there will be 3 installments of $2000 each.

### acceptance criteria
- [x] ecommerce models
- [x] payment view
- [x] order fulfillment callback

pdpinch:
@noisecapella is this blocked now, waiting for the keys from VPF?

noisecapella:
It's blocked from completion, though there's still work to be done implementing the order fulfillment piece of it. That's what I'm working on now

PR #31, opened 2017-03-28
Added payment view

noisecapella:
Part of #4 

#### What's this PR do?
Adds a view for the payment and tweaks the skeleton UI to make use of it. This view creates an unfulfilled `Order` and returns a form which is submitted to CyberSource. This PR does not include the order fulfillment view, that will be in the next PR.

Create a `Installment` with related a `Klass` and `Bootcamp`. Then log in and type in the klass id into the textbox (this is `klass.klass_id`, not `klass.id`) and a total. Click 'Pay' and you should be directed to an error page on Cybersource. If you go into the database you should see an `Order` that has status `CREATED` and a `Line` which has the `klass_id` you typed in.

### DIFF
--- a/bootcamp/urls_test.py
+++ b/bootcamp/urls_test.py
@@ -1,0 +1,14 @@
+"""Tests for URLs"""
+from unittest import TestCase
+
+from django.core.urlresolvers import reverse
+
+
+class URLTests(TestCase):
+    """URL tests"""
+
+    def test_urls(self):
+        """Make sure URLs match with resolved names"""
+        assert reverse('bootcamp-index') == '/'
+        assert reverse('pay') == '/pay/'
+        assert reverse('create-payment') == '/api/v0/payment/'
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -2,16 +2,99 @@
 Functions for ecommerce
 """
 from base64 import b64encode
+from datetime import datetime
+from decimal import Decimal
 import hashlib
 import hmac
 import logging
+import uuid
 
 from django.conf import settings
+from django.db import transaction
+from django.db.models.aggregates import Sum
+import pytz
+from rest_framework.exceptions import ValidationError
+
+from backends.utils import get_social_username
+from ecommerce.models import (
+    Line,
+    Order,
+)
+from klasses.models import Klass
 
 
 ISO_8601_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 log = logging.getLogger(__name__)
 _REFERENCE_NUMBER_PREFIX = 'BOOTCAMP-'
+
+
+def get_total_paid(user, klass_id):
+    """
+    Get total paid for a klass for a user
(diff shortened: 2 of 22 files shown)


ecommerce/views.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        Create an unfulfilled order and return a response for it.
+        """
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        total = Decimal(serializer.data['total'])
+        klass_id = serializer.data['klass_id']
+
+        order = create_unfulfilled_order(self.request.user, klass_id, total)
+        redirect_url = self.request.build_absolute_uri('/')
```
giocalitri: shouldn't `/` be built as a reverse of an url?

noisecapella: Yes

giocalitri:
Functionality is fine.

Ferdi:
I cant resist , is this "Klass" and not "class"  is suppose to be some kind of a joke ? :-)

ecommerce/api.py
```diff
@@ -2,18 +2,101 @@
...
+            A klass id, a class within a bootcamp
+
+    Returns:
+        Decimal: The total amount paid by a user for a klass
+    """
+    return Line.objects.filter(
+        order__status=Order.FULFILLED,
+        order__user=user,
+        klass_id=klass_id,
+    ).aggregate(price=Sum('price'))['price'] or Decimal(0)
```
noisecapella: This spans multiple `Line`s so I figured a separate function made sense here

bootcamp/urls_test.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class URLTests(TestCase):
+    """URL tests"""
+
+    def test_urls(self):
+        """Make sure URLs match with resolved names"""
+        assert reverse('bootcamp-index') == '/'
+        assert reverse('pay') == '/pay/'
+        assert reverse('create-payment') == '/api/v0/payment/'
```
noisecapella: We don't have other tests that say what the URL name resolves to

giocalitri:
> I cant resist , is this "Klass" and not "class" is suppose to be some kind of a joke ? :-)
 I think someone told me it was called like this

ecommerce/api.py
giocalitri: yeah... but a classmethod would be the same, I guess... feel free to leave as is anyway

noisecapella:
@Ferdi It's the term they use in their database. I guess they used `klass` because `class` is often a reserved keyword

noisecapella:
Comments addressed

giocalitri:
I do not see the new code

noisecapella:
Forgot to push. It should be there now

giocalitri:
👍

Ferdi:
@noisecapella ah, ok thanks ! :-)

PR #37, opened 2017-04-04
Use different URLs for receipt and cancel URLs

### LINKED ISSUE #36: Fix redirect URLs
It looks like CyberSource doesn't like the same URL being used for both the order receipt and cancellation pages. We should make some placeholder URLs which are different from each other (ie {base_url}/receipt/ and {base_url}/cancel/)

noisecapella:
Fixes #36 

#### What's this PR do?
Makes different urls by appending `?status={status}` to the receipt and cancellation URLs. CyberSource errors if the URLs are exactly the same

If you have the right CyberSource credentials you should be able to start a purchase and see the CyberSource page asking for your information


giocalitri:
I do not think it makes sense to tests it
👍

Issue #6, opened 2017-02-27
US3 As a learner, I want to know my payment status for my bootcamp, the amount paid / owed.


pdpinch:
![image00](https://cloud.githubusercontent.com/assets/430126/24708950/68cbbf6e-19e6-11e7-9198-ceeddf292c72.png)

PR #38, opened 2017-04-13
User klass REST API

giocalitri:
closes of #6 

#### What's this PR do?
Adds REST API for listing the klasses where the user is allowed to pay and the list of all the payments done

There are 2 endpoint to test (the detail and the list for klass info).
By default the `BootcampAdmissionClient` now returns an authorization for any local klass in the database, so just create a couple of klasses and check that they show up in the rest API


klasses/views.py
```diff
@@ -0,0 +1,76 @@
...
+from backends.edxorg import EdxOrgOAuth2
+from klasses.api import get_user_klasses_info, get_user_klass_info
+from klasses.permissions import CanReadIfSelf
+
+log = logging.getLogger(__name__)
+
+
+class UserKlassList(APIView):
+    """
+    Class based view for user klasse list view.
```
gsidebo: I see a lot of docstrings/comments that use this "klasse" spelling. Guessing that's a typo and it should be "klass"

klasses/views.py
```diff
@@ -0,0 +1,76 @@
...
+    permission_classes = (
+        permissions.IsAuthenticated,
+        CanReadIfSelf,
+    )
+
+    def get(self, request, username, *args, **kargs):  # pylint: disable=unused-argument
+        """
+        Returns information about the payments for the user in all the klasses she is allowed to pay.
+        """
+        user = get_object_or_404(
```
gsidebo: is there a reason we can't just get the user from `request.user`? we do this same redundant `get_object_or_404` call in the `CanReadIfSelf` permission, and i'm pretty sure pulling the user from `request` would only ever return the user's own klass list for this view. same comment for `UserKlassDetail`

klasses/views.py
```diff
@@ -0,0 +1,76 @@
...
+            User,
+            social_auth__uid=username,
+            social_auth__provider=EdxOrgOAuth2.name
+        )
+
+        return Response(get_user_klasses_info(user=user))
+
+
+class UserKlassDetail(APIView):
```
gsidebo: i think you can use `GenericAPIView` ([doc](http://www.django-rest-framework.org/api-guide/generic-views/#genericapiview)) here to automatically fetch the klass associated with the `klass_id` passed into this view. `mail.views.LearnerMailView` in micromasters ([link](https://github.com/mitodl/micromasters/blob/master/mail/views.py#L46)) has an example of this usage. You would specify extra class-level properties, something like this:

```python
    lookup_field = "id"
    lookup_url_kwarg = "klass_id"
    queryset = Klass.objects.all()
```

The klass would then be available in the `get` method like this: `self.get_object()`

klasses/views.py
giocalitri: this is because the username (like in micromasters) is the edX username and not the bootcamp django app username.

gsidebo: got it

klasses/views.py
giocalitri: I have mixed feelings about this: I agree with you on the fact that the code will look cleaner, but the view does not extract and serialize that object retrieved via lookup, so semantically it would not be correct... 
anyway... I will make the change

giocalitri: Actually, I do not deal with the klass object at all in this view, so I prefer to leave as is

klasses/api.py
```diff
@@ -0,0 +1,75 @@
...
+
+    response = []
+    for klass_id in sorted(all_klasses_ids):
+        klass_info = get_user_klass_info(user, klass_id, bootcamp_client)
+        if klass_info:
+            response.append(klass_info)
+    return response
+
+
+def get_user_klass_info(user, klass_id, bootcamp_client=None):
```
giocalitri: ok

klasses/views.py
giocalitri: typo I made everywhere (and that I thought I fixed)

ecommerce/serializers.py
```diff
@@ -1,14 +1,50 @@
...
+    class Meta:
+        model = Line
+        fields = (
+            'order_status',
+            'order_created_on',
+            'order_updated_on',
+            'klass_id',
+            'price',
+            'description',
+        )
```
giocalitri: ok

klasses/views.py
gsidebo: you don't deal with the `Klass` object in this view as it's written, but my suggestion is that you _should_. the view logic right now suggests that the `klass_id` should fetch us a single `Klass` object, and if it doesn't match, we eventually return a 404. implementing the view in the way i suggested would do that automatically via the tools that DRF already gives us. the line to cast `klass_id` as an int (`klass_id = int(klass_id)`) and fetch the `Klass` based on the id in `get_user_klass_info` are unnecessary boilerplate, IMO. using those `lookup_` parameters is a recommended pattern in DRF, and i think this is a good use case for it

klasses/views.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    authentication_classes = (
+        authentication.SessionAuthentication,
+    )
+    permission_classes = (
+        permissions.IsAuthenticated,
+        CanReadIfSelf,
+    )
+
+    def get(self, request, username, *args, **kargs):  # pylint: disable=unused-argument
```
gsidebo: another minor comment: please rename kargs -> kwargs here and on line 60

klasses/views.py
giocalitri: ok, discussing about this is more than the time to refactor it, so I refactored it

giocalitri:
I think I addressed all the comments

PR #50, opened 2017-04-25
Added 404 and 500 pages

### LINKED ISSUE #48: Error pages
I think we just need a 404 and a 500 page. Maybe a 403. 

It could be the same general idea as micromasters

giocalitri:
closes #48 

#### What's this PR do?
Adds 404 and 500 custom pages

in the `docker-compose.yml` set the `DEBUG: 'False'`
for the 404 just go to a fake url, for the 500 maybe modify a view introducing an exception.

There is no work on the UI, given that it has not been done yet.

### DIFF
--- a/bootcamp/urls.py
+++ b/bootcamp/urls.py
@@ -21,3 +21,6 @@ urlpatterns = [
     url('', include('klasses.urls')),
     url(r'^logout/$', auth_views.logout, {'next_page': '/'}),
 ]
+
+handler404 = 'bootcamp.views.page_404'
+handler500 = 'bootcamp.views.page_500'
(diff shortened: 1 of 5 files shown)


bootcamp/urls.py
```diff
@@ -21,3 +21,6 @@
     url('', include('klasses.urls')),
     url(r'^logout/$', auth_views.logout, {'next_page': '/'}),
 ]
+
+handler404 = 'bootcamp.views.page_404'
```
giocalitri: this is used "auto-magically" by django:
see https://docs.djangoproject.com/en/1.11/topics/http/views/#customizing-error-views

PR #51, opened 2017-04-25
Integrated klasses API information with front end

### LINKED ISSUE #41: Display klass payments based on API results
The bootcamp UI currently displays some hardcoded text for single klass that a user can pay for. We want to show the user the actual information from the API about the klasses they have paid for or can pay for.

Acceptance criteria:

- Payment page displays klass information based on API results
    - If user can pay for a single klass, show payment info/prompt for that klass (like [this design](https://projects.invisionapp.com/share/SFBD1T7DW#/screens/229728222))
    - If user can pay for a multiple klasses, show a dropdown with the options, and when one is selected, show payment info/prompt for that klass

This is the UI part of #6

gsidebo:
Closes #41 

#### What's this PR do?
Essentially this hooks together the back end and front end of the application. The `/pay` page is now reflective of the actual klasses that a user can pay for, and the user can switch between available klasses via a dropdown

1. Set up a single bootcamp/klass that a User can pay for. Navigate to `/pay` and make sure that the data is accurate and there is no dropdown to select among many klasses
1. Set up more bootcamps/klasses and navigate to `/pay`. Make sure the dropdown is visible and that the view is updated according to the selection

Probably `static/js/containers/Payment.js`

Submitting this form via the 'Pay' button doesn't work on this branch or on master due to missing CSRF settings

ss 2017-04-25 at 13 45 51

### DIFF
--- a/bootcamp/serializers_test.py
+++ b/bootcamp/serializers_test.py
@@ -1,0 +1,36 @@
+"""Tests for general bootcamp serializing functionality"""
+from unittest.mock import patch, Mock
+from django.test import TestCase
+
+from profiles.factories import UserFactory, ProfileFactory
+from bootcamp.serializers import serialize_maybe_user
+
+
+class SerializeMaybeUserTests(TestCase):
+    """Tests for serialize_maybe_user"""
+    @classmethod
+    def setUpTestData(cls):
+        cls.profile = ProfileFactory.create(name='Full Name')
+
+    def test_serialize_maybe_user(self):
+        """Test that a user is correctly serialized"""
+        with patch('bootcamp.serializers.get_social_username', return_value='abc'):
+            assert serialize_maybe_user(self.profile.user) == {
+                'full_name': self.profile.name,
+                'username': 'abc'
+            }
+
+    def test_serialize_maybe_user_without_profile(self):
+        """Test that a user without a profile is correctly serialized"""
+        user_without_profile = UserFactory.create()
+        with patch('bootcamp.serializers.get_social_username', return_value='abc'):
+            assert serialize_maybe_user(user_without_profile) == {
+                'full_name': None,
+                'username': 'abc'
+            }
+
+    def test_serialize_maybe_user_anonymous(self):
+        """Test that an anonymous user serializes to None"""
+        is_anonymous = Mock(return_value=True)
+        anon_user = Mock(is_anonymous=is_anonymous)
+        assert serialize_maybe_user(anon_user) is None
--- a/static/js/containers/Payment.js
+++ b/static/js/containers/Payment.js
@@ -3,28 +3,67 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import type { Dispatch } from 'redux';
+import _ from 'lodash';
 
 import type {
   UIState,
 } from '../reducers';
 import {
-  setKlassId,
-  setTotal,
+  clearUI,
+  setPaymentAmount,
+  setSelectedKlassIndex
 } from '../actions';
 import { actions } from '../rest';
 import type { RestState } from '../rest';
-import { createForm } from '../util/util';
+import { createForm, isNilOrBlank } from '../util/util';
 
 class Payment extends React.Component {
   props: {
+    dispatch: Dispatch,
     ui: UIState,
     payment: RestState,
-    dispatch: Dispatch,
+    klasses: RestState,
+    selectedKlass: Object,
   };
 
+  fetchAPIdata() {
+    this.fetchKlasses();
+  }
+
+  componentDidMount() {
+    this.fetchAPIdata();
+  }
+
+  componentDidUpdate() {
+    this.fetchAPIdata();
+  }
+
+  componentWillUnmount() {
+    const { dispatch } = this.props;
+    dispatch(clearUI());
+  }
+
+  fetchKlasses() {
+    const { dispatch, klasses } = this.props;
+    if (!klasses.fetchStatus) {
+      dispatch(
+        actions.klasses({username: SETTINGS.user.username})
+      );
+    }
+  }
+
   sendPayment = () => {
-    const { dispatch, ui: { total, klassId } } = this.props;
-    dispatch(actions.payment(total, klassId)).then(result => {
+    const {
+      dispatch,
+      ui: { paymentAmount },
+      selectedKlass
+    } = this.props;
+    dispatch(
+      actions.payment({
+        paymentAmount: paymentAmount,
+        klassId: selectedKlass.klass_id
+      })
+    ).then(result => {
       const { url, payload } = result;
       const form = createForm(url, payload);
       const body = document.querySelector("body");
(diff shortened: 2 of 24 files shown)


giocalitri:
I have only some general comments: 

1.  why the changes to `ecommerce`? It seems to me that they are just variable renames not important in the context of this PR.
2. is `get_social_username` defined in `profiles.api` a copy of the same function in `backends/utils`?
3. The user serializer class in `bootcamp/serializers.py` seems a bit overkilled considering that it just pulls the username and the full name. But if you want to keep it you need to add tests in a specific module.

For the JS part I think it is better if @aliceriot takes a look.

gsidebo:
@giocalitri 

1. I got tripped up by the use of "total" as a name for the amount that the user was paying. I changed it on the front end, and since that is posted in the body of the API request to the `ecommerce` view, (and, IMO, the change was reasonably small), it made sense to perform the renaming there 
1. It is. Thanks for catching that. I'll change
1. After a second look, I agree. I'll leave the file but get rid of the formal Serializer class

static/js/rest.js
```diff
@@ -102,13 +107,19 @@ export const endpoints: Array<Endpoint> = [
   {
     name: 'payment',
     url: '/api/v0/payment/',
-    makeOptions: (total, klassId) => ({
+    fetchOptions: (params) => ({
```
aliceriot: was there a specific reason for this change? personally I think we may want to stick with the rest parameter (for the argument passed to the function returned by `makeAction`) for flexibility but this would work equally well, just wondering what your thinking was here.

static/js/rest.js
```diff
@@ -29,11 +31,11 @@ export type RestState = {
   fetchStatus?: string,
 };
 
-const INITIAL_STATE: RestState = {
+const INITIAL_REST_STATE: RestState = {
   loaded: false,
   processing: false,
 };
-export const makeReducer = (endpoint: Endpoint, initialState: RestState = INITIAL_STATE) => {
+export const makeReducer = (endpoint: Endpoint, initialState: RestState = INITIAL_REST_STATE) => {
```
aliceriot: rename 👍

static/js/rest.js
```diff
@@ -12,8 +13,9 @@ import {
 
 export type Endpoint = {
   name: string,
-  url: string,
-  makeOptions: (...args: any) => Object
+  url?: string,
+  urlTemplate?: (params: ?Object) => string,
```
aliceriot: in micromasters I handled this issue (sometimes we have a static url, other times it has a url parameter like `/api/v0/resource/${id}`) by just doing a check in the `makeFetchFunc` function like:

```js
fetchJSONWithCSRF(_.isFunction(url) ? url(...args) : url);
```

where `url` is the `url` property of the`Endpoint` object and `args` is the rest parameter to the function returned by `makeFetchFunc`. basically if we have a dynamic url we set the `url` property to be a function which expects the same parameters as the function for generating the request body and returns a string, and if the url is not dynamic we set the `url` property to just a string. Sometimes polymorphism is fun! haha. Anyway, either approach works fine I think.

static/js/containers/Payment.js
```diff
@@ -70,12 +124,38 @@ class Payment extends React.Component {
       <a href="#">Print your statement</a>
     </div>;
   }
+
+  render() {
+    const { selectedKlass } = this.props;
+
+    let renderedKlassDropdown, renderedSelectedKlass;
+    if (this.shouldShowKlassDropdown()) {
+      renderedKlassDropdown = this.renderKlassDropdown();
```
aliceriot: why not just inline this (and the check right below) as ternarys in the JSX below?

static/js/containers/Payment.js
```diff
@@ -70,12 +124,38 @@ class Payment extends React.Component {
...
+    }
+    if (selectedKlass) {
+      renderedSelectedKlass = this.renderSelectedKlass();
+    }
+
+    return <div className="payment">
+      <h3 className="intro">{SETTINGS.user.full_name}, Welcome to MIT Bootcamps</h3>
```
aliceriot: it seems like `SETTINGS.user.full_name` can be empty:

empty_thing

not sure how you want to handle that, one suggestion:

```js
formatUserGreeting = () => (
  `${SETTINGS.user.full_name}, Welcome to MIT Bootcamps`.replace(/^,\s/, "")
);
```

static/js/rest.js
```diff
@@ -29,11 +31,11 @@ export type RestState = {
   fetchStatus?: string,
 };
 
-const INITIAL_STATE: RestState = {
+const INITIAL_REST_STATE: RestState = {
```
aliceriot: One thing I did in MM which you may want to do here is making the `RestState` type into a parameterized type, something like:

```js
type RestState<T> = {
  data: T,
  loaded: boolean,
  ...
};
```

makes Flow a little more useful when you're pulling the data out of the store later!

static/js/containers/Payment_test.js
```diff
@@ -1,90 +1,186 @@
...
 
-  it('sets a price', () => {
-    let payment = renderPayment();
-    payment.find('input[id="total"]').props().onChange({
-      target: {
-        value: "123"
-      }
+  it('sets a selected klass', () => {
+    let fakeKlasses = generateFakeKlasses(3);
+    klassesStub = fetchStub.withArgs(klassesUrl)
```
aliceriot: can you move this setup (with `klassesStub`) into the `beforeEach` function? seems like the same setup is happening in a few functions.

static/js/containers/Payment_test.js
```diff
@@ -1,90 +1,186 @@
...
+        .returns(Promise.resolve(fakeKlasses));
+
+      return renderFullPaymentPage().then((wrapper) => {
+        let title = wrapper.find(klassTitleSelector);
+        assert.equal(title.text(), fakeKlasses[0].klass_name);
+      });
+    });
+
+    it('shows a dropdown if multiple klasses are available', () => {
+      for (let numKlasses of [1, 3]) {
```
aliceriot: this may be personal preference, but for something like this (a sort of data-driven test) I prefer to fully parameterize it, so  something like:

```js
[
  [1, false],
  [2, true],
].forEach((numKlasses, shouldShowDropdown) => {
  // body of test
});
```

instead of computing `shouldShowDropdown` inside the test. imo it's more readable, up to you though!

static/js/containers/Payment_test.js
```diff
@@ -1,90 +1,186 @@
...
-import { makeRequestActionType, makeReceiveSuccessActionType } from '../rest';
+import {
+  makeRequestActionType,
+  makeReceiveSuccessActionType
+} from '../rest';
 
 const REQUEST_PAYMENT = makeRequestActionType('payment');
 const RECEIVE_PAYMENT_SUCCESS = makeReceiveSuccessActionType('payment');
+const REQUEST_KLASSES = makeRequestActionType('klasses');
+const RECEIVE_KLASSES_SUCCESS = makeReceiveSuccessActionType('klasses');
```
aliceriot: something else I did in MM (with the redux-rest stuff) is stick the generated action types on the generated API action dispatcher functions as properties, so that you can do something like `actions.automaticEmails.get.successType`. We may want to do something similar over in bootcamp, but we don't have to right now I think.

gsidebo:
@aliceriot Couple things:

1. If you need it, here's some code that will generate a bootcamp/klass that will show up on the `/pay` page 

    ```python
    from decimal import Decimal
    from django.contrib.auth.models import User
    from klasses.models import Bootcamp, Klass, Installment
    from ecommerce.models import Order, Line, Receipt
    #
    user = User.objects.get(username=YOUR_USERNAME)
    bootcamp = Bootcamp.objects.create(title='Bootcamp 1')
    klass1 = Klass.objects.create(bootcamp=bootcamp, title='Bootcamp 1 Klass 1', klass_id=1)
    Installment.objects.create(klass=klass1, installment_number=1, amount=Decimal(1000))
    ```
1. Just FYI I made an update to the klass dropdown behavior (defaults to nothing selected)

bootcamp/serializers_test.py
```diff
@@ -0,0 +1,36 @@
+"""Tests for general bootcamp serializing functionality"""
+from unittest.mock import patch, Mock
+from django.test import TestCase
```
giocalitri: I am not say that you should rewrite these tests now, but didn't we say that new modules should be `pytest` style?

giocalitri:
I think this PR is incomplete: the Rob's mockup has the list of the aggregated payments at the bottom of the page
bootcamp4 2x

gsidebo:
@giocalitri 

Re: pytest, I stuck to the 'tradition' here due to the large amount of work elsewhere in the PR, but you're right. If you're comfortable with re-reviewing, I'll change that test module. It should be easy enough

Re: payment history, the details/design for that were not finished when I began working on this, and the PR is probably large enough as it is. I created a separate issue to implement this: https://github.com/mitodl/bootcamp-ecommerce/issues/58

giocalitri:
@gsidebo 
for the tests, like I wrote in the comment, I think it is ok to leave as is, but FYI I already wrote a bunch of new modules with the pytest style. The ones that are not in the pytest style are copy/paste from micromasters.

Ok for the other comment.

static/js/rest.js
gsidebo: are you referring to the switch from individual named parameters to `params` (which is expected to be an object)? i did this b/c with the addition of parametrized urls, we have 2 different properties of `Endpoint` that can expect parameters (the `fetchOptions` function and the `urlTemplate`). here are example usages of the actions:

```javascript
dispatch(
  actions.payment({
    paymentAmount: paymentAmount,
    klassId: selectedKlass.klass_id
  })
)

dispatch(
  actions.klasses({username: SETTINGS.user.username})
)
```

`actions.klasses` uses `username` to apply to the url template, and `actions.payment` uses `paymentAmount` and `klassId` to pass to the `fetch...` function. just using sequential arguments and a splat operator wouldn't work for usage in those 2 different places. not sure what you mean by "the rest parameter". open to further suggestions here

static/js/rest.js
gsidebo: changed!

static/js/rest.js
gsidebo: definitely useful, but we don't even have types declared for the 2 REST actions we've defined, so i'm inclined to implement this in some other PR

static/js/rest.js
```diff
@@ -73,21 +75,24 @@ export const makeReducer = (endpoint: Endpoint, initialState: RestState = INITIA
   };
 };
 
-export function makeFetchFunc(endpoint: Endpoint): (...args: any) => Promise<*> {
-  return (...args) => fetchJSONWithCSRF(endpoint.url, endpoint.makeOptions(...args));
+export function makeFetchFunc(endpoint: Endpoint): (params?: Object) => Promise<*> {
+  return (params = {}) => {
+    let url = endpoint.urlTemplate ? endpoint.urlTemplate(params) : endpoint.url;
+    return fetchJSONWithCSRF(url || '', endpoint.fetchOptions(params));
```
gsidebo: @aliceriot sorry, had to revert back to the `url`/`urlTemplate` approach to make flow happy, and now i did this (`url || ''`) for the same reason. i dont like it, but it gets flow to STFU. let me know if you have any better suggestions

static/js/rest.js
aliceriot: ok, I think I see where you're coming from here. What I did in MM was just have the generated action take a rest parameter (i.e. a function argument of the form `...params`) and pass `...params` into all of the relevant functions, so the `fetchOptions`, the url generating function, etc. I think your approach will work equally well though!

static/js/rest.js
aliceriot: makes sense!

static/js/rest.js
aliceriot: womp womp, why flow why. I'll take a look, but yeah, sometimes you have to do that with flow...

in MM in a few places I had to do something like this:

```js
let el: HTMLElement = (document.querySelector('foo'): any);
```

because flow assigns `HTMLElement | null` as the return type of `querySelector`...

similar level of 'stupid hack' there, haha

gsidebo:
@aliceriot anything else on this PR?

giocalitri:
I think I am ok with it 👍

PR #59, opened 2017-04-26
Added integration for bootcamp admission

### LINKED ISSUE #57: Integrate Bootcamp admission API

giocalitri:
closes #57 

#### What's this PR do?
Adds integration for the Bootcamp Admissions web app

1. set the proper environment variables for the test website
2. create a user with the same email on your edx and the bootcamp admissions portal
3. set up your user as admitted in the bootcamp admissions portal
4. verify that you see the klass to pay in the bootcamp ecommerce


gsidebo:
Can you send along the URL/key for the bootcamp test site? I'm assuming I can also use that URL to to set up my own user

giocalitri:
I am trying to get an account myself: I will let you know tomorrow what to do.

klasses/bootcamp_admissions_client_test.py
```diff
@@ -0,0 +1,159 @@
...
+    Helper function to create a mocked response for the requests.get function
+    """
+    mocked_response = Mock()
+    mocked_response.status_code = status_code
+    mocked_response.json.return_value = json_obj
+    return mocked_response
+
+
+@pytest.fixture()
+def setup_test_data(mocker):
```
gsidebo: same name change comment as the one in `klasses/api_test.py`

klasses/bootcamp_admissions_client_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def create_mocked_get_response(status_code, json_obj):
+    """
+    Helper function to create a mocked response for the requests.get function
+    """
+    mocked_response = Mock()
+    mocked_response.status_code = status_code
+    mocked_response.json.return_value = json_obj
+    return mocked_response
```
gsidebo: i tinkered with this a little bit and came up with a decent way to generalize the mocking of a `requests` call. check out the gist here: https://gist.github.com/gsidebo/5d305c8def9399983242814606162c8e. it includes a patch as well as the raw files post-patch. basically it eliminates a lot of the need for configuring the patch as part of the test case

klasses/api_test.py
```diff
@@ -24,7 +25,7 @@
 
 
 @pytest.fixture()
-def setup_test_data():
+def setup_test_data(mocker):
```
giocalitri: I called `setup_test_data` because the same kind of setup is done in the `setUpTestData` classmethod in the unittest class based tests.
To me it makes more sense to be consistent in this sense.

klasses/api_test.py
```diff
@@ -38,6 +39,27 @@ def setup_test_data():
...
+                        {
+                            "klass_id": klass.klass_id,  # NOTE: this is the ID on the remote web service
+                            "klass_name": klass.title,
+                            "is_user_eligible_to_pay": True
+                        } for klass in bootcamp.klass_set.order_by('id')
+                    ]
+                } for bootcamp in Bootcamp.objects.all().order_by('id')
+            ]
+        }
+    )
```
giocalitri: this is a good idea, but I do not think the proper place for that function is `conftest.py`

klasses/bootcamp_admissions_client.py
```diff
@@ -24,27 +31,69 @@ def _get_admissions(self):
...
 
-        For now it just returns the first of the klasses in the local database for all the users.
-        Eventually this should make requests to the remote app and be fault tolerant
-        """
-        # this is the expected structure according to
-        # https://github.com/mitodl/bootcamp-ecommerce/issues/15
-        return {
-            "user": self.user_email,
-            "bootcamps": [
+        This should return a response like:
```
giocalitri: ok

klasses/bootcamp_admissions_client_test.py
giocalitri: sorry but why is this better than having a parameterized function that returns the object I need? What If I need another content in the JSON response? I do not think this is necessary

klasses/api_test.py
gsidebo: here's another simple pattern to achieve this with an actual fixture: http://stackoverflow.com/questions/33508060/create-and-import-helper-functions-in-tests-without-creating-packages-in-test-di/42156088#42156088

klasses/api_test.py
gsidebo: i don't know what benefit we get from the naming similarity with unittest/xUnit. if we keep this, all fixtures we define and use will be nouns while this one is a verb. i think it's more beneficial to think about fixtures the way pytest does and follow their conventions ([every fixture in the docs has a noun name](https://docs.pytest.org/en/latest/fixture.html)) than to keep a vague reference to our old test framework. anyway i'll leave this up to you. not something to hold up the PR for

klasses/bootcamp_admissions_client_test.py
gsidebo: Not sure I understand. Here are a few benefits of that change as I see them:

1. If you have a mocked request/response that you use multiple times (you use the same 200 request/response in here 3 times), you can define it once and declare it as a fixture for any test case that uses it.
1. If you need to set custom JSON response data/side effects, you can do that. The gist actually has several examples: [1](https://gist.github.com/gsidebo/5d305c8def9399983242814606162c8e#file-1_requests_fixture-patch-L46), [2](https://gist.github.com/gsidebo/5d305c8def9399983242814606162c8e#file-1_requests_fixture-patch-L54), [3](https://gist.github.com/gsidebo/5d305c8def9399983242814606162c8e#file-1_requests_fixture-patch-L127)
1. The patching of `requests` no longer happens in `setup_test_data`, which makes more 
sense since patching `requests` doesn't have anything to do with the test data
1. That fixture can be reused in other test cases in the future

giocalitri:
I addressed your comments, but I had to make a small change to handle real life URLs for the bootcamp admission client: sorry for the additional code change but I put it in a different PR.

gsidebo:
OK sounds good. Let me know when I can review the other one

giocalitri:
sorry: different commit :-) the last one
BTW I have just tested with the real (test) admission service. Not sure if you want to test as well

gsidebo:
Got it. Just reviewed & everything looks good in those recent commits 👍

giocalitri:
Sorry: just saw the EDIT: it was more a follow up after

> I am trying to get an account myself: I will let you know tomorrow what to do.

PR #69, opened 2017-05-02
Fixed redirect from index from cybersource

### LINKED ISSUE #65: Blank page after payment on cybersource
After the payment on cybersource the user gets redirected to the main website `https://bootcamp-ecommerce-ci.herokuapp.com/?status=receipt` but the page is empty.

Tested only on CI environment so far

giocalitri:
fixes #65 

#### What's this PR do?
fixes the behavior of the home page on the POST from cybersource

1. login in the bootcamp app
2. go to `/api/v0/klasses/<username>/`
3. in the browser console type `$.post('/?foo=bar&foo=baz')`

you should be able to see 2 response: first a 301 redirect, then a 200 ok from `/pay/?foo=bar&foo=baz`. Note that the parameters are kept during the redirect.

### DIFF
--- a/bootcamp/views.py
+++ b/bootcamp/views.py
@@ -5,7 +5,8 @@ import json
 
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render, redirect
+from django.shortcuts import render, redirect, reverse
+from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
 from raven.contrib.django.raven_compat.models import client as sentry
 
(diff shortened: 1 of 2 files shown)


bootcamp/views.py
```diff
@@ -23,9 +24,16 @@ def _serialize_js_settings(request):  # pylint: disable=missing-docstring
...
-def index(request):  # pylint: disable=missing-docstring
+@csrf_exempt
+def index(request):
+    """
+    Index page
+    """
     if request.user.is_authenticated():
-        return redirect(to='pay')
+        to_url = reverse('pay')
+        if request.GET:
```
noisecapella: If the user is coming from Cybersource wouldn't this always be a POST?

giocalitri: but the params are passed via GET, aren't they?

giocalitri: I mean, even in the POST request they are passed as URL parameters

PR #73, opened 2017-05-03
Added caching for admissions

### LINKED ISSUE #70: Basic cashing of the user bootcamp admissions
Very basic caching to be used for the User email reminders

giocalitri:
closes #70 

#### What's this PR do?
Adds basic caching for the admissions coming from bootcamp admissions

We need to create at least one admission on the admission system: talk to me for this.


klasses/api.py
```diff
@@ -9,6 +10,8 @@
 from klasses.models import Klass
 from klasses.serializers import InstallmentSerializer
 
+log = logging.getLogger(__name__)
```
giocalitri: yep: I'll remove it

klasses/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+            user=user,
+            klass=klass,
+            defaults=updated_values
+        )
+    # delete anything is not in the current admissions
+    BootcampAdmissionCache.delete_all_but(user, payable_klasses_keys)
+
+
+@async.task
+def async_cache_admissions(user_email, payable_klasses_json):
```
giocalitri: My approach is always put a minimal amount of code inside the task because they are much harder to debug via pdb; plus I always think that some future code might need the same functionality but in a sync way.
Not sure... just my approach :-)

PR #74, opened 2017-05-04
Added payment history component

### LINKED ISSUE #58: Add payment history to payment page
Invision design: https://projects.invisionapp.com/share/SFBD1T7DW#/screens/229943705

The UI design calls for a payment history for each klass at the bottom of the `/pay` page

gsidebo:
Closes #58 

#### What's this PR do?
- Adds the payment history component (detailed [here](https://projects.invisionapp.com/share/SFBD1T7DW#/screens/229943705))
- Changes some html/css to make content areas & backgrounds stretch across the length of the screen (discussed with @roberthouse54)
- Splits the payment component into a payment container + 2 child components (Payment and PaymentHistory)

- Navigate to `/pay` with no previous payments, make sure that the history component doesn't show up
- Navigate to `/pay` with no previous payments, make sure history component is visible, amounts are consistent, UI is correct, etc.

`static/js/components/PaymentHistory.js` and `static/js/containers/PaymentPage.js`

**The diff is a lot larger than the actual changes due to some movement of code.** `static/js/containers/PaymentPage.js` and `static/js/containers/Payment.js` were split from a single js file. `static/js/containers/Payment.js` has no new logic, and the only logic added in `static/js/containers/PaymentPage.js` is for deciding whether or not the `PaymentHistory` component should be rendered, and the code to actually render it.

ss 2017-05-04 at 13 49 42

### DIFF
--- a/static/js/components/PaymentHistory.js
+++ b/static/js/components/PaymentHistory.js
@@ -1,0 +1,60 @@
+// @flow
+/* global SETTINGS: false */
+import React from 'react';
+import R from 'ramda';
+
+import { formatDollarAmount } from '../util/util';
+import type { RestState } from '../rest';
+
+export default class PaymentHistory extends React.Component {
+  props: {
+    klasses: RestState
+  };
+
+  renderPaymentRow = (klass: Object) => {
+    let paymentAmountMsg = klass.total_paid !== klass.price ?
+      `${formatDollarAmount(klass.total_paid)} out of ${formatDollarAmount(klass.price)}` :
+      `${formatDollarAmount(klass.total_paid)}`;
+
+    return <tr key={klass.klass_key}>
+      <td>{klass.klass_name}</td>
+      <td>{paymentAmountMsg}</td>
+      <td className="statement-column">
+        <i className="material-icons">print</i><a href="#" className="statement-link">View Statement</a>
+      </td>
+    </tr>;
+  };
+
+  renderPaymentRows = R.pipe(
+    R.filter(
+      R.compose(
+        R.not,
+        R.equals(0),
+        R.propOr(0, 'total_paid')
+      )
+    ),
+    R.map(this.renderPaymentRow)
+  );
+
+  render() {
+    const { klasses } = this.props;
+
+    return <div className="payment-history-section">
+      <div className="payment-history-block">
+        <h3>Payment History</h3>
+        <table>
+          <thead>
+            <tr>
+              <th className="col1">Bootcamp</th>
+              <th className="col2">Amount Paid</th>
+              <th className="col3" />
+            </tr>
+          </thead>
+          <tbody>
+            {this.renderPaymentRows(klasses.data)}
+          </tbody>
+        </table>
+      </div>
+    </div>;
+  }
+}
--- a/static/js/containers/PaymentPage.js
+++ b/static/js/containers/PaymentPage.js
@@ -1,0 +1,164 @@
+// @flow
+/* global SETTINGS: false */
+import React from 'react';
+import { connect } from 'react-redux';
+import type { Dispatch } from 'redux';
+import _ from 'lodash';
+import R from 'ramda';
+
+import {
+  clearUI,
+  setPaymentAmount,
+  setSelectedKlassIndex
+} from '../actions';
+import { actions } from '../rest';
+import { createForm } from '../util/util';
+import Payment from '../components/Payment';
+import PaymentHistory from '../components/PaymentHistory';
+import type { UIState } from '../reducers';
+import type { RestState } from '../rest';
+import type { InputEvent } from '../flow/events';
+
+class PaymentPage extends React.Component {
+  props: {
+    dispatch: Dispatch,
+    ui: UIState,
+    payment: RestState,
+    klasses: RestState,
+    selectedKlass: Object,
+  };
+
+  fetchAPIdata() {
+    this.fetchKlasses();
+  }
+
+  componentDidMount() {
+    this.fetchAPIdata();
+  }
+
+  componentDidUpdate() {
+    this.fetchAPIdata();
+  }
+
+  componentWillUnmount() {
+    const { dispatch } = this.props;
+    dispatch(clearUI());
+  }
+
+  fetchKlasses() {
+    const { dispatch, klasses } = this.props;
+    if (!klasses.fetchStatus) {
+      dispatch(
+        actions.klasses({username: SETTINGS.user.username})
+      );
+    }
+  }
+
+  sendPayment = () => {
+    const {
+      dispatch,
+      ui: { paymentAmount },
+      selectedKlass
+    } = this.props;
+    if (selectedKlass && paymentAmount) {
+      dispatch(
+        actions.payment({
+          paymentAmount: paymentAmount,
+          klassKey: selectedKlass.klass_key
+        })
+      ).then(result => {
+        const {url, payload} = result;
+        const form = createForm(url, payload);
+        const body = document.querySelector("body");
+        body.appendChild(form);
+        form.submit();
+      });
+    }
+  };
+
+  setPaymentAmount = (event: InputEvent) => {
+    const { dispatch } = this.props;
+    dispatch(setPaymentAmount(event.target.value));
+  };
+
+  setSelectedKlassIndex = (event: InputEvent) => {
+    const { dispatch } = this.props;
+    dispatch(setSelectedKlassIndex(parseInt(event.target.value)));
+  };
+
+  hasPastPayments = () => {
+    const { klasses } = this.props;
+
+    if (!klasses.data) return false;
+
+    let totalPaid = R.compose(
+      R.sum,
+      R.map(
+        R.propOr(0, 'total_paid')
+      )
+    )(klasses.data);
+    return totalPaid > 0;
+  };
+
+  render() {
+    const {
+      ui,
+      payment,
+      klasses,
+      selectedKlass
+    } = this.props;
+
+    let renderedPaymentHistory = this.hasPastPayments()
+      ? (
+        <div className="body-row">
+          <PaymentHistory klasses={klasses} />
+        </div>
+      )
+      : null;
+
+    return <div>
+      <div className="body-row top-content-container">
+        <div className="top-content">
+          <Payment
+            ui={ui}
+            payment={payment}
+            klasses={klasses}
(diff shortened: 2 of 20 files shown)


aliceriot:
When I go to `/pay` I'm just seeing a blank page:

empty_should_it_be

is that what I should be seeing? AFAIK I have no payments on my user.

gsidebo:
@aliceriot you probably don't have any bootcamp data, and it looks like you don't have a profile name either. Try running this after replacing the all-caps variables at the bottom with your relevant values. There might be one or two imports missing. 

```python
from decimal import Decimal
import random
from datetime import datetime, timedelta
from django.db.models import Max, Sum
from django.contrib.auth.models import User
from klasses.models import Bootcamp, Klass, Installment
from ecommerce.models import Order, Line, Receipt

def get_max_klass_key():
    klass_key_max = Klass.objects.aggregate(max_klass_key=Max('klass_key'))['max_klass_key']
    return klass_key_max or 0

def create_bootcamp(user, total_price=1000, num_installments=1, set_deadlines=True):
    now = datetime.now()
    ts = now.isoformat()
    bootcamp = Bootcamp.objects.create(title='Bootcamp'.format(ts))
    max_klass_key = get_max_klass_key()
    klass = Klass.objects.create(
        bootcamp=bootcamp, 
        title='Bootcamp {} Klass 1'.format(bootcamp.id), 
        klass_key=max_klass_key + 1
    )
    installment_price = Decimal(total_price / num_installments)
    for i in range(num_installments):
        deadline = now + timedelta(weeks=i+1) if set_deadlines else None
        Installment.objects.create(klass=klass, installment_number=i, amount=installment_price, deadline=deadline)
    return bootcamp

def create_order_and_lines(user, klass, order_status=Order.FULFILLED, num_lines=1, amt_paid_per_line=0.1):
    total_price = Installment.objects.filter(klass=klass).aggregate(total_price=Sum('amount'))['total_price']
    line_payment = int(total_price * Decimal(amt_paid_per_line))
    payment_amounts = [
        ((line_payment) + random.randint(-(line_payment*amt_paid_per_line), (line_payment*amt_paid_per_line))) 
        for i in range(num_lines)
    ]
    order = Order.objects.create(user=user, status=order_status, total_price_paid=sum(payment_amounts))
    for i in range(num_lines):
        line = Line.objects.create(
            order=order, 
            klass_key=klass.klass_key,
            price=payment_amounts[i], 
            description='Line for Klass [{}] (klass_key: {})'.format(klass.title, klass.klass_key)
        )
    return order

def seed():
    bootcamp = create_bootcamp(staff_user)
    klass = bootcamp.klass_set.first()
    create_order_and_lines(staff_user, klass, num_lines=2)

    bootcamp = create_bootcamp(staff_user, num_installments=2)
    klass = bootcamp.klass_set.first()
    create_order_and_lines(staff_user, klass, num_lines=1)

    bootcamp = create_bootcamp(staff_user, set_deadlines=False)
    klass = bootcamp.klass_set.first()
    create_order_and_lines(staff_user, klass, num_lines=1)

def unseed():
    delete_all(Order)
    delete_all(Bootcamp)

username = YOUR_USERNAME
full_name = YOUR_FULL_NAME
Profile.objects.filter(user__username=username).update(name=full_name)
seed()
```

aliceriot:
@gsidebo I already have a bootcamp, a class, and an installment set up from previous testing. I didn't have a profile, but I just made one, and now I just see this:

hehehehehe

gsidebo:
@aliceriot right now, bootcamps are returned by our API under 2 conditions:

1. The external bootcamp API returned some potential bootcamps that the user can sign up for. We don't have a test bootcamp server yet and we don't have a way to mock results in-app, so there won't be any results of this kind while running locally.
1. The user has an `Order` with the status `fulfilled` and some number of `Lines` with payments toward a `Klass`. 

You probably don't have the `Order`/`Lines` necessary to get the API to return something. If you want to have a few `Orders` like that quickly, the code snippet above will set you up with 3

aliceriot:
Ah ok, I wasn't aware I needed orders. adding them now

aliceriot:
ok! I can see the UI now. The fonts here don't look quite like the mockup, but that might be fine for now:
fonts_n_stuff

other than that the functionality looks good! reading code now.

static/js/components/PaymentHistory.js
```diff
@@ -0,0 +1,62 @@
...
+
+    return R.pipe(
+      R.filter(
+        R.compose(
+          R.not,
+          R.equals(0),
+          R.propOr(0, 'total_paid')
+        )
+      ),
+      R.map(this.renderPaymentRow)
```
aliceriot: It looks like `renderPaymentRow` is a pure function (it does not depend on anything other than it's argument for calculating its return value) so it might be a good idea to pull `renderPaymentRow` and this `R.pipe` call out as module-level functions, so we can avoid re-declaring this function object every time we call `renderPaymentRows`.

gsidebo:
I matched the font-family definitions to the [MIT bootcamp site](http://bootcamp.mit.edu/) (eg: the paragraph that opens with "The MIT Bootcamps are short, ..."). We can get  @roberthouse54 to sign off before we merge

gsidebo:
I also noticed your screenshot doesn't have that little line after the Payment History header (see [here](https://projects.invisionapp.com/share/SFBD1T7DW#/screens/229943705)). Just pushed a possible fix. Let me know if you still don't see it after pulling that change

aliceriot:
Yep - I see it now!

static/js/util/test_utils.js
```diff
@@ -16,3 +18,13 @@ export function createAssertReducerResultState<State>(store: Store, getReducerSt
...
+
+export const generateFakeKlasses = (numKlasses = 1) => {
+  return _.times(numKlasses, (i) => ({
+    klass_name: `Bootcamp 1 Klass ${i}`,
+    klass_key: i + 1,
+    payment_deadline: moment(),
+    total_paid: 0,
+    price: 1000
+  }));
+};
```
aliceriot: might be worthwhile to put these sorts of functions in a `factories` subdirectory (`static/js/factories`) like we do in MM

static/js/util/util.js
```diff
@@ -24,3 +24,7 @@ export function createForm(url: string, payload: Object): HTMLFormElement {
 }
 
 export const isNilOrBlank = R.either(R.isNil, R.isEmpty);
+
+export const formatDollarAmount = (amount: ?number): string => (
+  amount ? `$${amount.toLocaleString()}` : ''
```
aliceriot: I'm not sure we want to use `toLocaleString` without setting an explicit locale here. The `toLocaleString` method on `Number` also has some facilities for formatting currency values automatically, this will take a number value and format it to USD:

```js
let num = 42;
num.toLocaleString('en-US', {style: "currency", currency: "USD"});
// $42.00
```

If we don't want the decimal points in the price there's an option called `maximumFractionDigits` which can be set to 0.

I think we want to set `en-US` explicitly, otherwise we could get some strange stuff that might confuse the user. For instance, setting the locale to `de-DE` gives `42,00 $` for the same value.

static/scss/payment.scss
```diff
@@ -20,10 +24,97 @@
...
+        background-color: $med-gray;
+        margin-left: 15px;
+        height: 1px;
+        width: 70px;
+        position: relative;
+        bottom: 5px;
+      }
+    }
+
+    table {
```
aliceriot: it looks in the JSX above like the `table` and `h3` elements which are being styled here are nested within the `.payment-history-block` selector - any reason why the blocks for these selectors aren't inside the block for `.payment-history-block` here in the CSS?

static/js/containers/PaymentPage.js
```diff
@@ -0,0 +1,162 @@
...
+      payment,
+      klasses,
+      selectedKlass
+    } = this.props;
+
+    let renderedPaymentHistory = this.hasPastPayments() ?
+      <div className="body-row">
+        <PaymentHistory klasses={klasses} />
+      </div> :
+      null;
```
aliceriot: total nit: I have a preference for leading-operator multi-line ternaries, so like this:

```js
bool
  ? ifTrue
  : ifFalse;
```

but this is a totally subjective thing! so feel free to ignore if you think trailing-operator is more readable :)

static/js/components/PaymentHistory.js
gsidebo: nice catch. i _think_ the new implementation accomplishes what youre suggesting here

static/js/util/test_utils.js
gsidebo: done

static/js/util/util.js
gsidebo: done

static/scss/payment.scss
gsidebo: no real reason. i generally like to avoid deep nesting but that's not a really compelling reason in this case. fixed

static/js/containers/PaymentPage.js
gsidebo: makes sense. i wasn't sure about this at first since the true/false cases can be multi-line (as it is here) but parenthesis can help with that. let me know what you think

static/js/util/util.js
```diff
@@ -25,6 +25,11 @@ export function createForm(url: string, payload: Object): HTMLFormElement {
 
 export const isNilOrBlank = R.either(R.isNil, R.isEmpty);
 
-export const formatDollarAmount = (amount: ?number): string => (
-  amount ? `$${amount.toLocaleString()}` : ''
-);
+export const formatDollarAmount = (amount: ?number): string => {
+  amount = (amount === undefined || amount === null) ? 0 : amount;
```
aliceriot: would `amount || 0` suffice here?

static/js/components/PaymentHistory.js
```diff
@@ -25,22 +25,20 @@ export default class PaymentHistory extends React.Component {
...
+  renderPaymentRows = R.pipe(
+    R.filter(
+      R.compose(
+        R.not,
+        R.equals(0),
+        R.propOr(0, 'total_paid')
+      )
+    ),
+    R.map(this.renderPaymentRow)
+  );
```
aliceriot: yep, this looks good!

static/js/util/util.js
gsidebo: :man_facepalming: of course. fixed

aliceriot: :)

PR #77, opened 2017-05-08
Fixed UI to show only payable klasses

### LINKED ISSUE #63: The payment UI should comply with the REST API
The UI shows the option to pay even if the backend returns `"is_user_eligible_to_pay": false,`

screen shot 2017-05-01 at 2 11 04 pm

screen shot 2017-05-01 at 2 11 13 pm

screen shot 2017-05-01 at 2 11 29 pm

gsidebo:
Fixes #63

#### What's this PR do?
Fixes the UI to show only payable klasses. A user will only be able to select/pay for klasses that are set to `is_user_eligible_to_pay=true`

Unfortunately we don't have a test bootcamp server or a straightforward way to return some fake results for that API. You can use [this patch](https://gist.github.com/gsidebo/699c66e0a780e6fa269838db1508a0c0) to get the API to return every `Bootcamp`/`Klass` in your database that doesn't already have an associated `Line`. Those `Klasses` will all be considered payable. 

Make sure to test that a result set with 1 payable `Klass` will not show a dropdown, and a result set with multiple payable `Klasses` will show all of them in the dropdown

`static/js/containers/PaymentPage.js`

- A lot of this diff is related to the variable name changes from `KlassIndex` to `KlassKey` et. al.
- I'm not sure why `static/js/containers/Payment.js` was still hanging around. That file was renamed to `PaymentPage.js` a few PRs ago. I got rid of it as part of this PR


static/js/containers/PaymentPage.js
```diff
@@ -135,30 +132,43 @@ class PaymentPage extends React.Component {
   }
 }
 
-const deriveSelectedKlass = state => {
+const withPayableKlasses = state => {
+  return Object.assign({}, state, {
```
noisecapella: Can you use the spread syntax instead of `Object.assign` here?

static/js/containers/PaymentPage.js
```diff
@@ -135,30 +132,43 @@ class PaymentPage extends React.Component {
...
+  }
+  else if (payableKlassesData.length === 1) {
+    selectedKlass = payableKlassesData[0];
   }
-  return klassesExist && _.isNumber(selectedKlassIndex)
-    ? klasses.data[selectedKlassIndex]
-    : undefined;
+
+  return selectedKlass
+    ? Object.assign({}, state, {selectedKlass: selectedKlass})
```
noisecapella: Can you use the spread syntax here? Ie `{ ...state, selectedKlass }`

PR #83, opened 2017-05-10
Prevent users from making a payment if forbidden from paying for a klass

### LINKED ISSUE #64: The payment API should not allow payments if user is not allowed to pay
This should not be really necessary because the UI eventually will not show the form, but we should do it anyway.

noisecapella:
Fixes #64 

#### What's this PR do?
Adds a validation check if the user is not eligible to pay

If we're already hiding the UI to purchase for non-eligible klasses I don't think we can manually test this

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -24,6 +24,7 @@ from ecommerce.models import (
     Line,
     Order,
 )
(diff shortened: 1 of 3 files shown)


giocalitri:
👍

PR #94, opened 2017-05-15
Removed integer keys from async_cache_admissions task

### LINKED ISSUE #93: JSON error when running async_cache_admissions
I was having problems running `async_cache_admissions.delay(email, klasses_json)` locally to populate the cache. I investigated and it looks like Celery converts all integer keys to strings, because the JSON spec doesn't support any other kind of key. The task was passed this:

    {13: {'is_user_eligible_to_pay': True,
      'klass_id': 13,
      'klass_name': 'Class 2 (Student)',
      'status': 'attending'}}

but it was converted to this by Celery:

    {'13': {'is_user_eligible_to_pay': True, 'klass_id': 13, 'status': 'attending', 'klass_name': 'Class 2 (Student)'}}

We need to change our code to handle this properly

noisecapella:
Fixes #93 

#### What's this PR do?
Changes format of admissions data passed to `async_cache_admissions` to no longer use integer keys

- Delete all `BootcampAdmissionsCache` objects for some user
- Open a shell and run `BootcampAdmissionsClient(email)` where email is the email address of that user
- Check that the task populated `BootcampAdmissionsCache` objects for that user

### DIFF
--- a/klasses/tasks.py
+++ b/klasses/tasks.py
@@ -11,13 +11,13 @@ from klasses.models import Klass, BootcampAdmissionCache
 log = logging.getLogger(__name__)
 
 
-def _cache_admissions(user_email, payable_klasses_json):
+def _cache_admissions(user_email, payable_klasses):
     """
     Caches the admissions for which the user is allowed to pay.
 
     Args:
         user_email (str): the email corresponding to an User
-        payable_klasses_json (dict): a dictionary built by the BootcampAdmissionClient._get_payable_klasses
+        payable_klasses (list): a list built by BootcampAdmissionClient._get_payable_klasses
     """
     try:
         user = User.objects.get(email=user_email)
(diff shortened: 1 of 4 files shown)


klasses/tasks.py
```diff
@@ -11,28 +11,29 @@
...
+        payable_klasses (list): a list built by BootcampAdmissionClient._get_payable_klasses
     """
     try:
         user = User.objects.get(email=user_email)
     except User.DoesNotExist:
         log.exception('User with email %s does not exists', user_email)
         return
 
-    payable_klasses_keys = list(payable_klasses_json.keys())
+    payable_klasses_keys = [klass['klass_id'] for klass in payable_klasses]
```
rhysyngsun: Take it or leave it: could be `payable_klasses_keys = payable_klasses_lookup.keys()` to lessen repetition.

noisecapella: Sure, it will also need to have `list(...)` around it to allow indexing

PR #95, opened 2017-05-16
Removed installment number from the Installment

giocalitri:
N/A

#### What's this PR do?
Removes the installment number: we realized it is not useful

No test required


giocalitri:
@noisecapella can you please review again?

PR #97, opened 2017-05-16
Change installation and payment deadline logic

### LINKED ISSUE #71: Reflect new payment/installment logic in UI
We've decided that the logic for payments and installments in bootcamp-ecommerce is as follows:

- A Klass should have one or more Installments
- The user will always see the total amount they've paid out of the total amount due, and will still be able to pay any amount towards the total at any time before the final Installment deadline
- The user will always see the date of the final deadline (i.e.: the last Installment deadline)
- If a user has not paid as much or more than the total amount up to and including the nearest future Installment amount, they will be shown a message that tells them how much they owe towards that amount, and that it is due by the Installment deadline date. EXAMPLE: A Klass has 4 Installments, $1000 each. 2 of those installments are in the past, the user has paid a total of $2500, and the next Installment has a deadline of May 30th. This message should indicate that the user needs to pay $500 by May 30th.
  - If a user has already paid more than the total amount required for the current Installment, this message will be hidden

We need to reflect this logic in the UI.

gsidebo:
Closes #71

#### What's this PR do?
Changes the payment deadline/installment logic such that we show information about individual installment deadlines rather than just showing the one that is furthest in the future

Check the `/pay` page and select a `Klass` that has multiple installments. Make sure that you see sensible messages when there is a single `Installment` for the `Klass`, and when there are multiple installments with an expired deadline. In the comments I'll link a script that I've used to seed data

`Payment.js`

A couple extra bits of work I did in this PR:

1. Changed our APIs to return actual numerical values for `Line.price` and `Installment.amount`, and set our default to avoid coercing `Decimal` values to a string. JSON has support for numerical values, so it didn't make sense to me that we leave it to the front-end to cast those values to numbers.
1. Added an multicolumn index for `Installment` on `klass` and `deadline`. Those values should be unique together and we perform sorting on them, so the index made sense 

ss 2017-05-16 at 16 03 32
ss 2017-05-16 at 16 03 21
ss 2017-05-16 at 16 03 15


gsidebo:
Here's a gist with a patch that will fake the external bootcamp API results to return Bootcamps from your own data, and also a script to create some seed data: https://gist.github.com/gsidebo/b7753b741de31fc6e1a0122e9dce3a39

I'm still not totally clear on how to work with the bootcamp test site to create bootcamps/klasses that will be returned by the `BootcampAdmissionClient`, hence the continued faking of the API response

static/js/components/Payment.js
```diff
@@ -19,11 +21,68 @@ export default class Payment extends React.Component {
...
+    } = this.props;
+
+    return R.compose(
+      R.sum,
+      R.map(R.prop('amount')),
+      R.slice(0, nextInstallmentIndex + 1)
+    )(installments);
+  };
+
+  getDeadlineDates = R.map(
```
rhysyngsun: Nit: doesn't need to be a class method.

static/js/components/Payment.js
```diff
@@ -19,11 +21,68 @@ export default class Payment extends React.Component {
...
+
+  getPaymentDeadlineText = (): string => {
+    const { selectedKlass, now } = this.props;
+
+    let installments = selectedKlass.installments;
+    let installmentDeadlineText = '',
+      finalInstallmentDeadline = '';
+
+    if (installments.length > 1) {
+      let deadlineDates = this.getDeadlineDates(installments);
```
rhysyngsun: Suggestion: move this line outside the `if` statement, and then remove the `else` block and also move line 77 outside the if statement, since all that code is doing the same thing.

static/js/components/Payment.js
```diff
@@ -19,11 +21,68 @@ export default class Payment extends React.Component {
...
+      let nextInstallmentIndex = R.findIndex(R.lt(now), deadlineDates);
+      if (nextInstallmentIndex !== installments.length - 1) {
+        let totalOwedForInstallment = this.getTotalOwedUpToInstallment(nextInstallmentIndex);
+        if (selectedKlass.total_paid < totalOwedForInstallment) {
+          installmentDeadlineText =
+            `A total downpayment of ${formatDollarAmount(totalOwedForInstallment)} ` +
+            `is due ${formatReadableDate(deadlineDates[nextInstallmentIndex])}.`;
+        }
+      }
+      if (this.hasMissedPreviousInstallment(nextInstallmentIndex - 1)) {
```
rhysyngsun: Suggestion: have this method parameterized just on `nextInstallmentIndex` and do the index math inside it since the concept of previous is relative to that.

static/js/components/Payment.js
```diff
@@ -19,11 +21,68 @@ export default class Payment extends React.Component {
...
+    const { selectedKlass, now } = this.props;
+
+    let installments = selectedKlass.installments;
+    let installmentDeadlineText = '',
+      finalInstallmentDeadline = '';
+
+    if (installments.length > 1) {
+      let deadlineDates = this.getDeadlineDates(installments);
+      let nextInstallmentIndex = R.findIndex(R.lt(now), deadlineDates);
+      if (nextInstallmentIndex !== installments.length - 1) {
```
rhysyngsun: Very much my personal style so take it or leave it: refactor `installments.length - 1` to `lastInstallmentIndex` so this expression is easily understood.

static/js/containers/PaymentPage.js
```diff
@@ -247,6 +247,7 @@ class PaymentPage extends React.Component {
             payment={payment}
             payableKlassesData={payableKlassesData}
             selectedKlass={selectedKlass}
+            now={moment()}
```
rhysyngsun: Should this be generated in `componentWillMount` instead so we don't suddenly update the message if the day changes?

gsidebo: We use a similar pattern of calculating `now` in the `render` method in MM too (ex: `CourseListCard`). This value is calculated when the component is rendered, and the changing time will not cause a state change that would cause a re-render

PR #99, opened 2017-05-17
Fixed app.json

giocalitri:
N/A

#### What's this PR do?
fixes app.json

check heroku, you should not see an error like this for this PR build

screen shot 2017-05-17 at 11 08 17 am


noisecapella:
There are some missing environment variable errors but that's a separate issue :+1:

PR #115, opened 2017-06-13
Upgraded celery to 4

giocalitri:
N/A

#### What's this PR do?
Upgrades celery to 4

Try to run the mail task and verify that it runs correctly

### DIFF
--- a/bootcamp/celery.py
+++ b/bootcamp/celery.py
@@ -16,19 +16,18 @@ from django.conf import settings  # noqa pylint: disable=wrong-import-position
 
 log = logging.getLogger(__name__)
 
+client = Client(**settings.RAVEN_CONFIG)
 
-class CustomCelery(Celery):
-    """Custom celery class to handle Sentry setup."""
+register_logger_signal(client, loglevel=settings.LOG_LEVEL)
 
-    def on_configure(self):
-        """Automatically register Sentry client for use with Celery tasks."""
-        client = Client(**settings.RAVEN_CONFIG)
-        register_logger_signal(client)
-        register_signal(client)
+# The register_signal function can also take an optional argument
+# `ignore_expected` which causes exception classes specified in Task.throws
+# to be ignored
+register_signal(client, ignore_expected=True)
 
-async = CustomCelery('bootcamp')
+app = Celery('bootcamp')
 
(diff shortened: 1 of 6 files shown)


bootcamp/celery.py
```diff
@@ -16,19 +16,27 @@
 
 log = logging.getLogger(__name__)
 
+client = Client(**settings.RAVEN_CONFIG)
 
-class CustomCelery(Celery):
-    """Custom celery class to handle Sentry setup."""
+# register a custom filter to filter out duplicate logs
+register_logger_signal(client)
```
blarghmatey: Is there a reason for calling this in two places?

bootcamp/celery.py
```diff
@@ -16,19 +16,27 @@
...
-        register_logger_signal(client)
-        register_signal(client)
+# The register_logger_signal function can also take an optional argument
+# `loglevel` which is the level used for the handler created.
+# Defaults to `logging.ERROR`
+register_logger_signal(client, loglevel=logging.INFO)
 
-async = CustomCelery('bootcamp')
+# hook into the Celery error handler
+register_signal(client)
```
blarghmatey: Is there a reason for calling this twice?

bootcamp/celery.py
giocalitri: No, I copied the configuration from Sentry docs.

bootcamp/celery.py
giocalitri: same as before

bootcamp/celery.py
giocalitri: The docs I was mentioning are here:
https://docs.sentry.io/clients/python/integrations/celery/

What I do not understand is how in this way the Sentry client is hooked to the Celery app.
In the other way it is much clearer, but I was getting this pylint error:
```
____________________________________________________________________________________________ [pylint] bootcamp/celery.py ____________________________________________________________________________________________
E: 23, 4: An attribute defined in celery.app.base line 296 hides this method (method-hidden)
```

giocalitri: Ok... I verified that everything works fine with this configuration

blarghmatey:
:+1:

Issue #137, opened 2017-11-15
Associate FluidReview user ids with ecommerce users


pdpinch:
@pdpinch commented on [Wed Nov 15 2017](https://github.com/mitodl/odl-video-service/issues/364)

As a developer, I'd like to be able to make requests for user information from FluidReview via the FluidReview API. However, the FluidReview APIs don't support querying by e-mail address like the old admissions application did. Instead, all requests to the user API must specify a FluidReview user id. 

See http://docs.fluidreview.com/fluidreview/resources/users.html

We can store an external (FluidReview) id on the user or profile? model that we can use for making queries. 

We can get the id one of two ways: either from a webhook executed by FluidReview when a user is accepted into a bootcamp, or by requesting the full list of users via `GET /api/v2/users/`.

noisecapella:
@mbertrand For this issue we should add to `WebhookRequest` all the fields from the schema here: https://github.com/mitodl/bootcamp-ecommerce/issues/136#issuecomment-349000434 except for the date of birth, I'm not sure we need that. The code needs to fail gracefully so that `body` will always contain the webhook data even if parsing failed. We should also add a field `status` to indicate whether we have attempted to parse the data yet and if the parsing went successfully, or if parsing failed. I don't think we should have a `ForeignKey` to user since the user may not exist yet in the system.

mbertrand:
Thanks @noisecapella

mbertrand:
@noisecapella @pdpinch do you think`IntegerField` would be adequate for all the WebhookRequest numeric fields (`submission_id`, `user_id`, `award_id`) or might `BigIntegerField` be necessary?  Not sure what range of values to expect.

Also wondering if there should be validation on the `amount_to_pay` and `award_cost` string values to make sure they can be parsed to valid numbers.

noisecapella:
I can't find any documentation on that and Django's docs say IntegerField should support up to 2 billion, but I think we should assume it won't go past that. If this assumption fails in the future we will have the raw data to reparse with later

noisecapella:
Regarding amount_to_pay and award_cost, there is the possibility that they won't be set at all so we should handle that. FluidReview does allow you to specify the type of metadata so I think we can be reasonably sure that the field won't contain non-numeric input

mbertrand:
I have this working, but needed to modify the OAuth login process to associate users by email instead of username: `social_core.pipeline.social_auth.associate_by_email`.  

[...]

Issue #136, opened 2017-11-15
FluidReview webhook endpoint


pdpinch:
@pdpinch commented on [Wed Nov 15 2017](https://github.com/mitodl/odl-video-service/issues/365)

As an application reviewer, I would like to have users added to bootcamp ecommerce when they are accepted into a bootcamp. 

We can set up FluidReview to call a webhook on bootcamp-ecommerce when a user is accepted into a bootcamp (in FluidReview-speak, when their submission status changes to "accepted"). The call can include the user's email, FluidReview ID, submission (application) id and submission metadata, including price. 

### Acceptance criteria:
- [ ] webhook configured on FluidReview that sends user id, email, bootcamp (submission) and personal price
- [ ] When the webhook is executed, we need to:
  - [ ] create a user if one doesn't already exist
  - [ ] associate the user with a bootcamp klass
  - [ ] store the personal price

noisecapella:
The user id, email and submission id are all available, I'm still investigating where the price information is

pdpinch:
The surveymonkey folks have set up a sample trigger for us at https://mitbootcamps.fluidreview.com/admin/triggers/#edit/127456

They say it the sample payload includes DOB, User email, and cost.

pdpinch:
George, I'm wondering now what, if any, other metadata we can reasonably retrieve, like the title of the bootcamp, title of the klass and klass dates.

noisecapella:
It looks like `award.name` contains the title of the bootcamp (we should include `award.id` as a unique id`

mbertrand:
@pdpinch the acceptance criteria described here seem to be covered by or dependent upon other issues for the most part: 
- create a user if one doesn't already exist -> #137 
- store the personal price -> #139 
- associate the user with a bootcamp klass -> consequence of #137?

Should I disregard these criteria in the PR review for this particular issue?

noisecapella:
We still need to update the webhook itself with the right information

pdpinch:
As @noisecapella says, PR #147 is only a partial solution to this issue.

noisecapella:
Here's what the webhook currently looks like:

    {
    "date_of_birth": "{{ submission.var__155310__BzgWgZZxIk }}",
    "user_email": "{{ user.email }}",
    "amount_to_pay": "{{ submission.field__27006 }}",
    "user_id": "{{ user.id }}",
    "submission_id": "{{ submission.id }}",
     "award_id": "{{ award.id }}",
[...]

PR #141, opened 2017-11-21
Remove BootcampAdmissionCache

noisecapella:
Related to #73 
Preparing for #136 and #137 

#### What's this PR do?
Removes `BootcampAdmissionCache` which is unused at this point. It was used for reminder emails but that functionality has been removed in #129 

Nothing should change


noisecapella:
@mbertrand all uses of `migrate` will include `--noinput` which will skip those questions

PR #143, opened 2017-11-22
Oauth requests for FluidReview API

### LINKED ISSUE #138: FluidReview application authentication
@pdpinch commented on [Wed Nov 15 2017](https://github.com/mitodl/odl-video-service/issues/362)

All FluidReview API requests must be authenticated using OAuth _application_ credentials. 

Documentation for requesting credentials and implementing authentcation at http://docs.fluidreview.com/fluidreview/authentication.html

mbertrand:
Closes #138 

#### What's this PR do?
Creates a FluidReviewAPI class that can be used for making OAuth requests to the FluidReview API.

FluidReview invalidates the refresh token every time it is used, so an OAuthToken model was created to save the latest access and refresh tokens.

- Get a current access, refresh token, client ID & secret online at ../admin/developer
- Set the following in your `.env` file with above values:
  ```
   FLUIDREVIEW_CLIENT_ID=
   FLUIDREVIEW_CLIENT_SECRET=
   FLUIDREVIEW_ACCESS_TOKEN=
   FLUIDREVIEW_REFRESH_TOKEN=
   ```
- start a python shell: `docker-compose run web python manage.py shell`:
   ```python
   from backends.fluidreview import FluidReviewAPI
   fr = FluidReviewAPI()
   fr.get('users').json()
   fr.put('users/94379401/metadata/22977', data={"value": 100.00})
   ```


.env.example
```diff
@@ -13,3 +13,7 @@ MAILGUN_RECIPIENT_OVERRIDE=
 BOOTCAMP_ADMISSION_BASE_URL=
 BOOTCAMP_ADMISSION_KEY=
 BOOTCAMP_ECOMMERCE_BASE_URL=
+FLUIDREVIEW_CLIENT_ID=
```
noisecapella: Can you add these to `app.json` too?

backends/fluidreview.py
```diff
@@ -0,0 +1,136 @@
...
+from urllib.parse import urljoin
+from datetime import timedelta
+from django.conf import settings
+from requests_oauthlib import OAuth2Session
+
+from backends.models import OAuthToken
+from backends.utils import utc_now
+
+API_NAME = 'FluidReview'
+API_BASE_URL = 'https://bootcampmit.fluidreview.com/api/v2/'
```
noisecapella: Can you make this an environment variable and pass it into FluidReviewAPI?

backends/fluidreview.py
```diff
@@ -0,0 +1,136 @@
...
+        Create or update the FluidReview token parameters.
+        Should be automatically called when a new token is required.
+        With the FluidReview API, the refresh token can only be used once and is then invalidated,
+        so it must be saved and updated in the database.
+
+        Args:
+            new_token(dict): New token returned by the FluidReview API
+
+        Returns:
+            (OAuthToken) the saved object
```
noisecapella: The syntax for return values should be: `OAuthToken: the saved object`. Can you change it here and elsewhere in this file?

noisecapella:
I don't think this functionality belongs in `backends`, that's for the social auth edx logic. Currently the admissions logic is in the `klasses` app but maybe it would be best to create a new app just for this, maybe `fluidreview`

backends/models.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db import models
+from django.db.models import TextField, CharField, DateTimeField
+
+
+class OAuthToken(models.Model):
+    """
+    Store OAuth access and refresh tokens for an external API, such as FluidReview,
+    that has short-lived refresh tokens (invalidated whenever used).
+    """
+    api = CharField(max_length=30, unique=True, primary_key=True, null=False)
```
noisecapella: I'm not sure it makes sense to have this model support multiple APIs. The only credentials we will support for the forseeable future is FluidReview credentials. Also, the social auth oauth credentials are stored elsewhere already.

backends/models.py
```diff
@@ -0,0 +1,35 @@
...
+        Return the token in an OAuth2Session-compatible format
+
+        Returns:
+            dict: Token attributes
+        """
+        return {
+            'access_token': self.access_token,
+            'refresh_token': self.refresh_token,
+            'token_type': self.token_type,
+            'expires_in': int((self.expires_on - datetime.datetime.now(tz=pytz.UTC)).total_seconds())
```
noisecapella: Can you use the `utc_now` function you defined in `utils`?

noisecapella:
> FluidReview invalidates the refresh token every time it is used
Do you mean access token? If it invalidated the refresh token how would we get another one?

.env.example
mbertrand: Sure

noisecapella:
Nevermind, I see from the docs that you swap in the old for the new

backends/fluidreview.py
mbertrand: It seems these are unlikely to change (except rarely for the version `v2`), and will always be the same in CI, RC, production?

backends/fluidreview.py
mbertrand: OK

backends/models.py
mbertrand: Will move everything to a `fluidreview` app

backends/models.py
mbertrand: Sure

mbertrand:
FluidReview invalidates both the access token and refresh token whenever the refresh token is used, so that necessitates saving both values in the database (`OAuthToken` model) and updating whenever necessary.

backends/fluidreview.py
noisecapella: The production instance is `mitbootcamps`, I'm not sure what `bootcampmit` is specifically. I think it's possible we could have two in the future, a testing instance and a production one. In any case I think this piece of the URL should be configurable so that we don't need a code change if we do end up needing to change it

mbertrand: OK, makes sense

bootcamp/settings.py
```diff
@@ -456,6 +457,14 @@ def get_var(name, default):
 CYBERSOURCE_REFERENCE_PREFIX = get_var("CYBERSOURCE_REFERENCE_PREFIX", None)
 
 
+# FluidReview
+FLUIDREVIEW_ACCESS_TOKEN = get_var("FLUIDREVIEW_ACCESS_TOKEN", None)
```
noisecapella: If these should be required can you add a validation check for these variables?

mbertrand: There aren't validation checks for any required settings/variables yet - the app starts up without errors when I clear out my `.env` file. Maybe this should be a separate issue/PR?

noisecapella: That's fine, I filed https://github.com/mitodl/bootcamp-ecommerce/issues/146

.env.example
```diff
@@ -13,3 +13,7 @@ MAILGUN_RECIPIENT_OVERRIDE=
...
+FLUIDREVIEW_CLIENT_SECRET=
+FLUIDREVIEW_ACCESS_TOKEN=
+FLUIDREVIEW_REFRESH_TOKEN=
```
noisecapella: Can you add `FLUIDREVIEW_BASE_URL` here too?

fluidreview/api.py
```diff
@@ -0,0 +1,138 @@
...
+        Prepare to make OAuth2 requests to the API
+        """
+        self.session = self.initialize_session()
+
+    def initialize_session(self):
+        """
+        Initialize an OAuth2 session with an auto-refreshing token.
+
+        Returns:
+            (OAuth2Session): an OAuth2 session
```
noisecapella: Can you remove the parentheses here and below for the `Returns` section here and below?

bootcamp/settings.py
```diff
@@ -456,6 +457,14 @@ def get_var(name, default):
...
+FLUIDREVIEW_REFRESH_TOKEN = get_var("FLUIDREVIEW_REFRESH_TOKEN", None)
+FLUIDREVIEW_CLIENT_ID = get_var("FLUIDREVIEW_CLIENT_ID", None)
+FLUIDREVIEW_CLIENT_SECRET = get_var("FLUIDREVIEW_CLIENT_SECRET", None)
+FLUIDREVIEW_BASE_URL = get_var("FLUIDREVIEW_BASE_URL", None)
```
noisecapella: I feel like this should exclude the `api/v2` path for consistency with our other `_BASE_URL`s, for example `BOOTCAMP_ADMISSION_BASE_URL` below. What do you think?

mbertrand: sounds good

.env.example
mbertrand: Sure

fluidreview/api.py
mbertrand: OK

noisecapella:
Actually it looks like you need to fix the trailing comma in app.json, since it's JSON it doesn't allow for that

mbertrand:
Done, is that file for heroku?

noisecapella:
Yes 👍

PR #147, opened 2017-11-30
Handle webhooks from FluidReview

### LINKED ISSUE #136: FluidReview webhook endpoint

noisecapella:
Part of #136 

#### What's this PR do?
Checks the authorization token for fluidreview webhooks, then stores the content in the database

 - Set `FLUIDREVIEW_WEBHOOK_AUTH_TOKEN` to some value in `.env`
 - Run this command: `curl http://bootcamp.mit.local:8099/api/v0/fluidreview_webhook/ -H "Authorization: Basic wrong_token" -X POST --data '{"a": "b"}'`, substituting your token and URL. You should see an error message saying that you don't have permission to do this.
 - Next, run the same command, but replacing your token in the command line. In the Django shell check that there is only one `WebhookRequest` object in the database and verify that its contents match `{"a": "b"}`


fluidreview/views_test.py
```diff
@@ -0,0 +1,37 @@
+"""Tests for FluidReview views"""
+import pytest
+from django.core.urlresolvers import reverse
+from rest_framework import status
+
+from fluidreview.models import WebhookRequest
+
+
+@pytest.mark.parametrize("data", ['', "some data", '{"json": "body"}'])
+def test_webhook(db, data, client, settings):  # pylint: disable=unused-argument
```
mbertrand: Could add `pytestmark = pytest.mark.django_db` before the tests instead of using `db` as a parameter for each test.  But either way works.

fluidreview/views.py
```diff
@@ -0,0 +1,19 @@
...
+class WebhookView(APIView):
+    """
+    Handle webhooks coming from FluidReview
+    """
+    permission_classes = (WebhookPermission,)
+    authentication_classes = ()
+
+    def post(self, request):
+        """Store webhook request for later processing"""
+        WebhookRequest.objects.create(body=request.body)
```
mbertrand: Is it reasonable to expect that the webhook request body should always be in JSON format?  If so, maybe there should be validation for it.  But if there are situations where the data could legitimately be anything then this is fine as is.

fluidreview/models.py
```diff
@@ -33,3 +34,13 @@ def json(self):
             'token_type': self.token_type,
             'expires_in': int((self.expires_on - utc_now()).total_seconds())
         }
+
+
+class WebhookRequest(TimestampedModel):
```
mbertrand: Maybe add this model to the admin page?

tox.ini
```diff
@@ -14,10 +14,21 @@ commands =
...
+passenv =
+    COVERAGE_DIR
+    DATABASE_URL
+    BROKER_URL
+    CELERY_RESULT_BACKEND
+    PORT
+    TRAVIS*
+    CI
 setenv =
+    ELASTICSEARCH_INDEX=testindex
```
mbertrand: What's this for?

noisecapella: It's from micromasters, a copy and paste error

fluidreview/views.py
noisecapella: The webhook content is created using a plain text template. It should usually be JSON but it's possible someone may make a typo or syntax error when editing the template. In a future PR there will be an additional model to store data parsed from the JSON, but even then we should keep the raw text to aid in debugging problems

noisecapella:
Comments addressed

PR #155, opened 2017-12-06
Parse WebhookRequests and import/sync FluidReview users

### LINKED ISSUE #137: Associate FluidReview user ids with ecommerce users

mbertrand:
Closes #137

#### What's this PR do?
- Parse a `WebhookRequest` body into separate fields
  - Added the separate fields
  - Added a status field to track if the parsing was successful or not
- If there is no existing user with the email contained in the `WebhookRequest`, create it.
- Create or update the user profile, populating the new field `fluidreview_id` with the FluidReview user id.
- Added a django command `importfluidusers` to do a batch import of FluidReview users
- Modified the python socialauth settings to associate users by email address.

Also a couple minor enhancements:
- Added `OAuthToken` to admin page
- Moved `python manage.py collectstatic` from `Dockerfile` to `docker-compose.yml`

- Send a webhook request containing valid JSON including an existing FluidReview user_id.  A `User` and `Profile` should be created in bootcamp (if not already there), and the `Profile.fluidreview_id` field should be populated.  The `WebhookRequest`'s fields should be correctly populated and the status should be set to `Succeeded`
- Send a webhook request containing invalid JSON or plain text.  The `WebhookRequest.body` field should be populated and the status should be `Failed`, other fields should be null.
- Run the command `docker-compose run web python manage.py importfluidusers` to create `User` and `Profile` objects for every FluidReview user.
- Create an account in Edx for one of the FluidReview user emails that has already been imported into bootcamp via Webhook or batch import.  Then log in to bootcamp via Edx with that account.  In another browser, log into `../admin/social_django/usersocialauth/` and verify that there is a `UserSocialAuth` object associated with the same user that was created from FluidReview.


mbertrand:
@noisecapella I'm about to make another change to this

fluidreview/api.py
```diff
@@ -139,3 +150,45 @@ def request(self, method, url_suffix, **kwargs):
...
+            if attr in body_json:
+                if attr in ('award_cost', 'amount_to_pay'):
+                    if body_json[attr]:
+                        setattr(webhook, attr, Decimal(body_json[attr]))
+                else:
+                    setattr(webhook, attr, body_json[attr])
+        webhook.status = WebhookParseStatus.SUCCEEDED
+    except (json.JSONDecodeError, InvalidOperation) as exc:
+        webhook.status = WebhookParseStatus.FAILED
+        log.error('Webhook %s body is not valid JSON or has invalid values: %s', webhook.id, str(exc))
```
noisecapella: Can you use `log.exception` instead?

fluidreview/api.py
```diff
@@ -139,3 +150,45 @@ def request(self, method, url_suffix, **kwargs):
...
+    """
+    Attempt to load a WebhookRequest body as JSON and assign its values to other attributes.
+
+    Args:
+        webhook (WebhookRequest): WebhookRequest instance
+
+    """
+    try:
+        body_json = json.loads(smart_text(webhook.body))
+        for attr in get_field_names(WebhookRequest):
```
noisecapella: What if the incoming payload has `id` in it? IMO we should explicitly enumerate the fields we want to set here

fluidreview/signals.py
```diff
@@ -0,0 +1,33 @@
...
+        parse_webhook(instance)
+        if instance.status == WebhookParseStatus.SUCCEEDED:
+            if not Profile.objects.filter(fluidreview_id=instance.user_id):
+                try:
+                    # Get user info from FluidReview API (ensures that webhook user_id is real).
+                    user_info = FluidReviewAPI().get('/users/{}'.format(instance.user_id)).json()
+                    serializer = UserSerializer(data=user_info)
+                    serializer.is_valid(raise_exception=True)
+                    process_user(serializer.data)
+                except (ConnectionError, serializers.ValidationError) as exc:
```
noisecapella: Do we need to log and silence the exception here? Why not just let it raise? Unhandled exceptions will be reported to sentry and will show up in the console logs

fluidreview/signals.py
```diff
@@ -0,0 +1,33 @@
...
+                    log.error('Problem processing a user from WebhookRequest %s: %s', instance.id, str(exc))
```
noisecapella: Can you use `log.exception` instead of `str(exc)` here?

fluidreview/api.py
```diff
@@ -139,3 +150,45 @@ def request(self, method, url_suffix, **kwargs):
...
+def process_user(fluid_user):
+    """
+    Create/update User and Profile model objects based on FluidReview user info
+
+    Args:
+        fluid_user (ReturnDict): Data from a fluidreview.serializers.UserSerializer object
+    """
+    user = User.objects.filter(email=fluid_user['email']).first()
+    if not user:
+        user = User.objects.create(email=fluid_user['email'], username=fluid_user['email'])
```
noisecapella: Can we use `get_or_create` here like you do in the next line?

fluidreview/api.py
```diff
@@ -139,3 +150,45 @@ def request(self, method, url_suffix, **kwargs):
...
+        body_json = json.loads(smart_text(webhook.body))
+        for attr in get_field_names(WebhookRequest):
+            if attr in body_json:
+                if attr in ('award_cost', 'amount_to_pay'):
+                    if body_json[attr]:
+                        setattr(webhook, attr, Decimal(body_json[attr]))
+                else:
+                    setattr(webhook, attr, body_json[attr])
+        webhook.status = WebhookParseStatus.SUCCEEDED
+    except (json.JSONDecodeError, InvalidOperation) as exc:
```
noisecapella: It seems like we want to include all exceptions here, since we would want to set `FAILED` for those cases

fluidreview/management/commands/importfluidusers.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Create user and profile objects for each FluidReview user
+    """
+    def handle(self, *args, **options):
+        """
+        Run the command
+        """
+        api = FluidReviewAPI()
+        url = 'users'
+        while url:
```
noisecapella: Is it possible to abstract out the code as a generator? I'm thinking something like

    for fluid_user in api.list_users():
        serializer = serializers.UserSerializer(data=fluid_user)
        if serializer.is_valid(raise_exception=False):
            process_user(serializer.data)
            ...

fluidreview/views_test.py
```diff
@@ -1,40 +1,106 @@
...
+        'last_name': 'Applicant',
+        'member_of': [],
+        'submissions': [],
+        'teams': [],
+    }
+    mock_api = mocker.patch('fluidreview.signals.FluidReviewAPI')
+    mock_api().get.return_value.json.return_value = user_data
+
+    existing_user = 'old@mit.edx'
+    ProfileFactory(
```
noisecapella: The factories can use double underscores in some cases, like this one:

    ProfileFactory(
        user__email=existing_user,
        fluidreview_id=None if should_update else fluid_id
    )

fluidreview/api.py
mbertrand: It's possible that there might be an existing user with that email but a username that doesn't equal the email (if someone logged in before a fluidreview import). In that case, a get_or_create with email and username will create a new user.  A get_or_create with email only will also require assigning the username and saving if new/created.  So I decided to do a filter on email followed by a create with email and username if nothing is returned.

fluidreview/api.py
mbertrand: will do

mbertrand: PS Actually`get_field_names` filters out autocreated fields like `id`, so the fields returned for `WebhookRequest` are: 
```
['created_on', 'updated_on', 'body', 'status', 'user_email', 'user_id', 'submission_id', 'award_id', 'award_name', 'award_cost', 'amount_to_pay']
```
But `created_on` and `updated_on` probably should not be modified.

fluidreview/api.py
mbertrand: OK

fluidreview/api.py
mbertrand: OK

fluidreview/management/commands/importfluidusers.py
mbertrand: sure

fluidreview/signals.py
mbertrand: Don't need to, I'll remove it

fluidreview/signals.py
mbertrand: OK

fluidreview/views_test.py
mbertrand: thanks

fluidreview/api.py
noisecapella: I think you can specify `username` in `defaults`, like this:

    user, _ = User.objects.get_or_create(email=fluid_user['email'], defaults={'username': fluid_user['email']})

fluidreview/api.py
```diff
@@ -139,3 +155,68 @@ def request(self, method, url_suffix, **kwargs):
...
+            raise FluidReviewException("Missing required field(s)")
+        for att in required_fields + optional_fields:
+            if att in body_json:
+                if att in ('award_cost', 'amount_to_pay'):
+                    if body_json[att]:
+                        setattr(webhook, att, Decimal(body_json[att]))
+                else:
+                    setattr(webhook, att, body_json[att])
+        webhook.status = WebhookParseStatus.SUCCEEDED
+    except Exception:  # pylint: disable=broad-except
```
noisecapella: This can just be `except:`

fluidreview/management/commands/importfluidusers.py
```diff
@@ -0,0 +1,24 @@
...
+        """
+        Run the command
+        """
+        for fluid_user in list_users():
+            serializer = serializers.UserSerializer(data=fluid_user)
+            if serializer.is_valid(raise_exception=False):
+                process_user(serializer.data)
+                self.stdout.write('Processed User & Profile for {}'.format(fluid_user['email']))
+            else:
+                self.stdout.write('Invalid data for {}: {}'.format(fluid_user['email'], serializer.errors))
```
noisecapella: Do you know if invalid data is a common or rare case? Can you keep track of if there are invalid cases and return a non-zero exit code in that case?

fluidreview/api.py
mbertrand: will do, thanks!

fluidreview/signals.py
```diff
@@ -0,0 +1,25 @@
...
+def handle_parse_webhook(sender, instance, created, **kwargs):  # pylint:disable=unused-argument
+    """
+    Extract individual attribute values from a WebhookRequest body; create/update User & Profile if necessary
+    """
+    if created and instance.status == WebhookParseStatus.CREATED:
+        parse_webhook(instance)
+        if instance.status == WebhookParseStatus.SUCCEEDED:
+            if not Profile.objects.filter(fluidreview_id=instance.user_id):
+                # Get user info from FluidReview API (ensures that webhook user_id is real).
+                user_info = FluidReviewAPI().get('/users/{}'.format(instance.user_id)).json()
```
noisecapella: I'm wondering if we should add a `resp.raise_for_status()` in the `FluidReviewAPI.get` method before returning the request. What do you think?

noisecapella: Or maybe the `request` method instead?

fluidreview/management/commands/importfluidusers.py
mbertrand: I haven't yet encountered it in testing with the API, so hopefully it is rare.  Will add a non-zero exit code if it happens.

fluidreview/api.py
mbertrand: OK

fluidreview/signals.py
mbertrand: Sure, I'll add it to the `request` method

noisecapella:
Functionality looks great

PR #161, opened 2017-12-13
Look up admissions in WebhookRequest

### LINKED ISSUE #151: Update BootcampAdmissionClient to look up admissions in WebhookRequest
Blocked by #137 

We should:

 - add a field to `Bootcamp` saying whether we use the legacy admissions system for this bootcamp
 - update `BootcampAdmissionClient` to look up whether the user is admitted in `WebhookRequest` if the bootcamp is not a legacy bootcamp. I'm not sure if we should match up by email address or if there's another way to do it

mbertrand:
Closes #151

#### What's this PR do?
Modifies the `BootcampAdmissionsClient` to get both a user's legacy klasses and FluidReview klasses.

- Create a `Bootcamp` with a `Klass` (`klass_key=72713`) and `Installment`.
- Create another `Bootcamp` with a `Klass` (`klass_key=13`)  and `Installment`.
- Send a `WebhookRequest` to your bootcamp instance at `../api/v0/fluidreview_webhook/` with this data:
```json
{
  "date_of_birth": "",
  "user_email": "staff@example.com",
  "amount_to_pay": "",
  "user_id": 95292895,
  "submission_id": 999999,
  "award_id": 72713,
  "award_cost": "1000",
  "award_name": "TEST CAMP"
}
```
- Log in to bootcamp via edx as `staff@example.com`.
- It should show a select field: `Which bootcamp do you want to pay for?`, with the two bootcamps as options.
- Click each one and verify that it shows the correct information for that bootcamp (dates, amounts paid/owed, titles).
- Make a payment for each (use admin console if necessary to change the `Order` status to `Fulfilled`).
- The 'You have paid $X out of $Y' amounts should now be updated for each bootcamp.

### DIFF
--- a/klasses/bootcamp_admissions_client.py
+++ b/klasses/bootcamp_admissions_client.py
@@ -8,6 +8,8 @@ import requests
 from django.conf import settings
 from rest_framework import status
 
+from fluidreview.constants import WebhookParseStatus
+from fluidreview.models import WebhookRequest
 
 log = logging.getLogger(__name__)
 
@@ -76,6 +78,24 @@ def fetch_legacy_admissions(user_email):
         return {}
 
 
+def fetch_fluidreview_klass_keys(fluid_user_id):
+    """
+    Collect all the unique award ids (== klass ids) from WebhookRequests by user email
+
+    Args:
+        fluid_user_id(int): FluidReview id for a user
+
+    Returns:
+        list: FluidReview award_ids for a user
+    """
+    return list(
+        WebhookRequest.objects.filter(
+            user_id=fluid_user_id,
+            status=WebhookParseStatus.SUCCEEDED
+        ).distinct('award_id').values_list('award_id', flat=True)
+    )
+
+
 def get_legacy_payable_klass_ids(admissions):
     """
     Returns a list of the payable klass ids.
@@ -104,7 +124,8 @@ class BootcampAdmissionClient:
             user (User): A user
         """
         legacy_admissions = fetch_legacy_admissions(user.email)
-        self._klass_keys = get_legacy_payable_klass_ids(legacy_admissions)
+        self._klass_keys = get_legacy_payable_klass_ids(legacy_admissions) + \
+            fetch_fluidreview_klass_keys(user.profile.fluidreview_id)
 
     @property
     def payable_klasses_keys(self):
--- a/klasses/bootcamp_admissions_client_test.py
+++ b/klasses/bootcamp_admissions_client_test.py
@@ -7,11 +7,13 @@ import pytest
 from django.conf import settings
 from rest_framework import status
 
+from fluidreview.constants import WebhookParseStatus
+from fluidreview.factories import WebhookRequestFactory
 from klasses.bootcamp_admissions_client import (
     BootcampAdmissionClient,
     fetch_legacy_admissions,
 )
-from profiles.factories import UserFactory
+from profiles.factories import ProfileFactory
 
 # pylint: disable=missing-docstring,redefined-outer-name,unused-argument
 
@@ -68,7 +70,10 @@ def test_data(db):
             'key': settings.BOOTCAMP_ADMISSION_KEY,
         })
     )
-    return UserFactory.create(email=user_email), url
+    profile = ProfileFactory.create(user__email=user_email, fluidreview_id=9999)
+    payable_klass = 16
+    unpayable_klass = 12
+    return profile.user, url, payable_klass, unpayable_klass
 
 
 @pytest.fixture()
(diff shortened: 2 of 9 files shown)


klasses/bootcamp_admissions_client_test.py
```diff
@@ -140,11 +142,26 @@ def test_json_raises(test_data, mocked_get_200):
...
     assert boot_client.can_pay_klass('foo') is False
+
+
+def test_can_pay_klass_webhook_only(test_data, mocked_requests_get):
+    """
+    Test BootcampAdmissionClient.can_pay_klass when legacy admissions API is unavailable
+    """
+    mocked_requests_get.side_effect = ConnectionError
+    user, _ = test_data
+    WebhookRequestFactory(award_id=19, user_email=user.email, user_id=user.profile.fluidreview_id)
```
noisecapella: Can you put `award_id` in a variable and reference it below as well?

klasses/bootcamp_admissions_client_test.py
```diff
@@ -140,11 +142,26 @@ def test_json_raises(test_data, mocked_get_200):
...
 def test_can_pay_klass(test_data, mocked_get_200):
     """
-    Test BootcampAdmissionClient.can_pay_klass
+    Test BootcampAdmissionClient.can_pay_klass with a mix of legacy and fluidreview klasses
     """
     user, _ = test_data
+    WebhookRequestFactory(award_id=19, user_email=user.email, user_id=user.profile.fluidreview_id)
 
     boot_client = BootcampAdmissionClient(user)
     assert boot_client.can_pay_klass(16) is True
```
noisecapella: Can you make variables for `12` `16` near the sample data and `19` locally?

klasses/bootcamp_admissions_client.py
```diff
@@ -76,6 +77,19 @@ def fetch_legacy_admissions(user_email):
         return {}
 
 
+def fetch_fluidreview_klass_keys(fluid_user_id):
+    """
+    Collect all the unique award ids (== klass ids) from WebhookRequests by user email
+
+    Args:
+        email(str): A user's email address
```
noisecapella: Can you update this argument?

klasses/bootcamp_admissions_client.py
```diff
@@ -76,6 +77,19 @@ def fetch_legacy_admissions(user_email):
...
+
+    Returns:
+        list: FluidReview award_ids for a user
+    """
+    return list(set(WebhookRequest.objects.filter(user_id=fluid_user_id).values_list('award_id', flat=True)))
```
noisecapella: Can you use [https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.distinct](distinct) instead of using `set` here to remove duplicates?

noisecapella:
I got an error where the fetch failed in the `handle_parse_webhook` failed. Do we retry that `process_user` call at any point afterwards in case of failure? It looks like since the `status` is set to success, I'm not sure we have a way to know that the `process_user` call failed other than the sentry error that would result

mbertrand:
The `process_user` is not automatically tried again.  It can be manually triggered by running the `importfluidusers` command.  If you prefer I can change the code to call `process_users` in the parse function directly and set the webhook status to failure if it doesn't work.   Did the failure occur because the user id was nonexistent or there was an issue with the token?

noisecapella:
Yeah moving it into the parse function sounds good to me. The user was nonexistent, probably a credentials problem but I wanted to make sure we handle the error gracefully.

klasses/bootcamp_admissions_client.py
mbertrand: ✅

klasses/bootcamp_admissions_client.py
mbertrand: ✅

klasses/bootcamp_admissions_client_test.py
mbertrand: ✅

klasses/bootcamp_admissions_client_test.py
mbertrand: ✅

klasses/bootcamp_admissions_client.py
```diff
@@ -76,6 +77,21 @@ def fetch_legacy_admissions(user_email):
...
+    Collect all the unique award ids (== klass ids) from WebhookRequests by user email
+
+    Args:
+        fluid_user_id(int): FluidReview id for a user
+
+    Returns:
+        list: FluidReview award_ids for a user
+    """
+    return list(
+        WebhookRequest.objects.filter(user_id=fluid_user_id).distinct('award_id').values_list('award_id', flat=True)
```
noisecapella: You need to also filter on the status of `WebhookRequest`

mbertrand: Added the filter and a test

PR #163, opened 2017-12-14
Update FluidReview with total amounts paid 

### LINKED ISSUE #140: POST payment status to FluidReview
As an admin, I would like to be able to view applicant's payment status in the FluidReview application. 

We can update the applicant's submission metadata with their total amount paid whenever they complete an order.

### LINKED ISSUE #146: Add validation check for required environment variables

mbertrand:
Closes #140
Closes #146

#### What's this PR do?
Submits total amount paid to the appropriate FluidReview submission metadata when a user pays for a klass and the order is fulfilled via the `OrderFulfillmentView`

- Add the following env variable: `FLUIDREVIEW_AMOUNTPAID_ID=22977` (27007 for production).
- Create a `Bootcamp` with a `Klass` (`klass_key=72713`) and `Installment`. Do not check the 'Legacy' box.
- Send a `WebhookRequest` to your bootcamp instance at `../api/v0/fluidreview_webhook/` with this data:
```json
{
  "date_of_birth": "",
  "user_email": "staff@example.com",
  "amount_to_pay": "",
  "user_id": 95292895,
  "submission_id": 4621059,
  "award_id": 72713,
  "award_cost": "1000",
  "award_name": "TEST CAMP"
}
```
- Via the front-end, submit a payment of $25.00 for this klass. 
- Via the admin page, change the `Order` status to 'Fulfilled'
- From a django shell:
  ```python
  from ecommerce.models import Order
  from fluidreview.api import *
  order = Order.objects.get(id=<order_id>)
  post_payment(order)

  api = FluidReviewAPI()
  api.get('submissions/4621059/metadata/').json()
  ```
The result should include: 
```python
{'id': 22977, 'name': 'Amount Paid', 'value': '25.00'},
```


bootcamp/settings.py
```diff
@@ -498,3 +494,13 @@ def get_all_config_keys():
     MIDDLEWARE_CLASSES = (
         'debug_toolbar.middleware.DebugToolbarMiddleware',
     ) + MIDDLEWARE_CLASSES
+
+MANDATORY_SETTINGS = [
```
noisecapella: 👍

ecommerce/models.py
```diff
@@ -76,6 +73,19 @@ def to_dict(self):
...
+        """
+        klass being paid for
+
+        Returns:
+            Klass: klass that order is for.
+
+        """
+        line = self.line_set.first()
+        if not line:
+            return ""
```
noisecapella: I don't think returning a string is the right thing here, `None` would be more appropriate.

ecommerce/views.py
```diff
@@ -114,6 +115,8 @@ def post(self, request, *args, **kwargs):
                     )
         else:
             order.status = Order.FULFILLED
+            post_payment(order)
```
noisecapella: What should happen if `post_payment` fails? It seems like the way it is the order would remain with status `CREATED`. If the `post_payment` step is not essential we should keep the order as `FULFILLED` and log the exception so we get a Sentry report. If it's essential we should mark the order as `FAILED`. However I don't think we should do that since we already have their money at that point.

ecommerce/views_test.py
```diff
@@ -131,6 +131,7 @@ def test_order_fulfilled(self):
         assert order.receipt_set.first().data == data
 
         assert send_email.call_count == 0
+        assert post_payment.call_count == 1
```
noisecapella: Can you assert that the right order was passed into `post_payment`? You can use `post_payment.call_args[0][0]` to get the order object and make sure its id compares successfully

fluidreview/api.py
```diff
@@ -235,3 +236,29 @@ def list_users():
             yield fluid_user
         next_page = urlparse(response['next']).query
         url = 'users?{}'.format(next_page) if next_page else None
+
+
+def post_payment(order):
+    """
+    Update amount paid by a user for a class when an order is fulfilled.
```
noisecapella: Can you add `order` to the docstring?

fluidreview/api.py
```diff
@@ -235,3 +236,29 @@ def list_users():
...
+    total_paid = Line.total_paid_for_klass(order.user, klass.klass_key).get('total') or Decimal('0.00')
+    payment_metadata = {
+        'value': '{:0.2f}'.format(total_paid)
+    }
+    webhook = WebhookRequest.objects.filter(user_id=user.profile.fluidreview_id, award_id=klass.klass_key).last()
+    response = FluidReviewAPI().put(
+        'submissions/{}/metadata/{}/'.format(webhook.submission_id, settings.FLUIDREVIEW_AMOUNTPAID_ID),
+        data=payment_metadata
+    )
+    if response.status_code != 200:
```
noisecapella: Instead of checking the status code can you rewrite like this:

    try:
      response = FluidReviewAPI().put(...)
      response.raise_for_status()
    except Exception as ex:
        raise FluidReviewException(...) from ex

This way you won't need to include `response.json()` in the exception, it will exist in the `HTTPError` exception in the exception chain

fluidreview/api.py
```diff
@@ -235,3 +236,29 @@ def list_users():
...
+        return
+    klass = order.get_klass()
+    user = order.user
+    if not klass or klass.bootcamp.legacy:
+        return
+    total_paid = Line.total_paid_for_klass(order.user, klass.klass_key).get('total') or Decimal('0.00')
+    payment_metadata = {
+        'value': '{:0.2f}'.format(total_paid)
+    }
+    webhook = WebhookRequest.objects.filter(user_id=user.profile.fluidreview_id, award_id=klass.klass_key).last()
```
noisecapella: Can you add an [ordering](https://docs.djangoproject.com/en/2.0/ref/models/options/#django.db.models.Options.ordering) field to `WebhookRequest`? Otherwise `last` isn't guaranteed to be the most recent `WebhookRequest`.

fluidreview/api.py
```diff
@@ -235,3 +236,29 @@ def list_users():
...
+    response = FluidReviewAPI().put(
+        'submissions/{}/metadata/{}/'.format(webhook.submission_id, settings.FLUIDREVIEW_AMOUNTPAID_ID),
```
noisecapella: Can you add a check if `webhook` is `None` before this line?

ecommerce/models.py
mbertrand: ✅

ecommerce/views.py
mbertrand: Changed to catch any exception that occurs during the post_payment function and log it.

fluidreview/api.py
mbertrand: I was thinking it would be okay for an exception to be thrown on the next line if it is None, because this shouldn't happen?  Or is it okay for a non-legacy klass to not have any WebhookRequests when a payment is submitted?

fluidreview/api.py
mbertrand: ✅

fluidreview/api.py
mbertrand: ✅

ecommerce/views_test.py
mbertrand: ✅

fluidreview/api.py
mbertrand: ✅

ecommerce/views.py
```diff
@@ -114,6 +115,11 @@ def post(self, request, *args, **kwargs):
                     )
         else:
             order.status = Order.FULFILLED
+            try:
+                post_payment(order)
+            except FluidReviewException:
```
noisecapella: I think we should make this an `except:` instead, just in case there are other exceptions coming from that function.

fluidreview/api.py
noisecapella: I think I meant to ask for a check for `webhook.submission_id`. If `webhook` was `None` then we would get an exception here which has an obvious source. However if `webhook.submission_id` was none it would be `submissions/None/metadata/{}/` which would be confusing to diagnose

ecommerce/views.py
mbertrand: ✅

fluidreview/api.py
mbertrand: ✅ with test

PR #174, opened 2018-01-03
Use docker overrides for travis and local configuration differences

noisecapella:
None

#### What's this PR do?
Use docker override files to reduce configuration duplication

 - Tests should pass and your local docker environment should be unaffected


docker-compose.yml
```diff
@@ -67,15 +63,14 @@ services:
...
     environment:
-      NODE_ENV: 'development'
+      NODE_ENV: 'production'
+      DOCKER_HOST: ${DOCKER_HOST:-missing}
+      CONTAINER_NAME: 'watch'
     env_file: .env
 
   celery:
-    image: bootcamp_python
+    image: open_discussions_python
```
mbertrand: Is this correct?

noisecapella: No, good catch

noisecapella:
In this pr there should be no change of configuration so theres a bug somewhere. I'll look into it

Issue #184, opened 2018-01-11
Increase logging around webhooks


pdpinch:
The FluidReview web app doesn't log our response to webhooks, so it's very difficult to debug issues. We already worked through a 403 response because of badly-pasted token. Now we're struggling with a 406 response. In both these cases an entry isn't made to the webhook table. 

### acceptance critera
- [ ] raise an exception to Sentry on a 403 response to the webhook endpoint. We want to be notified if there are authentication issues
- [ ] raise an exception to Sentry on a 406 response to the webhook endpoint. 
- [ ] on any failed response, log the POST data. If possible, send it to Sentry as well.

mbertrand:
@pdpinch should there be a separate issue/PR for preventing a 406 response?

pdpinch:
separate issue (#186) & PR please.

PR #187, opened 2018-01-11
Raise an exception if anything goes wrong with posting to WebhookView

mbertrand:
#184 

#### What's this PR do?
If anything goes wrong when POSTing a request to `WebhookView`, an exception will be raised (including info on request body and headers) instead of returning a JSON error response.

- Post a request to `../api/v0/fluidreview_webhook/` with an incorrect value or no value for the `Authentication` header.
- Post a request with an `Accept: text/plain` header.
- An exception should be raised, with the request body & headers included in the exception message.

### DIFF
--- a/fluidreview/views.py
+++ b/fluidreview/views.py
@@ -3,6 +3,7 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from bootcamp.negotiations import IgnoreClientContentNegotiation
+from fluidreview.api import FluidReviewException
 from fluidreview.permissions import WebhookPermission
 from fluidreview.models import WebhookRequest
 
@@ -14,6 +15,12 @@ class WebhookView(APIView):
     permission_classes = (WebhookPermission,)
     authentication_classes = ()
     content_negotiation_class = IgnoreClientContentNegotiation
+
+    def handle_exception(self, exc):
+        """Raise any exception with request info instead of returning response with error status/message"""
+        raise FluidReviewException(
+            "REST Error (%s). BODY: %s, META: %s" % (exc, self.request.body, self.request.META)
+        ) from exc
 
     def post(self, request):
         """Store webhook request for later processing"""
--- a/fluidreview/views_test.py
+++ b/fluidreview/views_test.py
@@ -7,6 +7,7 @@ from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
 from rest_framework import status
 
+from fluidreview.api import FluidReviewException
 from fluidreview.models import WebhookRequest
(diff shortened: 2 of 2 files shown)


fluidreview/views.py
```diff
@@ -13,6 +14,10 @@ class WebhookView(APIView):
     permission_classes = (WebhookPermission,)
     authentication_classes = ()
 
+    def handle_exception(self, exc):
+        """Raise any exception with request info instead of returning response with error status/message"""
+        raise FluidReviewException("REST Error (%s). BODY: %s, META: %s", exc, self.request.body, self.request.META)
```
noisecapella: The `%s` values aren't getting replaced with anything here

fluidreview/views_test.py
```diff
@@ -42,8 +43,24 @@ def test_webhook_fail_auth(client, settings, token_missing, mocker):  # pylint:
...
+    If the accept header contains an unsupported type, raise an exception
+    """
+    mocker.patch('fluidreview.api.FluidReviewAPI')
+    settings.FLUIDREVIEW_WEBHOOK_AUTH_TOKEN = 'xyz'
+    headers = {
+        'HTTP_ACCEPT': 'video/mp4',
+        'HTTP_AUTHORIZATION': 'Basic xyz'
+    }
+    url = reverse('fluidreview-webhook')
+    with pytest.raises(FluidReviewException):
```
noisecapella: Can you assert the content of the exception, at least enough of it to verify that it's the accept header causing the problem and not the auth token?

mbertrand: I'll add it to the test above it (authentication failures), because once the `Accept` headers are ignored this won't throw exceptions anymore.

fluidreview/views.py
mbertrand: ✅

fluidreview/views.py
```diff
@@ -13,6 +14,10 @@ class WebhookView(APIView):
     permission_classes = (WebhookPermission,)
     authentication_classes = ()
 
+    def handle_exception(self, exc):
+        """Raise any exception with request info instead of returning response with error status/message"""
+        raise FluidReviewException("REST Error (%s). BODY: %s, META: %s" % (exc, self.request.body, self.request.META))
```
noisecapella: Instead of putting the exception message in the string, can you chain it like this: `raise FluidReviewException("...") from exc`?

mbertrand: Sure, but I'd like to keep `exc.detail` in the string. (`Could not satisfy the request Accept header.`, `You do not have permission to perform this action`).

PR #192, opened 2018-01-17
Facebook pixel tracking

### LINKED ISSUE #182: Send FB event to bootcamp.mit.edu FB account on payment
As a Bootcamp admin, I'd like to know whether my Facebook ad purchases are converting to paid bootcamp attendees. To do this, I need the bootcamp-ecommerce site to send an event to "purchase" event to Facebook 

### acceptance criteria
- [ ] add "Facebook Pixel" code to `/pay` view (see below). (Facebook asks that we add the pixel to every page on the site, but I don't see why that's necessary)
- [ ] add Facebook event code for "purchase" event that fires *after* the purchase is complete (i.e., after Cybersource redirects the user back to us with a completed purchase). Details below.
- [ ] the event should include the amount paid by the user

### similar issues
- #178 The bootcamp team would like to do similar conversion tracking with Google.

mbertrand:
Closes #182

#### What's this PR do?
- Adds Facebook pixel tracking using the `react-facebook-pixel` plugin
- Sends pageview info to FB on the payment page
- Sends payment info to FB on successful order payment

- Create your own Facebook pixel id at https://www.facebook.com/events_manager/pixel/
- Set `FACEBOOK_PIXEL_ID=<your_id>` in your `.env` file
- Go to the `/pay` page with the network tab open.  There should be a request made to Facebook with parameters:
    - `ev=PageView`
    - `dl=<your local bootcamp payment url>`
    - `id=<your_pixel_id>`
- Start a payment for your klass.  
- In the admin console, set the order status to `Fulfilled`
- Go to `../pay?status=receipt&order=<order_id>` with the network tab open.  There should be a request made to Facebook with parameters:
    - `ev=Purchase`
    - `cd[value]=<amount_paid>`
    - `cd[currency]='USD'`
- Eventually, pageview and purchase events should show up at https://www.facebook.com/events_manager/pixel/


gsidebo:
Functionality looks great. Getting around to code now

static/js/util/util_test.js
```diff
@@ -82,6 +83,41 @@ describe("util", () => {
...
+      const klass = klasses[0]
+      klass.payments[0].order.status = "created"
+      assert.deepEqual(
+        getPaymentWithFulfilledOrder(
+          klass.payments,
+          klass.payments[0].order.id
+        ),
+        undefined
+      )
+    })
```
gsidebo: I think this test case and the first test case at line 87 can be parametrized. I don't actually know if this will work & I'm not wild about the variable scope & the changing of value inside a `beforeEach`. Anyway, I think it's valuable to parametrize very similar tests where we can. Let me know what you think

```javascript
const klass, payment;

beforeEach(() => {
  klass = generateFakeKlasses(1, { hasPayment: true })[0]
  payment = klass.payments[0]
})

[
  ["created", undefined, "undefined", "not fulfilled"],
  ["fulfilled", payment, "a payment", "fulfilled"],
].forEach((orderStatus, expectedValue, valueDescriptor, statusDescriptor) => {
  it(`returns ${valueDescriptor} if the order is {statusDescriptor}`, () => {
      klass.payments[0].order.status = orderStatus
      assert.deepEqual(
        getPaymentWithFulfilledOrder(
          klass.payments,
          klass.payments[0].order.id
        ),
        expectedValue
      )
    })
```

README.md
```diff
@@ -73,6 +73,8 @@ a template to create your ``.env`` file. For Bootcamp to work, it needs 6 values
 
     The URL of your bootcamp instance on Docker (``http://<BOOTCAMP_DOCKER_MACHINE_IP>:8099``)
 
+- ``FACEBOOK_PIXEL_ID`` - Log into Facebook and create one at https://www.facebook.com/events_manager/pixel/
```
gsidebo: We have the devops FB account now. We should probably mention that as option 1, and suggest creating one's own pixel as a fallback

static/js/util/util_test.js
mbertrand: In this case I feel like parameterization would make it more difficult to understand at a glance what the test code is trying to accomplish; it would sacrifice readability for compactness.  But I like the idea of creating the klass and payment in `beforeEach`

mbertrand: Added the `beforeEach` function but not parameters here and in previous test case, let me know if okay

static/js/util/util_test.js
```diff
@@ -82,6 +85,44 @@ describe("util", () => {
...
+  describe("getPaymentWithFulfilledOrder", () => {
+    let klass, payment
+
+    beforeEach(() => {
+      klass = generateFakeKlasses(1, { hasPayment: true })[0]
+      payment = klass.payments[0]
+    })
+
+    it("gets the payment for the fulfilled order", () => {
+      assert.deepEqual(
```
gsidebo: Would be better to be explicit about setting the status to "fulfilled" rather than relying on the generate function to do that automatically.

mbertrand: Will do

PR #218, opened 2018-08-29
Remove IS_OSX from env.sh

noisecapella:
See also https://github.com/mitodl/cookiecutter-djangoapp/pull/136

#### What's this PR do?
Removes detection for OS X since Docker for Mac is used by everyone

Nothing should break

### DIFF
--- a/scripts/envs.sh
+++ b/scripts/envs.sh
@@ -1,19 +1,6 @@
 #!/usr/bin/env bash
 set -e -o pipefail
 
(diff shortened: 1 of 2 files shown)


pdpinch:
@annagav can you review this for George?

annagav:
@noisecapella, the README file still instructs to run docker-machine. Maybe we should update that as well so that we are not contradicting on how the app should be run?

PR #221, opened 2018-09-06
Add award id as parameter to success url

### LINKED ISSUE #217: Include award id as parameter in success URL
As a Bootcamp team member, I'd like to track goals for each bootcamp separately. 

See ZenDesk [#20277](https://odl.zendesk.com/agent/tickets/20277)

We currently track the same success URL for all bootcamps, `/pay/?status=receipt&order=[order.id]`. To track distinct success URLs for each bootcamp, we can append the award id to the URL, e.g. `pay/?status=receipt&order=17&award=84420` -- where the value of `award` comes from klass.id

annagav:
Fix #217

#### What's this PR do?
Add award id as parameter to success url

Tests should pass.

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -158,7 +158,7 @@ def generate_cybersource_sa_payload(order, redirect_url):
         'override_custom_receipt_page': "{base}?status=receipt&order={order}&award={award}".format(
             base=redirect_url,
             order=order.id,
-            award=klass.id
+            award=klass_key
         ),
         'reference_number': make_reference_id(order),
         'profile_id': settings.CYBERSOURCE_PROFILE_ID,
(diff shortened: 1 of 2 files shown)


ecommerce/api.py
```diff
@@ -155,9 +155,10 @@ def generate_cybersource_sa_payload(order, redirect_url):
         'item_0_unit_price': str(order.total_price_paid),
         'line_item_count': 1,
         'override_custom_cancel_page': "{base}?status=cancel".format(base=redirect_url),
-        'override_custom_receipt_page': "{base}?status=receipt&order={order}".format(
+        'override_custom_receipt_page': "{base}?status=receipt&order={order}&award={award}".format(
             base=redirect_url,
             order=order.id,
+            award=klass.id
```
noisecapella: I talked with @pdpinch just now and it should be `klass_key` instead of `klass.id`

Issue #208, opened 2018-04-23
Display warning/error message when user tries to pay more than amount owed


gsidebo:
Issue #207 is being solved temporarily by disabling the overpayment error message entirely so we can fix this issue for a specific user. We should be showing a warning or error message when a user attempts to pay more than the amount they owe, and the amount owed should be based on the user's personal price.

A couple ways to address this:
1. Add a keypress event to the payment field that displays a warning when the user enters an amount that is greater than the amount owed, but let them go through with the payment
1. Return an error in the payment API and show the user an error message (instead of failing silently as we were doing previously). This could be trivially extended to display other `ValidationError`s that we raise for other faulty inputs and likewise do not show an error message (e.g.: entering an amount less than zero)

roberthouse54:
I think I prefer option 2. So we can just show a validation error on submission correct?

pdpinch:
I don’t completely trust our data for required payments, so I’d like users to have a way to “overpay” if they choose. 

Instead of a simple validation failure, could we display a pop-up with an “see you sure” message?

PR #224, opened 2018-09-11
Added confirmation dialog for over pay

amir-qayyum-khan:
fixes https://github.com/mitodl/bootcamp-ecommerce/issues/208

#### What's this PR do?
- Adds a conformation dialog when user over pay for a entity.

- go to payment module and add a big amount, bigger the `selectedKlass.price`

@pdpinch 
screen shot 2018-09-18 at 7 24 31 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
     "url": "https://github.com/mitodl/bootcamp-ecommerce.git"
   },
   "dependencies": {
+    "@material/dialog": "0.27.0",
+    "@mitodl/mdl-react-components": "0.0.3",
     "autoprefixer": "^7.2.4",
     "babel-core": "^6.26.0",
     "babel-eslint": "8.1.2",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -66,6 +67,7 @@
     "react-hot-loader": "^3.1.3",
     "react-page-click": "^4.0.2",
     "react-redux": "^5.0.6",
+    "react-tap-event-plugin": "2.0.1",
```
mbertrand: Is this necessary? It's been deprecated and "since the problem it solves has been fixed in most browsers by now you should migrate away from this plugin." - https://www.npmjs.com/package/react-tap-event-plugin

package.json
```diff
@@ -48,6 +48,7 @@
     "jsdom-global": "^3.0.2",
     "keycode": "^2.1.9",
     "lodash": "^4.17.4",
+    "material-ui": "0.18.1",
```
mbertrand: Maybe it would be best to use `@material/dialog` like open-discussions does for consistency, @aliceriot what do you think?

package.json
amir-qayyum-khan: i was facing exception, had to add it.
```js
Warning: Unknown prop `onTouchTap` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by Overlay)
    in Overlay (created by DialogInline)
    in div (created by DialogInline)
    in DialogInline
End of output
npm ERR! Test failed.  See above for more details.```

package.json
amir-qayyum-khan: sounds good to me

amir-qayyum-khan:
Done @mbertrand

PR #225, opened 2018-09-13
Create Bootcamp when award_id has no corresponding klass_key

### LINKED ISSUE #183: When we receive a webhook with an unrecognized bootcamp/klass, create it
Currently, we only process webhooks when the incoming `award_id` matches an existing `klass_id`. This puts an unnecessary configuration roadblock (i.e. me) in the way of creating new bootcamps  . 

We should be able to use the FluidReview award API to gather enough information to create the necessary bootcamp, class and installment. If there is data that isn't available, we can request the that it be added to the FluidReview schema. 

Since creating a new bootcamp is a significant change to the system, we may want to send an e-mail notification to the django admin address when they are created.

annagav:
Fix #183

#### What's this PR do?
Create a new Bootcamp and Klass when an incoming webhook's  award_id does not match an existing klass_id.

Tests should pass.

### DIFF
--- a/fluidreview/api.py
+++ b/fluidreview/api.py
@@ -12,7 +12,7 @@ from django.contrib.auth.models import User
 from django.utils.encoding import smart_text
 from requests_oauthlib import OAuth2Session
 
-from klasses.models import Klass
+from klasses.models import Klass, Bootcamp
 from ecommerce.models import Line, Order
 from profiles.models import Profile
 from fluidreview.serializers import UserSerializer
@@ -235,14 +235,19 @@ def parse_webhook_user(webhook):
         user = profile.user
     if webhook.award_id is not None:
         personal_price = webhook.award_cost if webhook.amount_to_pay is None else webhook.amount_to_pay
-        klass = Klass.objects.get(klass_key=webhook.award_id)
-        if personal_price is not None:
-            user.klass_prices.update_or_create(
-                klass=klass,
-                defaults={'price': personal_price}
-            )
+        klass = Klass.objects.filter(klass_key=webhook.award_id).first()
+        if klass:
+            if personal_price is not None:
+                user.klass_prices.update_or_create(
+                    klass=klass,
+                    defaults={'price': personal_price}
+                )
+            else:
+                user.klass_prices.filter(klass=klass).delete()
         else:
-            user.klass_prices.filter(klass=klass).delete()
+            klass_info = FluidReviewAPI().get('/awards/{}'.format(webhook.award_id)).json()
+            bootcamp = Bootcamp.objects.create(title=klass_info['name'])
+            Klass.objects.create(bootcamp=bootcamp, title=klass_info['description'], klass_key=klass_info['id'])
 
 
 def list_users():
--- a/fluidreview/api.py
+++ b/fluidreview/api.py
@@ -14,6 +14,7 @@ from requests_oauthlib import OAuth2Session
 
 from klasses.models import Klass, Bootcamp
 from ecommerce.models import Line, Order
+from mail.api import MailgunClient
 from profiles.models import Profile
 from fluidreview.serializers import UserSerializer
 from fluidreview.constants import WebhookParseStatus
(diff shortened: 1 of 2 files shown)


fluidreview/api.py
```diff
@@ -235,14 +235,19 @@ def parse_webhook_user(webhook):
...
         personal_price = webhook.award_cost if webhook.amount_to_pay is None else webhook.amount_to_pay
-        klass = Klass.objects.get(klass_key=webhook.award_id)
-        if personal_price is not None:
-            user.klass_prices.update_or_create(
-                klass=klass,
-                defaults={'price': personal_price}
-            )
+        klass = Klass.objects.filter(klass_key=webhook.award_id).first()
+        if klass:
+            if personal_price is not None:
```
mbertrand: This won't get run if the klass & bootcamp need to be created, can you reorder the code to something like this:
```python
    if webhook.award_id is not None:
        klass = Klass.objects.filter(klass_key=webhook.award_id).first()
        if not klass:
            klass_info = FluidReviewAPI().get('/awards/{}'.format(webhook.award_id)).json()
            bootcamp = Bootcamp.objects.create(title=klass_info['name'])
            klass = Klass.objects.create(bootcamp=bootcamp, title=klass_info['description'], klass_key=klass_info['id'])
        personal_price = webhook.award_cost if webhook.amount_to_pay is None else webhook.amount_to_pay
        if personal_price is not None:
            user.klass_prices.update_or_create(
                klass=klass,
                defaults={'price': personal_price}
            )
        else:
            user.klass_prices.filter(klass=klass).delete()
```

fluidreview/api_test.py
```diff
@@ -200,6 +200,37 @@ def test_process_user_both_exist_with_fluid_id(mocker):
...
+        'award_cost': '400',
+        'award_name': 'TEST CAMP'
+    }
+    body = json.dumps(data)
+    hook = WebhookRequest(body=body)
+    parse_webhook(hook)
+
+    assert Klass.objects.filter(klass_key=award_id).exists()
+    assert Bootcamp.objects.filter(title=award_name).exists()
+
```
mbertrand: Can you add a test to check that a `PersonalPrice` object exists?

mbertrand:
The issue also mentions maybe sending an email when a new bootcamp/klass is created, not sure if Peter still wants that.

mbertrand:
@annagav Peter suggested raising an exception if a new bootcamp/klass needs to be created but both `award_cost` and `amount_to_pay` are null/empty in the webhook (to avoid users owing $0), and sending an email to the support address when a bootcamp & klass are successfully created.

fluidreview/api_test.py
```diff
@@ -200,6 +201,59 @@ def test_process_user_both_exist_with_fluid_id(mocker):
     mock_save_profile.assert_not_called()
 
 
+@pytest.mark.parametrize(['amount_to_pay', 'award_cost', 'sends_email'], [
+    ['', '1000.10', False],
+    ['25.99', '', False],
+    ['', '', True]
```
mbertrand: Should send an email as long as either `award_id` or `amount_to_pay` is a numerical value:
```python
    ['', '1000.10', True],
    ['25.99', '', True],
    ['', '', False]
```

fluidreview/api_test.py
```diff
@@ -200,6 +201,59 @@ def test_process_user_both_exist_with_fluid_id(mocker):
...
+        'award_id': award_id,
+        'award_cost': award_cost,
+        'award_name': 'TEST CAMP'
+    }
+    body = json.dumps(data)
+    hook = WebhookRequest(body=body)
+
+    with patch(
+        'fluidreview.api.MailgunClient.send_individual_email',
+    ) as send_email:
```
mbertrand: Could also do `send_email = mocker.patch('fluidreview.api.MailgunClient.send_individual_email')` instead, near beginning of test code.

fluidreview/api_test.py
```diff
@@ -200,6 +201,59 @@ def test_process_user_both_exist_with_fluid_id(mocker):
...
+        if sends_email:
```
mbertrand: Should call `parse_webhook(hook)` first:
```python
    parse_webhook(hook)
    if not sends_email:
        assert hook.status == WebhookParseStatus.FAILED
        assert send_email.call_count == 0
    else:
        assert hook.status == WebhookParseStatus.SUCCEEDED
        assert Klass.objects.filter(klass_key=award_id).exists()
```

annagav:
@mbertrand, thanks. I addressed your comments.

PR #237, opened 2018-11-11
update requirements

pdpinch:

none

#### What's this PR do?

bumps versions of requests and edx-api-client

Does the app pass tests? Can you sign in with edX credentials?


pdpinch:
@annagav can you review?

PR #240, opened 2018-12-04
Turn off codecov status updates

shaidar:
None

#### What's this PR do?
Disables the codecov status checks. Coverage will still get uploaded to codecov and I think codecov will still comment on the PR. This is needed to prevent failed code coverage from getting in the way of the automatic heroku deploys.

N/A


blarghmatey:
:+1:

PR #243, opened 2018-12-06
Add name and url to email

annagav:
None

#### What's this PR do?
When new bootcamp is created the email notifies about the creation with name and base url.

Try creating a new bootcamp


pdpinch:
Can we set the `sender_address` to a no-reply address?

PR #245, opened 2018-12-06
Check SMA webhooks for awards

annagav:
None

#### What's this PR do?
Allow the user to pay for awarded programs in SMApply.

Award user's application and make sure the user can pay for it.


pdpinch:
@mbertrand Thanks for catching that HTML issue. I saw that in production over the Thanksgiving holiday and I meant to open an issue but forgot. 

IMO we should address that in a separate PR.

PR #257, opened 2019-02-20
create klass title with award name instead of description

### LINKED ISSUE #255: SMApply: when creating klass from webhook, we are populating klass.title with the wrong field
It looks like we're using "Description" instead of "Title"

screen shot 2019-01-28 at 12 10 49 pm

I'm not sure if this is because the webhook is sending us the wrong value, or if we are requesting the wrong value from SMApply. 

Note: the bootcamp and the Klass can use the same title. They are no longer distinct concepts.

tasawernawaz:
fixes #255 

#### What's this PR do?
Update parse webhook method to created klass using award name instead of description.

### DIFF
--- a/smapply/api.py
+++ b/smapply/api.py
@@ -256,7 +256,7 @@ def parse_webhook_user(webhook):
             klass = Klass.objects.create(
                 bootcamp=bootcamp,
                 source=ApplicationSource.SMAPPLY,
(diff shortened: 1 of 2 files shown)


tasawernawaz:
@pdpinch Do we want to fix this for `fluidreview` as well ?

PR #258, opened 2019-03-03
update django-server-status, django, urllib3; remove pyyaml

pdpinch:

None

#### What's this PR do?

- upgrade django-server-status
- upgrade django 
- remove pyyaml, because I don't think we use it
- upgrade urllib3

tests pass and the app runs


pdpinch:
@tasawernawaz can you review this?

pdpinch:
@tasawernawaz

PR #279, opened 2019-07-11
update frontend dependencies

### LINKED ISSUE #277: Update/upgrade frontend packages
**Update front-end packages:**

- [ ] "lodash-es": "^4.17.14",
- [ ] "lodash.merge": "^4.6.2",
- [ ] "lodash.mergewith": "^4.6.2",
- [ ] "mem": "^4.0.0",
- [ ] "handlebars": "^4.1.0"

FYI @pdpinch

tasawernawaz:
Fixes #277 

#### What's this PR do?
Update following dep.
- lodash.mergewith@^4.6.2
- lodash-es@^4.17.14
- mem@^4.0.0 handlebars@^4.1.0
- lodash.merge@^4.6.2

- Nothing should break
- Test should pass.


tasawernawaz:
@annagav Issue you have sent me on slack, can be fixed by pinning `django-debug-toolbar==1.11`. Please review #280, once it will be merged we can move forward with this.

noisecapella:
This isn't an update when you are only adding packages to `package.json`. Can you remove the lines on `package.json` and use `yarn upgrade` to update the packages instead, so that the yarn lockfile gets updated?

tasawernawaz:
@noisecapella In my previous commit I had run `yarn upgrade lodash.mergewith@^4.6.2 lodash-es@^4.17.14 mem@^4.0.0 handlebars@^4.1.0 lodash.merge@^4.6.2` and `yarn.lock` was also updated. 

I just run `yarn upgrade` and it has upgraded some other packages as well. Honestly, I am not sure if we want to upgrade all or upgrade only listed in this PR.

Please take a look at this once more.

noisecapella:
What I'm getting at is that `handlebars` and the other dependencies that are added to `package.json` in this PR are not required by us but are dependencies of other packages. If there is a package which is requiring the old version of the dependency then we should update the package which depends on `handlebars`. We should not add `handlebars` to `package.json` because it will become a requirement for us when we upgrade.

You can see what packages depend on `handlebars` by looking at `yarn.lock`. You can see that `istanbul-reports` lists it as a dependency and in turn `nyc` lists `istanbul-reports` as a dependency. `nyc` is listed in `package.json`. So you should update `nyc` instead via `yarn add nyc@latest`, and the `handlebars` version should be updated at the same time. You can do the same for the other dependencies listed in the alerts too.

tasawernawaz:
@noisecapella thank you for add such detailed explanation. I will go through the list again and update the PR.

tasawernawaz:
@noisecapella I made the changes in the PR, could you please review it. Let me know if I missed anything or did incorrectly.

noisecapella:
@tasawernawaz I reverted the update of enzyme and react-redux. I think those libraries will require updating React and other libraries which needs some extra testing. I'm happy for us to do that update in a different PR. I'm not sure if it's needed for the security alert but we can find out after this is merged.

tasawernawaz:
@noisecapella thanks, are we good to merge this ?

noisecapella:
@tasawernawaz Yes please do

PR #296, opened 2019-09-17
Sync new SMApply users with local User and Profile models

### LINKED ISSUE #289: Periodically query SMApply for new users, and create profiles in bootcamp
There are no webhooks that fire when new accounts are created in SMApply, so we are left to setting up a celery task that periodically checks for new users, and creates them in bootcamp-ecommerce when necessary. 

### Acceptance Criteria

- [ ] Periodic task that makes a request to /api/v2/users/ 
- [ ] if user account doesn't exist, create it
- [ ] add a jsonfield to user profile to store the API response

sesterki:
closes #289 

#### What's this PR do?
This PR adds a periodic task that syncs SMApply users with the local User and Profile models. User JSON data is stored in a JSONField on Profile.

Open the terminal and run `sync_all_users()` from `smapply/tasks.py`
Check the User and Profile models to see that data was populated.


smapply/serializers.py
```diff
@@ -0,0 +1,10 @@
+"""Serializers for smapply"""
+from rest_framework import serializers
+
+
+class UserSerializer(serializers.Serializer):
+    """Serializer for SMApply user data"""
+    email = serializers.CharField()
+    first_name = serializers.CharField(allow_blank=True)
```
annagav: I think the first name and last name should be required.

sesterki: Should be required now.

Issue #298, opened 2019-09-17
Add json field to user profile to store response to demographic questions in SMApply


pdpinch:
As marketer, I want to see all the demographic information that arrives through Stage 1 of the SMApply application process synced to the Hubspot contact

#### acceptance criteria

- [ ] When we receive a webhook from SMApply, make a request to SMApply for the applicants most recent demographic data
- [ ] Store the response in a json field, so we don't have to worry about schema matching
- [ ] If the applicant applies to multiple klasses (programs), we will use the data submitted to the most recent klass (as indicated in the webhook)

PR #300, opened 2019-09-23
Sync user demographics when app receives webhooks

sesterki:
#298 

#### What's this PR do?
This PR adds code to sync demographics data with SMApply when the app receives a webhook. The data is stored on the Profile model.

Expose the app to the internet and direct webhooks at it. Change a user's status in SMApply and wait for the webhook to come in. This may be a bit tough.


smapply/api.py
```diff
@@ -243,6 +249,14 @@ def parse_webhook_user(webhook):
         user = process_user(serializer.data)
     else:
         user = profile.user
+
+    # Sync user demographics data
+    try:
+        store_demographics_data(user.profile, webhook.submission_id)
```
annagav: Why did you remove submission_id from required fields, but you are expecting it to be present in the webhook here?

sesterki: I didn't. I only removed `award`.

sesterki: `required_fields = field_mapping.keys()`. In this case keys was `id, award, user_id`.

smapply/api.py
```diff
@@ -206,17 +214,15 @@ def parse_webhook(webhook):
         webhook (WebhookRequestSMA): WebhookRequestSMA instance
 
     """
-    try:  # pylint:disable=too-many-nested-blocks
+    try:
         body_json = json.loads(smart_text(webhook.body))
         field_mapping = {'id': 'submission_id', 'award': 'award_id', 'user_id': 'user_id'}
-        required_fields = field_mapping.keys()
+        required_fields = ['id', 'user_id']
```
annagav: I am referring to these fields here. Why this change?

smapply/api.py
annagav: It looks like you are setting required fields to this `required_fields = ['id', 'user_id']` on line 220.

sesterki: My intention was to use this function for every webhook, and not all of them will have an `award`. The code that creates the PersonalPrice objects already handles this (`if webhook.award_id is not None:`). So I removed `award` from required fields.

annagav: In my first comment I was talking about `submission_id`, I did not mention anything about award_id.  Because in your call `store_demographics_data(user.profile, webhook.submission_id)` you are relying on submission_id, so why remove it from required fields?

smapply/api.py
sesterki: `field_mapping.keys()` = `['id', 'award', 'user_id']`

The field `id` is required in the webhook body, and maps to `submission_id` in the database. The required fields are applied to the webhook body. Requiring `id` is requiring `submission_id`.

This change does not remove the requirement for `submission_id`, it only removes the requirement for `award`/`award_id`.

smapply/api.py
sesterki: https://github.com/mitodl/bootcamp-ecommerce/pull/300/files#r328303582

smapply/api.py
annagav: Thanks for the explanation! 

I think there is a misunderstanding about the meaning of award. Award_id is used for retrieval of program info in SMApply, and it corresponds to our klass/bootcamp. So I don't think there can be a submission without the award(program) that they are applying for.

sesterki: Just tested and you are right, there is an award.

PR #314, opened 2019-10-15
Hubspot sync updates

sesterki:
Skip contact sync if message does not include email. Sync contact during smapply sync task

#### What's this PR do?
Skip contact sync if message does not include email. Sync contact during smapply sync task

Attempt to sync profiles without smapply_user_data and check that there are no new hubspot sync errors. Run the smapply task `sync_all_users` and check that new profiles were synced with hubspot.

### DIFF
--- a/hubspot/management/commands/sync_hubspot.py
+++ b/hubspot/management/commands/sync_hubspot.py
@@ -41,6 +41,13 @@ class Command(BaseCommand):
         sync_messages = [
             make_object_sync_message(obj.id, **kwargs)[0] for obj in objects
         ]
+
+        if object_type == "CONTACT":
+            # Skip sync if message is missing required field
+            sync_messages = [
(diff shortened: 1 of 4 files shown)


hubspot/management/commands/sync_hubspot.py
```diff
@@ -41,6 +41,13 @@ def bulk_sync_model(objects, make_object_sync_message, object_type, **kwargs):
         sync_messages = [
             make_object_sync_message(obj.id, **kwargs)[0] for obj in objects
         ]
+
+        if object_type == "CONTACT":
+            # Skip sync if message is missing required field
+            sync_messages = [
+                message for message in sync_messages if message.get('propertyNameToValues', {}).get('email')
```
kyfantaz: Just noting: this assumes that if the email field is populated, all the other fields we care about are there, too. A safe assumption if the only reason email isn't there is if the data population didn't happen.

PR #320, opened 2019-10-21
Fix hubspot sync issues, update tests

### LINKED ISSUE #317: Hubspot: first name, last name and some other custom fields aren't syncing
### Steps to Reproduce

1. Create a user in SMApply
2. Complete the first step of the application

### Expected Behavior

- A contact is created in Hubspot with the name, email and other demographic information for the user. 

### Actual Behavior

- The contact is created but the first name and last name are missing, as well as some other fields. 

### Example case

Compare the demographic data the came from SMApply in https://bootcamp.odl.mit.edu/admin/profiles/profile/5799/change/ and the data that arrived in Hubspot at https://app.hubspot.com/contacts/6119748/contact/12101/properties?eschref=%2Fcontacts%2F6119748%2Fcontact%2F12101&esclabel=Back

Empty fields include First name, Last name, City, Gender, Phone Number and Occupation

sesterki:
This PR deals with fixing hubspot sync issues. The hubspot bridge will need to be configured again on production by running the management command `configure_hubspot_bridge`. Since some data is missing it may also make sense to sync/resync all contacts again by running the management command `sync_hubspot --contacts`.

closes #317 

#### What's this PR do?
This PR fixes hubspot sync to properly handle first name, last name, and all multiple choice fields.

Load the shell and manually try syncing any profile that was missing information.

### DIFF
--- a/smapply/tasks.py
+++ b/smapply/tasks.py
@@ -21,3 +21,7 @@ def sync_all_users():
                 user.profile.save()
 
                 sync_hubspot_user(user.profile)
+        elif not profile.smapply_user_data:
+            profile.smapply_user_data = sma_user
+            profile.save()
+            sync_hubspot_user(profile)
--- a/smapply/tasks.py
+++ b/smapply/tasks.py
@@ -21,7 +21,7 @@ def sync_all_users():
                 user.profile.save()
 
                 sync_hubspot_user(user.profile)
-        elif not profile.smapply_user_data:
+        else:
             profile.smapply_user_data = sma_user
             profile.save()
             sync_hubspot_user(profile)
--- a/smapply/tasks_test.py
+++ b/smapply/tasks_test.py
@@ -56,23 +56,6 @@ def test_sync_all_users_bad_data(mocker):
     assert Profile.objects.count() == 1
 
 
-def test_sync_existing_user_lacking_data(mocker):
-    """
-    Test that an already existing profile without smapply_user_data has that data populated during sync_all_users.
-    """
-    profile = ProfileFactory.create(smapply_id=12345678)
-
-    mocker.patch(
-        'smapply.tasks.list_users',
-        return_value=test_user_data,
-    )
-    sync_all_users()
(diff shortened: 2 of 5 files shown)


smapply/tasks.py
```diff
@@ -21,3 +21,7 @@ def sync_all_users():
                 user.profile.save()
 
                 sync_hubspot_user(user.profile)
+        elif not profile.smapply_user_data:
```
mbertrand: Should this just be `else:`?  Otherwise, if a profile exists and `profile.smapply_user_data` is set once, it won't be updated again.

sesterki: Changed to else.

smapply/tasks_test.py
```diff
@@ -53,3 +54,20 @@ def test_sync_all_users_bad_data(mocker):
...
+
+    mocker.patch(
+        'smapply.tasks.list_users',
+        return_value=test_user_data,
+    )
+    sync_all_users()
+    assert Profile.objects.count() == 2
+
+    profile.refresh_from_db()
+    assert profile.smapply_user_data
```
mbertrand:
```python
@pytest.mark.parametrize("sma_data", [None, {"foo": "bar"}])
def test_sync_existing_user_data(mocker, sma_data):
    """
    Test that an existing profile with and without smapply_user_data has data populated during sync_all_users.
    """
    profile = ProfileFactory.create(smapply_id=12345678, smapply_user_data=sma_data)

    mocker.patch(
        'smapply.tasks.list_users',
        return_value=test_user_data,
    )
    sync_all_users()

    profile.refresh_from_db()
    assert profile.smapply_user_data == test_user_data[0]
```

sesterki: Added

mbertrand: You can remove the above test (`test_sync_existing_user_lacking_data`), it should test the same conditions.

sesterki: Okay

PR #323, opened 2019-10-24
Sync contacts in bulk and add a retry to handle too many requests errors

sesterki:
Fixes Hubspot 500 errors

#### What's this PR do?
This PR adds try again functionality to hubspot requests. It also uses bulk sync on `smapply.tasks.sync_all_users` and reduces how many users get synced each time the task runs.

Delete some or all Profiles from your local database and run `sync_all_users`. Check that there are no errors. When deployed to RC watch for errors. Also make sure that new Profiles are successfully synced to hubspot.

### DIFF
--- a/hubspot/decorators.py
+++ b/hubspot/decorators.py
@@ -1,0 +1,34 @@
+"""
+Decorators for hubspot module
+"""
+import functools
+from time import sleep
+
+from requests import HTTPError
+
+
+def try_again(func):
+    """
+    Wrapper for sending requests to hubspot. Attempts to send requests multiple times.
+    Hubspot does not seem to return 429 errors so any error is considered a too many requests error for the purposes
+    of try again.
+
+    The wrapped function should return a requests response object.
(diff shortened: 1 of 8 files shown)


hubspot/decorators.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Wrapper for sending requests to hubspot. Attempts to send requests multiple times.
+    Hubspot does not seem to return 429 errors so any error is considered a too many requests error for the purposes
+    of try again.
+
+    The wrapped function should return a requests response object.
+    """
+    @functools.wraps(func)
+    def wrapper_try_again(*args, **kwargs):
+        max_attempts = 3
```
mbertrand: Maybe (in a followup PR, this is fine for now) this could be configurable via a .env/request setting, so it could be tweaked without a code change.  Something like
```python
max_attempts = settings.HUBSPOT_MAX_RETRIES
```

PR #339, opened 2020-02-21
Hubspot contact serializer allow missing fields

### LINKED ISSUE #332: Missing fields while syncing data for Profile with SMApply Fields: {'why_apply', 'n...
Sentry Issue: [BOOTCAMP-3R](https://sentry.io/organizations/mit-office-of-digital-learning/issues/1273132606/?referrer=github_integration)

```
Missing fields while syncing data for Profile with SMApply Id 109988635. Fields: {'why_apply', 'nationality', 'referral_information', 'agree_see_links', 'jobtitle', 'industry', 'participant_license_release', 'alumni', 'last_name', 'referred_by', 'highest_education', 'linkedin_profile', 'first_name', 'mit_honor_code', 'liability_release', 'country', 'gender', 'date_of_birth', 'work_experience', 'confidentiality', 'company', 'hear_about_bootcamp', 'occupation', 'phone', 'other_programs', 'email'}
```

annagav:
Fix #332

#### What's this PR do?
For programs that have webhook setup for firing on application creation, this allows for the contact to successfully sync with hubspot, given that they haven't filled out the application yet, and the demographic data is still missing.

1) Hubspot requires only one field for contact creation: `email`.
The email is found in `Profile.smapply_user_data`. I updated the code to fill out this data in `process_user` procedure, instead of `sync_all_users` scheduled task.

2) Removed the check for missing fields when syncing user data

Setup a webhook for your program. It is already setup for "Test Program".
Create a new application in SMApply for a user that does not have a profile in your local Bootcamp and Hubspot.
If the webhook succeeds then a contact should get created in Hubspot for your user.

### DIFF
--- a/hubspot/serializers.py
+++ b/hubspot/serializers.py
@@ -87,12 +87,6 @@ class HubspotContactSerializer(serializers.ModelSerializer):
                 else:
                     response = responses[field_key]['response']
                 demographics_data[mapping] = response
-
-        # If any fields aren't present then the associated program has a typo, or is otherwise missing data
-        missing_fields = set(demo_sync_fields.values()) - set(demographics_data.keys())
-        if missing_fields:
-            log.error("Missing fields while syncing data for Profile with SMApply Id %s. Fields: %s",
-                      instance.smapply_id, missing_fields)
         return demographics_data
 
     def to_representation(self, instance):
--- a/smapply/api.py
+++ b/smapply/api.py
@@ -232,7 +232,6 @@ def process_user(sma_user):
     Returns:
         User: user modified or created by the function
     """
-
     serializer = UserSerializer(data=sma_user)
     serializer.is_valid(raise_exception=True)
 
--- a/smapply/api.py
+++ b/smapply/api.py
@@ -232,6 +232,10 @@ def process_user(sma_user):
     Returns:
         User: user modified or created by the function
     """
+
+    serializer = UserSerializer(data=sma_user)
+    serializer.is_valid(raise_exception=True)
+
     user, _ = User.objects.get_or_create(
         email__iexact=sma_user['email'],
         defaults={'email': sma_user['email'], 'username': sma_user['email']}
@@ -239,6 +243,7 @@ def process_user(sma_user):
     profile, _ = Profile.objects.get_or_create(user=user)
     if not profile.smapply_id:
         profile.smapply_id = sma_user['id']
+        profile.smapply_user_data = sma_user
         profile.name = '{} {}'.format(sma_user['first_name'], sma_user['last_name'])
         profile.save()
     return user
@@ -282,9 +287,8 @@ def parse_webhook_user(webhook):
     if not profile:
         # Get user info from SMApply API (ensures that user id is real).
         user_info = SMApplyAPI().get('/users/{}'.format(webhook.user_id)).json()
-        serializer = UserSerializer(data=user_info)
-        serializer.is_valid(raise_exception=True)
-        user = process_user(serializer.data)
+
+        user = process_user(user_info)
     else:
         user = profile.user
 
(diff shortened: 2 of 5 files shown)


smapply/api.py
```diff
@@ -232,13 +232,17 @@ def process_user(sma_user):
...
+    serializer.is_valid(raise_exception=True)
+
     user, _ = User.objects.get_or_create(
         email__iexact=sma_user['email'],
         defaults={'email': sma_user['email'], 'username': sma_user['email']}
     )
     profile, _ = Profile.objects.get_or_create(user=user)
     if not profile.smapply_id:
         profile.smapply_id = sma_user['id']
+        profile.smapply_user_data = sma_user
```
mbertrand: Can you add a line in `test_process_new_profile` to check that this field is populated correctly?
```python
assert Profile.objects.get(smapply_id=sma_user['id']).smapply_user_data == sma_user
```

smapply/tasks.py
```diff
@@ -22,9 +22,7 @@ def sync_all_users():
         if not profile:
             serializer = UserSerializer(data=sma_user)
             if serializer.is_valid():
-                user = process_user(serializer.data)
-                user.profile.smapply_user_data = sma_user
-                user.profile.save()
+                user = process_user(sma_user)
```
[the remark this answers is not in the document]
annagav: what if I remove the validation from `sync_all_users`, and we just call it inside `process_user`?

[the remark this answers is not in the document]
mbertrand: Sure, though I think you'd need to catch the exception there if the serializer is invalid, so other users will continue to be synced.

hubspot/serializers.py
```diff
@@ -87,12 +87,6 @@ def get_demographics_data(self, instance):
                 else:
                     response = responses[field_key]['response']
                 demographics_data[mapping] = response
-
-        # If any fields aren't present then the associated program has a typo, or is otherwise missing data
```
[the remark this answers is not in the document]
annagav: I think the only thing we care about is email, because they want to be able to email users about unfinished application.

annagav:
@mbertrand I addressed your comments, thanks!

smapply/tasks.py
```diff
@@ -20,13 +22,11 @@ def sync_all_users():
...
-            if serializer.is_valid():
-                user = process_user(serializer.data)
-                user.profile.smapply_user_data = sma_user
-                user.profile.save()
-
+            try:
+                user = process_user(sma_user)
                 profiles_to_sync.append(user.profile)
+            except:  # pylint: disable=bare-except
+                log.exception('Syncing profile for SMA user %s failed', sma_user['id'])
```
mbertrand: As soon as I started bootcamp up I got some of these exceptions, caused by users in smapply without last names.  Not sure if that's a good thing (we were missing errors we should have known about before) or a bad thing (these errors are expected and should be ignored)?

```ValidationError: {'last_name': [ErrorDetail(string='This field may not be blank.', code='blank')]}```

mbertrand: These errors will keep popping up whenever `smapply.tasks.sync_all_users` is called (default every 15 minutes) unless/until all the users are fixed in smapply.

mbertrand: works without logging exceptions now, @pdpinch should we be notified if some users are invalid when processed via `sync_all_users`?  On the dev smapply account there are users without last names which cause the validation errors.

pdpinch: Anna and I discussed this and I think it's OK as is. We care more about this users once they start the application process.

annagav: @mbertrand,I spoke with Peter about this, we are okay not getting notified at this stage, hoping that we are going to get an error when processing the webhook for that user, if the user decides to create an application.

annagav:
@mbertrand, I think I addressed all your comments. This is ready for another look. Thank you!

PR #387, opened 2020-04-14
Rename a couple UWSGI env vars, remove redundant if-env blocks

noisecapella:
Related to https://github.com/mitodl/micromasters/issues/4597

#### What's this PR do?
 - Renames `UWSGI_THREAD_COUNT` to `UWSGI_THREADS`
 - Removes redundant if-env blocks

If you have either of the above UWSGI_ environment variables, rename them. Then load the site and make sure it functions.


markbreedlove:
Please see mitodl/micromasters#4597 (comment)

> I recently made this change in cookiecutter-djangoapp: mitodl/cookiecutter-djangoapp#167

> My goal there was to preserve some kind of prefix, but not have uWSGI assign any configuration directives based on the variables.

PR #412, opened 2020-04-22
Add react-router, set up App.js

### LINKED ISSUE #408: Add react-router
We need react-router so we can re-use the front-end code for profiles from xPRO. It's not in bootcamps already because bootcamps predates our addition of react-router to the project cookie-cutter.

┆Attachments: <a href="https://github.com/mitodl/bootcamp-ecommerce/pull/412">bootcamp-ecommerce #412 Add react-router, set up App.js</a>

noisecapella:
Fixes #408 

#### What's this PR do?
Adds App.js and react-router so we can add future react pages similar to our other projects

Nothing should look different. In particular you should be able to go to `/pay` and everything should look the same


static/js/entry/payment.js
```diff
@@ -1,20 +1,29 @@
 // @flow
+import { createBrowserHistory } from "history"
```
mbertrand, line 2: What do you think of changing this filename to `root.js` and updating the webpack config to reflect that?

noisecapella: :+1:

annagav:
Everything seems to work well for me. This is probably not related to this PR but when I try to logout it takes me to the admin logout page. Should I open a separate issue for that?

noisecapella:
@annagav Yes, make a separate issue for that. Anything else for this PR?

PR #414, opened 2020-04-23
Port of authentication app and friends

### LINKED ISSUE #379: Support account creation and login via email
We want the UX of this to be the same as Open and xPro. It should still support login for edx.org users.

### LINKED ISSUE #400: OFAC/SDN export compliance
As a bootcamp admin, I want to be OFAC and SDN compliant. Any user whose name is on the SDN list, or lives in a country on the OFAC list should be prevented from applying to a bootcamp. 

We should handle this the exact same way we do in xPRO, checking the Cybersource export compliance API after the user enters their name and address in the onboarding flow.

rhysyngsun:
Part of #400 
Part of #379 

#### What's this PR do?
This PR adds backend support for registration and login via email/password. It is largely a port of the relevant applications from mitxpro. At a high level, the `authentication`, `compliance`, and `user` (`profile` in this project) apps were copied over, plus any related utilities or other code that was used by those applications, hence the extremely large diff here.

Additionally, `settings.py` was updated to use the new `env` module from xpro as well, since so many new vars were brought over it was worth making sure that our configuration was more robust.

Some adjustments were made in the `authentication` app, particularly `authentication.pipeline` to support both edx and email login simultaneously, this was required because the social auth actions written for xpro were built expecting to be called from a restful api. Instead, they now check to see if the API is being invoked in a traditional browser flow or via the API, redirecting to the URLs where our react auth frontend will be when we add that.

- Test the app without making any configuration changes, everything should still work, particularly:
  - Login via edX
  - Registration via edX
  - SMApply integration
  - Making payments
- Next, test the app with `FEATURE_SOCIAL_AUTH_API=True`:
  - If you register an edx user, you should be redirected to `/create-account/extra?partial_token=TOKEN`, this page isn't functional, we just want to be sure the redirect happens.
  - For an existing edx user, set `Profile.is_complete` to `True`, and login, you should end up on the dashboard as usual.

Note: we're not going to test the compliance app in this PR, it is pretty much entirely a copy/paste so there's not much worry and it'll be disabled in all envs anyway until we configure it. It's very well covered by tests in terms of the auth flow though, so it's not a huge concern.


profiles/admin.py
```diff
@@ -1,9 +1,84 @@
...
+    classes = ["collapse"]
+
+    def has_delete_permission(self, request, obj=None):
+        return True
+
+
+class UserAdmin(ContribUserAdmin, HijackUserAdminMixin, TimestampedModelAdmin):
+    """Admin views for user"""
+
+    include_created_on_in_list = True
```
mbertrand, line 57: Getting `Unable to lookup 'created_on' on User or UserAdmin` on the admin page

profiles/migrations/0006_enhanced_profiles.py
```diff
@@ -0,0 +1,111 @@
+# Generated by Django 2.2.10 on 2020-04-22 21:24
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+import profiles.models
+
+
+class Migration(migrations.Migration):
```
mbertrand, line 9: Would it be possible to add a python operation to create empty `LegalAddress` objects for each existing `Profile` in the database?

mbertrand:
Initial testing observations:
- after logging in via edx with an existing account, redirected to `http://bc.odl.local:8099/create-account/extra/?partial_token=<token>` (still not logged in)

- Logged in via hijack from django admin as an existing user:
  - Payment page works
  - SMApply webhook parsing is broken (still works on master).  API code is the same so my guess is this was caused by changing `SMAPPLY_AMOUNT_TO_PAY_ID = get_var("SMAPPLY_AMOUNT_TO_PAY_ID", None)` to `SMAPPLY_AMOUNT_TO_PAY_ID = get_string("SMAPPLY_AMOUNT_TO_PAY_ID", None)` in `settings.py`
```
web_1     | [2020-05-01 12:02:04] ERROR 18 [smapply.api] api.py:274 - [0b8bc70a5d16] - Webhook 88 body is not valid JSON or has invalid/missing values.
web_1     | Traceback (most recent call last):
web_1     |   File "./smapply/api.py", line 270, in parse_webhook
web_1     |     parse_webhook_user(webhook)
web_1     |   File "./smapply/api.py", line 311, in parse_webhook_user
web_1     |     personal_price = get_custom_field(application_meta['custom_fields'], settings.SMAPPLY_AMOUNT_TO_PAY_ID)
web_1     |   File "./smapply/api.py", line 465, in get_custom_field
web_1     |     if custom_field['id'] == field_id][0]['value']
web_1     | IndexError: list index out of range
```

If you want to keep the settings change, another way to fix this is by changing the function `smapply.api.get_custom_fields`:
```python
    return [custom_field for custom_field in custom_fields
            if str(custom_field['id']) == field_id][0]['value']
```

main/urls.py
```diff
@@ -35,8 +35,28 @@
...
+        name="password-reset-confirm",
+    ),
+    path("create-account/", index, name="signup"),
+    path("create-account/details/", index, name="signup-details"),
+    path("create-account/extra/", index, name="signup-extra"),
+    path("create-account/denied/", index, name="signup-denied"),
+    path("create-account/error/", index, name="signup-error"),
+    path("create-account/confirm/", index, name="register-confirm"),
+    path("account/inactive/", index, name="account-inactive"),
+    path("account/confirm-email/", index, name="account-confirm-email-change"),
```
mbertrand, line 58: I think these need to point to the `react` view instead of `index`

mbertrand:
PS If I login as an existing user with a complete profile, I am redirected to the payment page as expected.

rhysyngsun:
Discussed with @mbertrand regarding the redirect to `/?next=/create-account/extra/...`. This is because the `react` view is annotated with `@login_required`, which we'll need to remove, but we're going to defer on removing that until we have a `/signin` page that the frontend can redirect to instead.

I think this means unless there's any other issues with this PR, we're looking good?

@mbertrand  @annagav

mbertrand:
LGTM, just needs a rebase with master to resolve conflicts

annagav:
@rhysyngsun, I haven't had a change to test it yet. Would you like me to test the SMApply integration specifically?

rhysyngsun:
@annagav I was thinking yeah if you could test that and also review that area of code to see if there's anything that doesn't seem right to you

PR #415, opened 2020-04-23
Front-end code for profiles, registration

### LINKED ISSUE #379: Support account creation and login via email

### LINKED ISSUE #398: Learner Profile
As a learner, I'd like to be able to enter my basic profile information once, and re-use it for any bootcamp application. 

The list of profile fields is still under discussion, but we can start with the list we used for xPRO, and the same UX.

┆Attachments: <a href="https://github.com/mitodl/bootcamp-ecommerce/pull/415">bootcamp-ecommerce #415 Front-end code for profiles, registration</a>

### LINKED ISSUE #451: Login/registration front end
Copy over front-end code from xpro and adjust as necessary for updated dependencies

┆Attachments: <a href="https://github.com/mitodl/bootcamp-ecommerce/pull/415">bootcamp-ecommerce #415 Front-end code for profiles, registration</a>

mbertrand:
Closes #398
Closes #451
Closes #379 

#### What's this PR do?
Adds frontend code for profile pages and auth functionality, copied over from xpro, and installs required packages.  It does not add any links/buttons to the home page to get to the login, registration, or profile pages - that will be in a future PR.

With the setting `FEATURE_SOCIAL_AUTH_API` set to `True` or `False`, all the prior functionality (front page, login/registration via edx, payment page, receipt page) should still work. Tests should all pass.

- Set `FEATURE_SOCIAL_AUTH_API=True`
- Start at `http://bc.odl.local:8099/create-account/` and go through the registration flow.  You should get sent a confirmation email (if mailgun settings are properly configured), the confirmation link should send you to step 1 then 2 of new account creation, and finally to the payment page.
- Sign out.  Then go to `http://bc.odl.local:8099/signin/` and sign in as the new user.  You should be asked for your password then sent to the payment page.
- Go to `http://bc.odl.local:8099/account-settings/` and change your email.  You should get a confirmation link.  Clicking the link should bring you back to the account settings page and show your new email.
- Change your password.  Log out then back in with the new password.
- Once logged in, go to `../profile` and you should see your profile info.  Click the 'Edit Profile' button.  The profile edit form should load.  Make some changes and save, you should return to the profile page and it should reflect the changes you made.

#### Screenshots

Screen Shot 2020-05-01 at 7 04 25 AM

Screen Shot 2020-05-01 at 7 04 49 AM
Screen Shot 2020-05-01 at 7 03 52 AM

Screen Shot 2020-05-04 at 4 34 30 PM

Screen Shot 2020-05-04 at 4 35 27 PM

Screen Shot 2020-05-04 at 5 00 18 PM

Screen Shot 2020-05-04 at 5 01 56 PM

Screen Shot 2020-05-04 at 5 02 33 PM


mbertrand:
@pdpinch @abdulkdawson screenshots attached

mbertrand:
@aliceriot I made a few more changes and expanded the description.

@pdpinch @abdulkdawson one thing I didn't copy over from xpro was the notification popup (which would show when a confirmation email is sent for example) - should I add that here too, or hold off for now?

static/js/components/forms/EditProfileForm_test.js
```diff
@@ -0,0 +1,156 @@
...
+        errorMessage
+      )
+    })
+  })
+
+  //
+  ;[true, false].forEach(hasEmpty => {
+    it(`sets initialValues for the form${
+      hasEmpty ? "with some empty fields" : ""
+    }`, async () => {
```
aliceriot, line 128: it doesn't look like you're doing any async stuff here

static/js/components/forms/ProfileFormFields.js
```diff
@@ -0,0 +1,499 @@
...
+const US_POSTAL_CODE_REGEX = /[0-9]{5}(-[0-9]{4}){0,1}/
+const CA_POSTAL_CODE_REGEX = /[A-Z][0-9][A-Z] [0-9][A-Z][0-9]/
+const COUNTRIES_REQUIRING_POSTAL_CODE = [US_ALPHA_2, CA_ALPHA_2]
+const COUNTRIES_REQUIRING_STATE = [US_ALPHA_2, CA_ALPHA_2]
+
+const ADDRESS_LINES_MAX = 4
+const seedYear = moment().year()
+
+export const legalAddressValidation = yup.object().shape({
+  profile: yup.object().shape({
```
aliceriot, line 30: will `yup` let you do something like this?

```js
const requiredString = yup
  .string()
  .trim()
  .required()
```

then you could just supply the `.label` for a lot of these fields, like `name: requiredString.label('Full Name').min(2)`, etc. Could make this code a little more compact and DRY

aliceriot: also, should these big validation functions in general live in a separate file? maybe `lib/profile_validation`? not sure, maybe it's more convenient just to have them here to test alongside the component.

static/js/components/forms/ProfileFormFields.js
```diff
@@ -0,0 +1,499 @@
...
+        name="profile.job_function"
+        className="form-control"
+      >
+        <option value="">-----</option>
+        {EMPLOYMENT_FUNCTION.map((jobFunction, i) => (
+          <option key={i} value={jobFunction}>
+            {jobFunction}
+          </option>
+        ))}
+      </Field>
```
aliceriot, line 411: we have a few `selects` which look pretty similar, would it be worthwhile to add a component like this?

```js
function SelectField(props) {
  const { name, label, options } = props

  return (
    <>
      <label htmlFor={name} className="font-weight-bold">
        { label }
      </label>
      <Field
        component="select"
        name={name}
        className="form-control"
      >
        <option value="">-----</option>
        {options.map(([value, label], i) => (
          <option key={i} value={value}>
            {label}
          </option>
        ))}
      </Field>
    </>
  )
}
```

static/js/components/forms/elements/FormError.js
```diff
@@ -0,0 +1,12 @@
...
+type Props = {
+  children: any
+}
+
+export default class FormError extends React.Component<Props> {
+  render() {
+    const { children } = this.props
+    return <div className="form-error">{children}</div>
+  }
+}
```
aliceriot, line 12: this could just be a function instead of a class

static/js/lib/queries/users.js
```diff
@@ -0,0 +1,57 @@
...
+const DEFAULT_OPTIONS = {
+  options: {
+    method:  "PATCH",
+    headers: {
+      "X-CSRFTOKEN": getCookie("csrftoken")
+    }
+  }
+}
+
+export default {
```
aliceriot, line 33: why not just export individual functions?

static/js/lib/selectors.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import { createSelector } from "reselect"
+import { propOr } from "ramda"
+import qs from "query-string"
+
+export const qsSelector = createSelector(
+  (_, ownProps) => ownProps.location.search,
+  qs.parse
+)
```
aliceriot, line 9: I think instead of using this selector you could just lean on `react-routers`'s `useParams` hook. you should be able to do something like

```js
const { verification_code } = useParams()
```

in all your components

aliceriot: https://reacttraining.com/react-router/web/api/Hooks/useparams

static/js/components/PrivateRoute.js
```diff
@@ -0,0 +1,42 @@
...
+      }}
+    />
+  )
+}
+
+const mapStateToProps = createStructuredSelector({
+  currentUser: currentUserSelector
+})
+
+export default compose(connect(mapStateToProps))(PrivateRoute)
```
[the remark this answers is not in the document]
mbertrand: All these new files aren't really new - they're copied over from xpro, with just the bare minimum changes necessary to get them working and tests passing using the updated package dependencies.

static/js/lib/selectors.js
mbertrand: @aliceriot @pdpinch should the code be refactored according to these suggestions now or later in a subsequent PR?  And should the same changes be made to the matching files in mitxpro?

static/js/components/PrivateRoute.js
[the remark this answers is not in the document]
aliceriot: ah. well, in that case they should be cleaned up at some point, but we don't need to do it here I guess

static/js/lib/selectors.js
aliceriot: I didn't realize that all of this was just copied over code, I thought maybe just some of the library functions and so on were. I think it should be changed at some point but it doesn't need to be now. Maybe we should just file an issue for it? There are definitely things that I think should be cleaned up, like the `<select>` components, some code re-use improvements for the validation, and so on, but it's not necessary to do it to merge I think.

mbertrand:
@aliceriot I pushed a commit that fixes the issue with login/registration with edX when `FEATURE_SOCIAL_AUTH_API=True`.  As part of that I added another url query parameter, and so tried to modify some of the existing code to use react-router's `getParams()` instead as you suggested.  But I got an error stating that it could only be used in a function (I had placed it in `mapStateToProps`). Because of that I decided to emulate the selector approach used for the other params for now, and figure it can be fixed along with the others as part of #471

rhysyngsun:
Looks good on the edx side for new and existing users. One UX nit I noticed is that the registration page stays scrolled down when it advances to step 2 - would be nice if it scrolled to the top

aliceriot:
yeah hooks cannot be used in class-based components, only in functional ones

Issue #425, opened 2020-04-27
Rename "Klass"


gsidebo:
The idea of "klasses" is no longer really relevant, and it's confusing to leave it as-is for this "V2" update to the app. We should rename that model to something like `BootcampRun`, and the `klasses` app to `bootcamps`. For the latter, we would also want to rename the main Django app to something other than `bootcamp`.

┆Attachments: <a href="https://github.com/mitodl/bootcamp-ecommerce/pull/428">bootcamp-ecommerce #428 Renamed &#39;bootcamp&#39; app to &#39;main&#39;</a> | <a href="https://github.com/mitodl/bootcamp-ecommerce/issues/425">bootcamp-ecommerce #425 Rename &quot;Klass&quot;</a>

odlbot:
➤ Gavin Sidebottom commented:

pdpinch Just FYI this isn't completely done. The related issue that merged only partially addressed this

PR #428, opened 2020-04-28
Renamed 'bootcamp' app to 'main'

gsidebo:
Partially addresses https://github.com/mitodl/bootcamp-ecommerce/issues/425

#### What's this PR do?
Renames the base Django app from `bootcamp` to `main`

Just run the app

As it says in the linked issue, the name "klass" is no longer sensible/relevant to this app. Unfortunately the key Django app in this repo is called `klasses`, which might be too difficult to change. `bootcamp` is what the `klasses` app should have been called from the start. We may decide this change is too risky, but it was relatively straightforward for me to perform this renaming and the get app working. I figured that if we can't rename the `klasses` app, it will be good to at least rename this app to reduce confusion, especially once the `Klass` model itself is renamed


noisecapella:
I think the tricky part with `klasses` is the database tables. You could write some custom migrations to rename those.

PR #433, opened 2020-04-29
asadiqbal08/added resource pages in CMS

asadiqbal08:
[ticket](https://trello.com/c/hIdWN0H3/76-terms-of-service-page)

#### What's this PR do?
Added generic resource page structure.

Try to create some resource pages e.g. Privacy Policy, Term of Service etc. 

#### Screenshot: 
Screen Shot 2020-05-04 at 2 57 35 PM

##### NOTE: As we do not have much info about design so keep it in mind for future changes in corresponding scss, images and templates structure.


annagav:
@asadiqbal08, you have some conflicts that need to be resolved.

cms/models.py
```diff
@@ -38,3 +39,33 @@ def get_context(self, request, *args, **kwargs):
         context = super(BootcampPage, self).get_context(request)
         context["title"] = self.title
         return context
+
+
+class ResourcePage(Page):
+    """
+    Basic resource page for all resource page.
+    """
+    template = "../../bootcamp/templates/resource_template.html"
```
annagav, line 48: This path seems to be incorrect

asadiqbal08: loading fine at my end. further elaborate ?

asadiqbal08: updated: I think, something has been updated in settings.py file after rebase. anyway, I will do the corresponding fix.

annagav: Screen Shot 2020-04-30 at 10 11 24 PM

umarmughal824: Yes @annagav you are right. @asadiqbal08 gavin has renamed the base app from `bootcamp` to `main` so all your bootcamp app changes should be in `main` app. That's the issue here.

static/scss/resource.scss
```diff
@@ -0,0 +1,86 @@
...
+    text-decoration: none;
+
+    &:after {
+      content: "\e905";
+      font-family: icomoon;
+      margin-left: 10px;
+    }
+  }
+}
+
```
umarmughal824, line 86: It should be an empty line but now its two characters at the end of the file.

Issue #431, opened 2020-04-28
Jobma integration


odlbot:
As a bootcamp applicant, I need to receive an email invitation from the jobma site to record my video interview response and add it to my bootcamp application. 

As a bootcamp application reviewer, I need to be able to see which applicants have completed their video. I also want a link to view their video, so I can decide whether to accept them to the bootcamp.

odlbot:
➤ Peter Pinch commented:

After an initial conversation, georgeschneeloch said that we could write a django app with a couple of views that should be able to handle this.

odlbot:
➤ Peter Pinch commented:

georgeschneeloch what's next for this?

odlbot:
➤ George Schneeloch commented:

pdpinch For the first part, I think we need to ask people at Jobma how we can programatically get a link for an interview. I haven't been able to find anything myself.

For the second part, we should be set up to receiving notifications of finished interviews. But we need to update the webhook access token environment variable. I think ferdial may have this? From their doc: "Regarding the requests originating from the JOBMA, an access token will be generated and communicated upon the integration creation. This token will be included in the Authorization Header as well and token will be provided by integration website and configured". We also need to ask Jobma how to obtain a link to the video. I only see references to interview templates in their API mentioned in their doc.

odlbot:
➤ FerdiA commented:

georgeschneeloch pdpinch I sent them an e-mail

PR #444, opened 2020-04-30
Initial work for supporting Jobma

noisecapella:
Part of #431 

#### What's this PR do?
Some initial models for Jobma, a function to create interviews, and a webhook to update interview status

No manual testing yet


jobma/permissions.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class JobmaWebhookPermission(BasePermission):
+    """Restrict access to jobma via access token"""
+
+    def has_permission(self, request, view):
+        header = request.headers.get("authorization", "")
+        if header.startswith("Token "):
+            token = header[len("Token "):]
+            return token == settings.JOBMA_WEBHOOK_ACCESS_TOKEN
```
rhysyngsun, line 14: I think this should return `False` and log an error if `JOBMA_WEBHOOK_ACCESS_TOKEN` isn't set as a fail-safe mode.

jobma/api.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_jobma_client():
+    """
+    Get an authenticated client for use with Jobma APIs
+
+    Returns:
+        Session: A Jobma session object
+    """
+    session = Session()
+    session.headers["Authorization"] = f"Bearer {settings.JOBMA_ACCESS_TOKEN}"
```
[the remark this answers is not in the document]
noisecapella: Good idea

jobma/permissions.py
```diff
@@ -0,0 +1,25 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class JobmaWebhookPermission(BasePermission):
+    """Restrict access to jobma via access token"""
+
+    def has_permission(self, request, view):
+        header = request.headers.get("authorization", "")
+        if header.startswith("Token "):
```
rhysyngsun, line 17: I think this should assert on the webhook access token so if the var isn't set, it's not possible for an empty bearer token to successfully authenticate:

```suggestion
        if settings.JOBMA_WEBHOOK_ACCESS_TOKEN and header.startswith("Token "):
```

noisecapella: Oh sorry, I had this in a commit but I forgot to push it. What do you think of the new commit?

PR #446, opened 2020-05-01
Added bootcamp application models and admin

### LINKED ISSUE #399: Application & Enrollment data structure
[Wireframes](https://projects.invisionapp.com/share/D6WYOGRBHKY#/screens/414528114)
[Engineering doc](https://docs.google.com/document/d/1wFsZz_8DjT-8Vztmsro4yNE4Da70a_7y9jbLJKoIQIw/edit?usp=sharing)

We need to create some models to support the new bootcamp application workflow. The models are described in the engineering doc above. Note that the fields described for each model are not intended to be a complete list. 

#### Acceptance criteria: 
- Create models:
  - ApplicationStep
  - BootcampRunApplicationStep
  - BootcampApplication
  - ApplicationSubmission
  - VideoInterviewSubmission
  - QuizSubmission
- Add the `django-fsm` library as a dependency
- Create admin classes

#### What we're not doing as part of this issue:
- Adding the `CourseEnrollment` model. This can be done in a later issue, and the bootcamp application process does not require it to function correctly
- Renaming `Klass` to `BootcampRun`. Theres a separate issue for that: #425

┆Attachments: <a href="https://docs.google.com/document/d/1wFsZz_8DjT-8Vztmsro4yNE4Da70a_7y9jbLJKoIQIw/edit?pli=1#heading=h.smdet69ype0l">Bootcamps V2 - Application Process</a>

gsidebo:
Closes #399 

#### What's this PR do?
Adds the initial models and admin to support bootcamp applications

Click around the admin, refer to the Google Doc, see if everything makes sense

[Google doc](https://docs.google.com/document/d/1wFsZz_8DjT-8Vztmsro4yNE4Da70a_7y9jbLJKoIQIw/edit#heading=h.smdet69ype0l)


applications/admin.py
```diff
@@ -0,0 +1,201 @@
...
+
+class QuizSubmissionAdmin(SubmissionTypeAdmin):
+    """Admin for QuizSubmission"""
+    model = models.QuizSubmission
+
+    def get_list_display(self, request):
+        return tuple(super().get_list_display(request) or ()) + ("started_date",)
+
+
+class ApplicationStepSubmissionAdmin(TimestampedModelAdmin):
```
noisecapella, line 147: Is there a way to add validation that `content_object` exists before updating it? If the user enters an incorrect id number, the page crashes and they can't edit it to correct it.

applications/constants.py
```diff
@@ -0,0 +1,36 @@
...
+from enum import Enum
+
+
+class SubmissionTypes(Enum):
+    """Enum of possible bootcamp application submission types"""
+    VIDEO_INTERVIEW = "video_interview"
+    QUIZ = "quiz"
+
+
+VALID_SUBMISSION_TYPE_CHOICES = zip(
```
noisecapella, line 12: Can you add a `list(...)` around this and the other uses in this file? The `zip` will be exhausted after the first use.

main/utils.py
```diff
@@ -27,6 +29,15 @@ def webpack_dev_server_url(request):
     return 'http://{}:{}'.format(webpack_dev_server_host(request), settings.WEBPACK_DEV_SERVER_PORT)
 
 
+def now_in_utc():
```
noisecapella, line 32: Feel free to keep, but we can get the same functionality from https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.timezone.now

applications/constants.py
gsidebo: Great catch

applications/admin.py
gsidebo: Right, I meant to address this. My feeling is that object id shouldn't be editable at all

applications/admin.py
```diff
@@ -167,6 +184,13 @@ def get_queryset(self, request):
             .prefetch_related("content_object")
         )
 
+    def get_readonly_fields(self, request, obj=None):
+        # Only allow the content_type and object_id to be set if a new object is being created. Otherwise, set those
+        # to readonly.
+        if obj and obj.id is not None:
```
noisecapella, line 190: If we're only allowing setting these fields on objects which don't exist yet, should this condition be flipped?

gsidebo: `obj and obj.id is not None` means that the object has an id already, meaning it isn't a new object being created, hence we want these fields _added_ to the readonly field list

PR #455, opened 2020-05-05
asadiqbal08/Basic Bootcamp Run Page

asadiqbal08:
[Ticket](https://trello.com/c/P1QZQeQm/99-wagtail-product-page)

#### What's this PR do?
Add a basic bootcamp product page which will be accessible to `/bootcamps/[bootcamp_slug]`


cms/migrations/0002_auto_20200505_0933.py
```diff
@@ -0,0 +1,38 @@
...
+        ('wagtailforms', '0004_add_verbose_name_plural'),
+        ('klasses', '0014_no_max_len_application_stage'),
+        ('wagtailimages', '0001_squashed_0021'),
+        ('wagtailredirects', '0006_redirect_increase_max_length'),
+        ('cms', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='BootCampRunPage',
```
ahmed-belal, line 22: Please rename to `BootcampRunPage` (lowercase "c") following the convention from production website.

cms/migrations/0002_auto_20200505_0933.py
```diff
@@ -0,0 +1,38 @@
...
+            fields=[
+                ('page_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='wagtailcore.Page')),
+                ('description', wagtail.core.fields.RichTextField(blank=True, help_text='The description shown on the product page')),
+                ('content', wagtail.core.fields.StreamField([('rich_text', wagtail.core.blocks.RichTextBlock())], blank=True, help_text='The content of the benefits page')),
+                ('klasses', models.OneToOneField(help_text='The Klass for this page', null=True, on_delete=django.db.models.deletion.SET_NULL, to='klasses.Klass')),
```
ahmed-belal, line 27: Are you going to wait for @gsidebo PR for Klasses rename?

cms/models.py
```diff
@@ -10,10 +13,15 @@
 from cms.blocks import ResourceBlock
 
 
-class BootcampPage(Page):
+class BootcampProductPage(Page):
```
ahmed-belal, line 16: Any specific reason for this rename? Bootcamps are the only products here so I guess we're fine with just "BootcampPage" unless you have another reason.

cms/models.py
```diff
@@ -69,3 +76,33 @@ def get_context(self, request, *args, **kwargs):
...
+        return page context.
+        """
+        context = super(BootCampRunPage, self).get_context(request)
+        return context
+
+    def save(self, *args, **kwargs):
+        # autogenerate a unique slug so we don't hit a ValidationError
+        if not self.title:
+            self.title = self.__class__._meta.verbose_name.title()
+        self.slug = slugify("{}-{}".format(self.title, self.id))
```
ahmed-belal, line 107: Given that these URLs are going to be shared I think we should have a better slug pattern than `{title}-{id}`. Plus this looks like it was taken from xPRO `CourseProgramChildPage` which isn't intended to be used standalone that's why this slugging pattern works there but might need tweaks as these will be indexed by search engines as well.

cms/templates/product_page.html
```diff
@@ -0,0 +1,13 @@
...
+
+
+{% block title %}{{ site_name }} | {{ page.title }}{% endblock %}
+
+{% block headercontent %}
+  <div id="header"></div>
+  {% render_bundle 'header' %}
+{% endblock %}
+{% block content %}
+  <div>PRODUCT PAGE</div>
```
ahmed-belal, line 12: Can you also please include `{{ page.description }}` here?

cms/models.py
asadiqbal08: not specific, Just for better clarity that we are pointing to product.

cms/models.py
asadiqbal08: suggestion  ?

cms/migrations/0002_auto_20200505_0933.py
asadiqbal08: yes

cms/models.py
asadiqbal08: I guess, name `Bootcamp` would be good enough as we have this the only product.

Issue #197, opened 2018-01-22
Revert #192 Facebook pixel tracking


pdpinch:
The bootcamp team has decided to use Google Tag Manager which consolidates tracking across multiple sites and services, including Facebook.

PR #465, opened 2020-05-06
asadiqbal08/backend + Frontend changes for Learning Area Page

asadiqbal08:
##### NOTE for Reviewer:  We may need to update the design on later stage. 

#197 ## What are the relevant tickets?
[Ticket](https://trello.com/c/2xMyTPbH/97-product-page-learning-areas-section)

#### What's this PR do?
Added a backend stuff for adding learning area under bootcamp product run. 
```
    BOOTCAMPRUN (Product)
         /              \
        /                \
   LearningAreasPage .... OtherChildPages
```
#### Screenshots
--- 
Screen Shot 2020-05-06 at 6 14 25 PM

---
Screen Shot 2020-05-07 at 4 38 11 PM

----

Screen Shot 2020-05-07 at 4 38 18 PM

#### FrontEnd Possible Design

Screen Shot 2020-05-08 at 5 09 48 PM

screencapture-bc-odl-local-8099-bootcamps-bootcamp-1-2020-05-12-15_21_58

screencapture-bc-odl-local-8099-bootcamps-bootcamp-1-2020-05-08-17_10_14


cms/models.py
```diff
@@ -2,15 +2,44 @@
...
+    parent_page_types = ["BootcampRunPage"]
+
+    # disable promote panels, no need for slug entry, it will be autogenerated
+    promote_panels = []
+
+    @classmethod
+    def can_create_at(cls, parent):
+        # You can only create one of these page under bootcamp product.
+        return (
+            super(BootcampProductChildPage, cls).can_create_at(parent)
```
ahmed-belal, line 32: `super().can_create...`

cms/models.py
```diff
@@ -80,6 +113,42 @@ def save(self, *args, **kwargs):
...
+    )
+
+    image = models.ForeignKey(
+        Image,
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        help_text="Learning Area image size must be at least 150x50 pixels.",
+    )
```
ahmed-belal, line 143: Learning areas section requires exactly three images and rich text sections for each image.

cms/models.py
```diff
@@ -80,6 +113,42 @@ def save(self, *args, **kwargs):
...
+        blank=False,
+        help_text="Heading highlighting the learning areas generally.",
+    )
+    sub_heading = RichTextField(
+        null=True, blank=True, help_text="Sub heading for learning areas."
+    )
+
+    body = RichTextField(
+        null=True, blank=True, help_text="body section for learning areas."
+    )
```
ahmed-belal, line 134: The body has to be specific per image.

cms/models.py
```diff
@@ -80,6 +113,42 @@ def save(self, *args, **kwargs):
         self.slug = slugify("bootcamp-{}".format(self.product.klass_key))
         super().save(*args, **kwargs)
 
+class LearningAreasPage(BootcampProductChildPage):
+    """
+    Learning area's page for learning benefits.
+    """
+
+    subpage_types = []
```
ahmed-belal, line 121: This should be set on `BootcampProductChildPage` as almost all child pages would be the end leaf nodes.

cms/models.py
asadiqbal08: good catch 👍

cms/models.py
```diff
@@ -76,11 +117,24 @@ def save(self, *args, **kwargs):
...
         super().save(*args, **kwargs)
 
+class LearningAreasPage(BootcampProductChildPage):
+    """
+    Learning area's page for learning benefits.
+    """
+    learning_areas = StreamField(
+        [("learning_areas", LearningAreaBlock())],
+        blank=False,
+        help_text="Enter detail about learning areas upto max 3 blocks.",
```
ahmed-belal, line 127: This needs to be _exactly_ 3 blocks, not _up to_ 3 blocks.

ahmed-belal: Please add `min_num=3, max_num=3`. We need exactly 3 blocks, not up to 3 blocks here.

asadiqbal08:
> Blocking on streamfield block count.
Does it not applicable to `StreamBlock` instead `StructBlock` ?

cms/models.py
```diff
@@ -1,18 +1,53 @@
...
+        return (
+            super().can_create_at(parent)
+            and parent.get_children().type(cls).count() == 0
+        )
+
+    def save(self, *args, **kwargs):
+        # autogenerate a unique slug so we don't hit a ValidationError
+        if not self.title:
+            self.title = self.__class__._meta.verbose_name.title()
+        super().save(*args, **kwargs)
```
umarmughal824, line 43: Don't we need that line before `super().save()` like we did in xPRO?
```self.slug = slugify("{}-{}".format(self.get_parent().id, self.title))```

File: cms/models.py [hunk 7]
```diff
@@ -54,13 +89,25 @@ class Meta:
...
 
+    def _get_child_page_of_type(self, cls):
+        """Gets the first child page of the given type if it exists"""
+        child = self.get_children().type(cls).live().first()
+        return child.specific if child else None
+
     def get_context(self, request, *args, **kwargs):
         return {
-            **super().get_context(request),
-            "js_settings_json": json.dumps(_serialize_js_settings(request)),
```
umarmughal824, line 60: don't we need it anymore?

umarmughal824: `js_settings_json `

cms/models.py
ahmed-belal: @umarmughal824 is correct here. This will always generate the same title so the same kind of child page which in turn will translate into the same slug, thereby creating an integrity error at save.

cms/models.py
ahmed-belal, line 60: Please restore this or JS will be broken on all pages.

cms/models.py
```diff
@@ -77,7 +124,12 @@ class BootcampRunPage(BootcampPage):
         help_text="The bootcamp run for this page",
     )
 
-    content_panels = BootcampPage.content_panels + [FieldPanel("bootcamp_run")]
+    content_panels = [FieldPanel("bootcamp_run")] + BootcampPage.content_panels
+
+    @property
+    def product(self):
+        """Gets the product associated with this page"""
+        return self.bootcamp_run
```
ahmed-belal, line 132: There is no reason for this, please remove. We already have `bootcamp_run` and are not looking to add multiple types of products right now. This looks like a wasteful pre-optimization.

cms/templates/partials/learning_areas.html
```diff
@@ -0,0 +1,16 @@
+{% load image_version_url %}
+<div class="learning-area">
+  <div class="row row-relative">
+      {% for learning_area in page.learning_areas %}
+        <div class="col-sm-4 px-0">
+          <img class='w-100' style="height: 300px;" src="{% image_version_url learning_area.value.image 'max-1000x500' %}" alt="Learning Area" />
```
ahmed-belal, line 6: The rendition should be something like `fill-800x500`

static/scss/cms/learning-area.scss
```diff
@@ -0,0 +1,24 @@
...
+
+  padding-left: 15px;
+
+  h5, p {
+    color: dimgray;
+  }
+
+  .row-relative {
+    position: relative;
+  }
```
ahmed-belal, line 17: Please remove and use bootstrap's `position-relative` class instead.

cms/models.py
```diff
@@ -1,18 +1,53 @@
...
+        )
+
+    def save(self, *args, **kwargs):
+        # autogenerate a unique slug so we don't hit a ValidationError
+        if not self.title:
+            self.title = self.__class__._meta.verbose_name.title()
+        super().save(*args, **kwargs)
 
-from cms.blocks import ResourceBlock
-from main.views import _serialize_js_settings
```
ahmed-belal, line 15: Why is this being removed?

cms/models.py
asadiqbal08: not generating the same title, I believe, parent_id would be the same in child.

cms/models.py
asadiqbal08: @ahmed-belal I am using this for slug.

cms/models.py
ahmed-belal: I don't see `parent_id` anywhere 🤷‍♂️

cms/models.py
ahmed-belal: Please use `bootcamp_run`, there is no need to create a duplicate property just for a single check.

asadiqbal08: my bad, I am using the `self.bootcamp_run.run_key` below but forgot to remove this property. It was not in use.

cms/models.py
asadiqbal08: My question is, would be the parent_id be different or the same in that case ?

```
self.slug = slugify("{}-{}".format(self.get_parent().id, self.title))
```

cms/models.py
```diff
@@ -94,10 +146,25 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
 
+class LearningAreasPage(BootcampProductChildPage):
```
ahmed-belal, line 149: Could you please rename this and related classes similar to "ThreeColumnSection", "ThreeColumnBlock" etc? I think we had the same issue in xPRO where we started off naming these classes after where they were going to be used like "WhoShouldLearn", "LearningOutcomes" etc and then we had to rename them to "TextImageSection", "TextVideoSection" etc. because they needed to be reused on the homepage.

PR #480, opened 2020-05-07
Convert all tests to pytest

noisecapella:
Related to #473 

#### What's this PR do?
Removes ddt and converts all tests to pytest format

Tests should pass on travis

### DIFF
--- a/backends/utils_test.py
+++ b/backends/utils_test.py
@@ -2,14 +2,9 @@
 Tests for the utils
 """
 from datetime import datetime, timedelta
-from unittest.mock import (
-    MagicMock,
-    Mock,
-    patch,
-)
 
+import pytest
 import pytz
-from django.test import TestCase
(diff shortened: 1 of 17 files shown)


backends/utils_test.py
```diff
@@ -25,147 +20,141 @@
...
+    user = UserFactory.create()
+    user.social_auth.create(
+        provider='not_edx',
+    )
+    user.social_auth.create(
+        provider=EdxOrgOAuth2.name,
+        uid="{}_edx".format(user.username),
+        extra_data=social_extra_data
+    )
+    yield user
```
aliceriot, line 39: you have a similar user fixture in `backends/pipeline_api_test.py`, would it make sense for that to live in a test util library or something instead?

noisecapella: Sounds good, I just combined the two and moved it to `fixtures/common.py`

PR #484, opened 2020-05-08
Added internal API for starting applications and setting correct state

### LINKED ISSUE #458: Create API functions to start a new bootcamp application and set application state
We will want some internal API functionality to support the bootcamp application process:
1. A function that takes a bootcamp application, determines the correct `state` for that application by querying all of the relevant tables, and sets that `state` (this should be idempotent)
1. A function that creates a bootcamp application for a user if one doesn't already exist
1. (optional) A management command that invokes the first function. Parameters could be an application id, or a user id/email/username and a bootcamp run id/name/title

The second function would invoke the first function if a new application was created to ensure that a user is in the correct state when the application is created. The management command will be helpful in the future in case the state value of an application somehow gets out of sync with the actual state that the relevant data indicates

gsidebo:
Closes #458

#### What's this PR do?
Adds internal API functions and a management command for starting a new bootcamp application and deriving the state of an application

Run the management command for an existing bootcamp application

### DIFF
--- a/applications/api.py
+++ b/applications/api.py
@@ -1,0 +1,61 @@
+"""API for bootcamp applications app"""
+from django.db import transaction
+
+from applications.constants import AppStates, REVIEW_STATUS_REJECTED
+from applications.models import BootcampApplication
+from ecommerce.models import Order
+
+
+@transaction.atomic
+def get_or_create_bootcamp_application(user, bootcamp_run):
+    """
+    Fetches a bootcamp application for a user if it exists. Otherwise, an application is created with the correct
+    state.
+
+    Args:
+        user (User): The user applying for the bootcamp run
+        bootcamp_run (klasses.models.BootcampRun): The bootcamp run to which the user is applying
+
+    Returns:
+        BootcampApplication: The bootcamp application
+    """
+    bootcamp_app, created = (
+        BootcampApplication.objects
+        .select_for_update()
+        .get_or_create(
+            user=user,
+            bootcamp_run=bootcamp_run,
+        )
+    )
+    if created:
+        derived_state = derive_application_state(bootcamp_app)
+        bootcamp_app.state = derived_state
+        bootcamp_app.save()
+    return bootcamp_app
+
+
+def derive_application_state(bootcamp_application):  # pylint: disable=too-many-return-statements
+    """
+    Returns the correct state that an application should be in based on the application object itself and related data
+
+    Args:
+        bootcamp_application (BootcampApplication): A bootcamp application
+
+    Returns:
+        str: The derived state of the bootcamp application based on related data
+    """
+    if not hasattr(bootcamp_application.user, "profile") or not bootcamp_application.user.profile.is_complete:
+        return AppStates.AWAITING_PROFILE_COMPLETION.value
+    if not bootcamp_application.resume_file:
+        return AppStates.AWAITING_RESUME.value
+    submissions = list(bootcamp_application.submissions.all())
+    submission_review_statuses = [submission.review_status for submission in submissions]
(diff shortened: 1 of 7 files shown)


applications/api.py
```diff
@@ -0,0 +1,62 @@
...
+        bootcamp_application (BootcampApplication): A bootcamp application
+
+    Returns:
+        str: The derived state of the bootcamp application based on related data
+    """
+    if not hasattr(bootcamp_application.user, "profile") or not bootcamp_application.user.profile.is_complete:
+        return AppStates.AWAITING_PROFILE_COMPLETION.value
+    if bool(bootcamp_application.resume_file) is False:
+        return AppStates.AWAITING_RESUME.value
+    submissions = bootcamp_application.submissions.all()
```
noisecapella, line 52: Can we add a `list(...)` here to avoid querying this twice in `len(submissions)` below?

applications/api.py
```diff
@@ -0,0 +1,62 @@
...
+    submission_review_statuses = [submission.review_status for submission in submissions]
+    if any([status == REVIEW_STATUS_REJECTED for status in submission_review_statuses]):
+        return AppStates.REJECTED.value
+    elif any([status is None for status in submission_review_statuses]):
+        return AppStates.AWAITING_SUBMISSION_REVIEW.value
+    elif len(submissions) < bootcamp_application.bootcamp_run.application_steps.count():
```
noisecapella, line 58: Is it possible there could be two or more submissions for the same application step?

conftest.py
```diff
@@ -1,5 +1,14 @@
...
 from fixtures.autouse import *
 from fixtures.cybersource import *
+
+
+TEST_MEDIA_ROOT = "/var/media/test_media_root"
+
+
+def pytest_configure():
+    """Pytest hook to perform some initial configuration"""
+    settings.MEDIA_ROOT = TEST_MEDIA_ROOT
```
noisecapella, line 14: I think this should be `def pytest_configure(settings):` and then set the settings that way. I don't think you can set the settings using `django.conf.settings` in a test

gsidebo: I tried that and I got some confusing exception, so I copied this over from xpro (I think). I'll poke at it again

applications/api.py
gsidebo: Even if we wanted the user to, say, upload multiple files for a single step, I think we would still want to count that as a single submission. A step should consist of a user submitting something, then the admins reviewing it

applications/api.py
```diff
@@ -0,0 +1,62 @@
...
+    bootcamp_app, created = (
+        BootcampApplication.objects
+        .select_for_update()
+        .get_or_create(
+            user=user,
+            bootcamp_run=bootcamp_run,
+        )
+    )
+    if created:
+        derived_state = derive_application_state(bootcamp_app)
```
annagav, line 31: Is the user going to be able to create application for a bootcamp before he/she fills out the profile information?

gsidebo: It _shouldn't_ be possible, but I added that logic into the function so that (a) it would return the correct derived state no matter what application object it received, and (b) it would catch any users that slipped through the cracks and somehow ended up with an incomplete profile. If we want to make the assumption that a user's profile must be filled out if they are logged in, we should get rid of the "awaiting profile" state entirely

conftest.py
gsidebo: The error I got from attempting to use the `settings` fixture: `Argument(s) {'settings'} are declared in the hookimpl but can not be found in the hookspec`. I left a comment explaining the settings hack. It does work, but I understand it might have some side effects we haven't seen yet

PR #486, opened 2020-05-08
Bootcamp enrollments models

### LINKED ISSUE #468: Track bootcamp enrollments
We will need a new model to track a user's enrollment in a bootcamp run, and we will need to add logic to order fulfillment so that (a) an enrollment record is created, and (b) the state of the user's application to that bootcamp run is set to "complete" ([django-fsm usage](https://github.com/viewflow/django-fsm#usage))

Fields we will definitely need:

* user (fk)
* bootcamp_run (fk)
* order (fk, nullable)

Field(s) we will probably need, so it would be preferable to define them now:

* coursware_url (str, nullable)

Acceptance criteria:

* `BootcampRunEnrollment` model created
* A bootcamp run enrollment object is created when a user completes payment for a bootcamp run
* The state of a user's application to the bootcamp run is set to "complete" when a user completes payment for that bootcamp run

noisecapella:
Fixes #468 

#### What's this PR do?
Sets up `BootcampRunEnrollment` to track enrollments. There is also code to update the application state to `COMPLETE` when all payments are received.

Pay for a course, but enter a dollar amount below the cost of the course. You should not see a `BootcampRunEnrollment` created for that run. Pay the extra value, and you should see that it exists in the database for the run and user.


ecommerce/views.py
```diff
@@ -124,6 +126,14 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
             order.status = Order.FULFILLED
 
         order.save_and_log(acting_user=None)
+
+        try:
+            application = order.applications.get()
+            application.state = AppStates.COMPLETE.value
+            application.save()
```
gsidebo, line 133: Can you define a `@transition` method on the `BootcampApplication` model and invoke that here instead? https://github.com/viewflow/django-fsm#usage

ecommerce/api_test.py
```diff
@@ -366,3 +368,24 @@ def test_status(bootcamp_run, user):
...
     assert ex.value.args[0] == "Unable to find order {}".format(order.id)
+
+
+@pytest.mark.parametrize("has_personal_price", [True, False])
+@pytest.mark.parametrize("delta_paid, expected", [
+    [-1, False],
+    [0, True],
+    [1, True]
+])  # pylint: disable=too-many-arguments
+def test_is_paid_in_full(mocker, bootcamp_run, user, delta_paid, has_personal_price, expected):
```
gsidebo, line 379: I think this would be easier to understand and read with a parameter change:
```suggestion
@pytest.mark.parametrize(
  "run_price,personal_price,total_paid,expected", [
    [10, None, 5, False,],
    [10, None, 10, True,],
    [10, 5, 3, False,],
    [10, 5, 5, True,],
  ]
)
def test_is_paid_in_full(mocker, bootcamp_run, user, run_price, personal_price, total_paid, expected):
```

ecommerce/views.py
```diff
@@ -124,8 +128,25 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            )
+
+            try:
+                application = order.applications.get()
+                application.complete()
+                application.save()
+            except BootcampApplication.DoesNotExist:
+                log.exception("Missing application for order %d. Unable to set application state.", order.id)
+            except TransitionNotAllowed:
+                log.exception("Application received full payment but state cannot transition to COMPLETE")
```
gsidebo, line 146: This view is hefty as it is. Please create a separate API function for this. Something like `complete_order(order, run)`. I also feel like it would make sense to move the code that sets `order.status = Order.FULFILLED` into that function as well since that's one of the side effects we want when an order is successfully completed. If that makes sense and it's not a problem code-wise, we could change the `if` statement that runs before completing the order: `if decision == CYBERSOURCE_DECISION_ACCEPT and is_paid_in_full(user=order.user, bootcamp_run=run):`

noisecapella: I made most of these changes but I left the `order.status = Order.FULFILLED` part out of `complete_order` for consistency with xPRO and other similar ecommerce logic.

ecommerce/api.py
```diff
@@ -244,3 +270,43 @@ def make_reference_id(order):
...
+            except TransitionNotAllowed:
+                log.exception(
+                    "Application received full payment but state cannot transition to COMPLETE from %s for order %d",
+                    application.state,
+                    order.id,
+                )
+        except BootcampApplication.DoesNotExist:
+            log.exception("Missing application for order %d. Unable to set application state.", order.id)
+
+    try:
```
gsidebo, line 303: If it's easy enough to just remove this try/accept, I think we should do it. Matt has an open PR to remove this anyway

ecommerce/api.py
```diff
@@ -244,3 +270,43 @@ def make_reference_id(order):
             A reference number for use with CyberSource to keep track of orders
     """
     return "{}{}-{}".format(_REFERENCE_NUMBER_PREFIX, settings.CYBERSOURCE_REFERENCE_PREFIX, order.id)
+
+
+def complete_order(order):
```
gsidebo, line 275: I was thinking that this function would be called to perform all the side effects we want if an order was _successfully_ completed (maybe `complete_successful_order` or something like that would be a better name in that case). As it is now, this just shuffles some code from the view to this API method with its various responsibilities and I think it has limited usefulness. At a high level I think we could improve the view and API like this:

```python
# views.py
def post(self, request, *args, **kwargs):
  order = get_the_order(request)
  # receipt stuff...
  # bail out if errors are found or if the order doesn't need to change...
  if decision != CYBERSOURCE_DECISION_ACCEPT:
    handle_rejected_order(order)
  else:
    complete_successful_order(order)
  # this happens whether the order was rejected or accepted...
  sync_hubspot_deal_from_order(order)
  return Response()

# api.py
def complete_successful_order(order):
  # basically what you have already, plus setting the status flag, minus the API calls to hubspot and fluid/SMA...
```

That would mean setting the order status to "fulfilled" in this function, which I would still argue for despite it not matching what we have in xpro. One added benefit of rewriting the function in that way: we could call it from management commands or from our own scripts to accurately simulate a successful order completion.

ecommerce/api.py
noisecapella: I'm happy to have it removed but I would rather wait until @mbertrand's PR because it is a major change in behavior

gsidebo: Yup, sounds good

PR #489, opened 2020-05-08
Fix Jobma webhook permissions check

noisecapella:
Related to #431 

#### What's this PR do?
Fixes the authorization header check. I used the DRF docs instead of the Jobma docs to took at what was expected here.

There isn't a full workflow to test with yet so it would be cumbersome to set up locally.

### DIFF
--- a/jobma/permissions.py
+++ b/jobma/permissions.py
@@ -17,8 +17,8 @@ class JobmaWebhookPermission(BasePermission):
             log.error("JOBMA_WEBHOOK_ACCESS_TOKEN not set")
             return False
         header = request.headers.get("authorization", "")
-        if header.startswith("Token "):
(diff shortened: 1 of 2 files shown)


gsidebo:
Whoops, sorry @mbertrand

jobma/permissions.py
```diff
@@ -17,8 +17,8 @@ def has_permission(self, request, view):
             log.error("JOBMA_WEBHOOK_ACCESS_TOKEN not set")
             return False
         header = request.headers.get("authorization", "")
-        if header.startswith("Token "):
-            token = header[len("Token "):]
+        if header.startswith("Bearer "):
```
mbertrand, line 20: Minor nitpick: could make a constant for "Bearer", but works either way.

PR #496, opened 2020-05-11
Add support for interview_link from Jobma

noisecapella:
Related to #431 

#### What's this PR do?
Jobma's response has changed, so this stores the URL from Jobma so the end user can be directed to a place where they can take their interview

N/A


jobma/api.py
```diff
@@ -46,7 +50,10 @@ def create_interview(interview):
     })
     response.raise_for_status()
     result = response.json()
-    interview.interview_id = result["interview_id"]
-    interview.save_and_log(None)
+    if "interview_link" in result:
+        interview.interview_url = result["interview_link"]
+        interview.save_and_log(None)
+    else:
+        log.error("Interview link not found in payload %s", result)
```
gsidebo, line 57: Tiny nit: add a colon/dash/etc before the payload - `log.error("Interview link not found in payload - %s", result)`

PR #497, opened 2020-05-11
Add support for uploading a resume to an existing application

### LINKED ISSUE #448: API: Add endpoint for uploading a resume to an in-progress bootcamp application
An API endpoint should be created to support uploading a resume to a user's bootcamp application.

Some technical details:
1. We will need to add a field to `BootcampApplication` to track the date that the resume was uploaded (something like `resume_upload_date`)
1. A user should only be able to upload a resume if their profile has been filled out. **NOTE: We'll need to figure out what other limitations we want to put on this, e.g.: do we want to allow users to re-upload their resume, and under what conditions**
1. After the resume is successfully uploaded, if the user has also filled out their profile, the bootcamp application state should transition to "awaiting user submissions". See [django-fsm usage](https://github.com/viewflow/django-fsm#usage) for details on how to implement those transitions
1. We're not sure exactly which file types will be allowed, but we should get out in front of that requirement by limiting this API so that it only accepts the following file types: pdf, docx, doc, odt

annagav:
Fix #448

#### What's this PR do?
Adds a new field for resume upload date, adds support for saving a resume file, with the allowed extension, and triggers a transition in state for the application.

Run migrations, create a bootcamp application with various states, create a file with various file extensions, and call `process_upload_resume`. If the file is valid the `BootcampApplication.status` should change.


applications/api_test.py
```diff
@@ -118,3 +119,14 @@ def test_get_or_create_bootcamp_application(mocker):
...
     bootcamp_app = get_or_create_bootcamp_application(bootcamp_run=bootcamp_runs[1], user=users[1])
     assert bootcamp_app == existing_app
+
+
+@pytest.mark.django_db
+def test_process_upload_resume():
+    """
+    process_upload_resume should raise an exception if in wrong state
+    """
+    existing_app = BootcampApplicationFactory(state=AppStates.AWAITING_PROFILE_COMPLETION)
```
noisecapella, line 129: I think you need to add `.value` to the end of `AWAITING_PROFILE_COMPLETION`

applications/models.py
```diff
@@ -130,6 +133,18 @@ class BootcampApplication(TimestampedModel):
...
+    @transition(
+        field=state,
+        source=[AppStates.AWAITING_RESUME, AppStates.AWAITING_USER_SUBMISSIONS],
+        target=AppStates.AWAITING_USER_SUBMISSIONS
+    )
+    def upload_resume(self, resume_file):
+        """Save resume and make sure that the state can be transitioned to a new state"""
+        validate_file_extension(resume_file)
+        self.resume_file = resume_file
+        self.resume_upload_date = datetime.datetime.now()
```
noisecapella, line 145: Can we use `main.utils.now_in_utc` here so we have the time zone infomation included?

applications/api.py
```diff
@@ -59,3 +59,25 @@ def derive_application_state(bootcamp_application):  # pylint: disable=too-many-
     if bootcamp_application.order is None or bootcamp_application.order.status != Order.FULFILLED:
         return AppStates.AWAITING_PAYMENT.value
     return AppStates.COMPLETE.value
+
+
+class InvalidApplicationException(Exception):
```
noisecapella, line 64: Can we create `applications/exceptions.py` and move this exception there?

applications/models.py
```diff
@@ -130,6 +133,18 @@ class BootcampApplication(TimestampedModel):
     )
     state = FSMField(default=AppStates.AWAITING_PROFILE_COMPLETION.value, choices=VALID_APP_STATE_CHOICES)
 
+    @transition(
+        field=state,
+        source=[AppStates.AWAITING_RESUME, AppStates.AWAITING_USER_SUBMISSIONS],
```
noisecapella, line 138: These should also all end with `.value`

applications/utils.py
```diff
@@ -0,0 +1,13 @@
+"""
+Application utils
+"""
+import os
+from django.core.exceptions import ValidationError
+
+
+def validate_file_extension(value):
+    """Check the file extension is allowed"""
```
noisecapella, line 9: Can we add arguments to the docstring?

annagav:
@noisecapella, I addressed your comments. Thank you!

PR #499, opened 2020-05-11
Make MAILGUN_SENDER_DOMAIN not required

noisecapella:
None

#### What's this PR do?
Makes a setting not required so that it doesn't block deployments of PR builds

N/A


rhysyngsun:
This is required for the registration and password reset functionality to work, so I think we should keep this as required.

noisecapella:
In that case we should also make `MAILGUN_URL` required

Issue #447, opened 2020-05-04
API: Add endpoint that returns bootcamps that are available for application


odlbot:
**DEPENDS ON: https://github.com/mitodl/bootcamp-ecommerce/issues/468**

[Wireframe](https://projects.invisionapp.com/share/D6WYOGRBHKY#/screens/414528126)

We need an endpoint that fetches a serialized version of every bootcamp that the logged-in user is able to apply for. 

We may add some fields that indicate an overall due date for applications, but for now, we can assume that a user can apply for a bootcamp under these conditions:
1. The bootcamp has a bootcamp run ("klass") that has not started yet (`start_date` < now)
1. The user is not already enrolled in the above bootcamp run

**NOTE: Under current logic, bootcamp runs are only enrollable if we get a certain response back from the fluidreview or SMApply APIs. We need to change this and we need to figure out how.** One option would be to set a flag on the bootcamp/run that indicates that the bootcamp's enrollment is not handled through either of those services, and therefore it is open for application as long as the start date has not passed.

pdpinch:
@noisecapella what's the status of this?

noisecapella:
@gsidebo This is already done, right?

gsidebo:
@noisecapella There is an endpoint for returning all of a user's applications, but I'm pretty sure we don't yet have the endpoint that returns bootcamp runs that a user can apply to

PR #500, opened 2020-05-11
Remove smapply and fluidreview

### LINKED ISSUE #477: Remove FluidReview

### LINKED ISSUE #478: Remove SMApply and FluidReview
~~**DEPENDS ON https://github.com/mitodl/bootcamp-ecommerce/issues/447**~~ _(Since users won't be using our new application process until July, it won't be necessary to wait for the new endpoint that returns bootcamps that are available for application)_

This app currently relies on either SMApply or FluidReview, two third-party services, to do the following:
1. Tell the app which bootcamps are available for enrollment for a given user
1. Manage the admin workflow for approving users for bootcamp enrollment

We are taking this functionality in-house and we will no longer be relying on either of those third-party services, so we need to remove all interactions with and dependencies on SMApply and FluidReview.

### Acceptance criteria:
- Orders/enrollments that were completed through SMApply/FluidReview still exist, and there are still some flag(s) (et. al.) that indicates that the order/enrollment happened through those services
- ~~Our API to retrieve bootcamps that are available for enrollment are no longer fetched via SMApply/FluidReview (see: https://github.com/mitodl/bootcamp-ecommerce/issues/447)~~ _(Not needed)_
- `BootcampRun.source` field set to nullable
- All dependencies on third-party services removed from checkout/payment process (e.g.: bootcamp admissions client request in `ecommerce.api.create_unfulfilled_order`, `post_payment_fluid`/`post_payment_sma` calls in `ecommerce.views.OrderFulfillmentView.post`)
- Bootcamp admissions client (`klasses.bootcamp_admissions_client.BootcampAdmissionClient`) removed
- The `fluidreview` and `smapply` Django apps removed
- References to both services removed from the README

mbertrand:
Closes #477 
Closes #478 

#### What's this PR do?
Removes fluidreview and smapply code.

I kept the smapply and fluidreview fields intact in the `Profile` model so we know who came in through those apps.

Tests should pass, nothing should break, you should still be able to view what you owe for existing bootcamps and make new payments on them.


klasses/api.py
```diff
@@ -3,13 +3,32 @@
...
+
+    Returns:
+        list of BootcampRun: bootcamp runs the user has applied for
+
+    """
+    return (
+        BootcampRun.objects.prefetch_related("applications")
+            .filter(applications__user=user)
+            .filter(applications__state=AppStates.AWAITING_PAYMENT.value)
+            .values_list("run_key", flat=True)
```
mbertrand, line 28: @gsidebo I wrote this as a quick replacement for the old `BootcampAdmissionClient. payable_bootcamp_run_keys` function, just to keep the PR development moving.  Do these criteria make sense at all or should I just wait until #447 is complete?

gsidebo: @mbertrand You can combine those filters into one, but that's just a nit. This is a perfectly good replacement while you're hacking on this. You might already be going for this, but just to make sure, my impression was that the `BootcampAdmissionClient` class would be removed entirely as part of this PR. If that class needs to stay for any reason, let me know.

One other comment: please put https://github.com/mitodl/bootcamp-ecommerce/issues/478 in the relevant tickets section as well. That issue has some more detail and some rough acceptance criteria I wrote up (sorry, there are lots of duplicates for this!)

gsidebo: Another thing: https://github.com/mitodl/bootcamp-ecommerce/issues/447 is listed as a dependency, but looking at it now, I think it would be fine to merge this without that endpoint. We're getting rid of SMApply and Jobma as part of this PR anyway, which removes the existing endpoint(s) that return payable bootcamps. In production, users won't be using our new application process for some time, so I don't think it's necessary to have a ready-made replacement for those endpoints. I'll make a note of it on the issue

Issue #436, opened 2020-04-29
Minimal Site TopNav


odlbot:
As a user, I'd like basic site nav at the top of the page

### Wireframe
* See attached
* Design is not final 

### Acceptance Criteria
* [ ] MIT logo links to mit.edu
* [ ] Bootcamp logo links to home page (`href="/"`)
* [ ] Home links to home page (`href="/"`)
* [ ] "Admissions" will link to a static page served by the CMS, at `href="/admissions"`
* [ ] "Team" will link to a static page served by the CMS, at `href="/team"`

### Out of Scope
* "Bootcamps" drop-down
* User menu

┆Attachments: <a href="https://trello-attachments.s3.amazonaws.com/5e448428fb73a343e4eeac2c/5ea9e76ff110672ccf1e0eeb/3ba7712bbcb2197ef33699f5675ea3cc/Screen_Shot_2020-04-29_at_4.43.28_PM.png">Screen Shot 2020-04-29 at 4.43.28 PM.png</a> | <a href="https://projects.invisionapp.com/prototype/MIT-Bootcamps-Wireframe-sketch-ck9vklrqz008w0q01ud3k6a52/play/4399efa2">https:&#x2F;&#x2F;projects.invisionapp.com&#x2F;prototype&#x2F;MIT-Bootcamps-Wireframe-sketch-ck9vklrqz008w0q01ud3k6a52&#x2F;play&#x2F;4399efa2</a>

odlbot:
➤ Peter Pinch commented:

Added Invision Link

odlbot:
➤ Peter Pinch commented:

Added some link URLs to the description

PR #502, opened 2020-05-12
Minimal site topnav - #436

ahmed-belal:
#436 

#### What's this PR do?
Adds a minimal site top navigation bar on all non-React pages.

Visit any page that is rendered by Django (bootcamp page, for example). The top nav should appear as expected.

Screenshot 2020-05-12 at 1 55 42 PM

Screenshot 2020-05-12 at 1 55 56 PM


cms/templates/product_page.html
```diff
@@ -5,6 +5,7 @@
 {% block title %}{{ site_name }} | {{ page.title }}{% endblock %}
 
 {% block content %}
+    {% include "top_nav.html" %}
```
asadiqbal08, line 8: It is not be the part of base.html  ?

main/templates/top_nav.html
```diff
@@ -0,0 +1,33 @@
...
+            </li>
+            {% if not user.is_anonymous %}
+                <li class="nav-item">
+                    <!-- User dropdown here -->
+                </li>
+            {% endif %}
+        </ul>
+    </div>
+
+
```
asadiqbal08, line 32: nit: extra space

cms/templates/product_page.html
ahmed-belal: I'd rather keep it separate because we might need to migrate that over to React in chunks.

cms/templates/product_page.html
```diff
@@ -5,6 +5,7 @@
 {% block title %}{{ site_name }} | {{ page.title }}{% endblock %}
 
 {% block content %}
+  {% include "top_nav.html" %}
```
asadiqbal08, line 8: I think, it should not be specific to product page.

PR #507, opened 2020-05-12
Added endpoint for fetching detailed user application data

### LINKED ISSUE #450: API: Add endpoint to fetch detailed user application state
[Wireframe](https://projects.invisionapp.com/share/D6WYOGRBHKY#/screens/414528125)
[Initial bootcamp application models PR](https://github.com/mitodl/bootcamp-ecommerce/pull/446)

~~DEPENDS ON https://github.com/mitodl/bootcamp-ecommerce/issues/458** (assuming that we want this function to create a bootcamp application for the user/bootcamp run if one doesn't already exist)~~

We need an endpoint that fetches information about a user's progress in a bootcamp application. As the wireframe indicates, that information must include:
1. The bootcamp title, start date, and end date of the bootcamp run (aka "klass") that the user is applying to
1. Whether or not the profile is complete, and the last updated date for the profile
1. A resume upload date, or a null value if the resume has not been uploaded yet (`BootcampApplication.resume`)
1. A string indicating the overall state of the application (`BootcampApplication.state`)
1. If the user has completed an order that includes payment for the given bootcamp, the completed date (`BootcampApplication.order`)
1. The full set of reviewable steps for the bootcamp that the user is applying to, and the due dates for those steps (`ApplicationStep` and `BootcampRunApplicationStep` models)
1. The user's submission dates for each reviewable step (if submitted) and the review status of those submissions (if reviewed) (`ApplicationStepSubmission` model and linked `content_object`, which is either `VideoInterviewSubmission` or `QuizSubmission`)

Some technical notes:
- **Our wireframes do not yet have a UI for starting a new application to a bootcamp run. For now, it's probably best just to create a new `BootcampApplication` for the user if one does not already exist when a request is made to this endpoint**
- This should be a login-only view

gsidebo:
Closes #450 

#### What's this PR do?
Adds an endpoint for fetching detailed bootcamp application data ([wireframe](https://projects.invisionapp.com/share/D6WYOGRBHKY#/screens/414528125))

TBD

While hacking on this I realized that if we're keeping the same concept of installments, then a user's payment toward the bootcamp run they're applying to might be spread out over multiple orders. To account for that I removed the `BootcampApplication.order` field and added the `Order.application` field


noisecapella:
Can you rebase please?

applications/views.py
```diff
@@ -0,0 +1,28 @@
...
+    owner_field = "user"
+
+    def get_serializer_class(self):
+        if self.action == "retrieve":
+            return BootcampApplicationDetailSerializer
+        else:
+            raise NotImplementedError()
+
+    def list(self, request, *args, **kwargs):
+        return Http404
```
noisecapella, line 28: This is returning an error because it expects a `Response`. However, I think you can remove this method altogether and use `BootcampApplicationViewset(mixins.RetrieveModelMixin, GenericViewSet)` above instead of `ReadOnlyModelViewSet`. If we omit the `ListModelMixin` which is included in `ReadOnlyModelViewSet`, it will return a 404 by default for the list view.

applications/views.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    View for fetching users' serialized bootcamp application(s)
+    """
+    authentication_classes = (SessionAuthentication,)
+    permission_classes = (IsAuthenticated, UserIsOwnerPermission,)
+    queryset = BootcampApplication.objects.prefetch_state_data()
+    owner_field = "user"
+
+    def get_serializer_class(self):
+        if self.action == "retrieve":
```
noisecapella, line 22: I don't think this check is necessary here. Without the list view we only have `retrieve` as a valid action.

ecommerce/api.py
```diff
@@ -34,23 +34,29 @@
...
     """
-    return Line.objects.filter(
+    line_qset = Line.objects.filter(
         order__status=Order.FULFILLED,
         order__user=user,
         run_key=run_key,
-    ).aggregate(price=Sum('price'))['price'] or Decimal(0)
+    )
+    if application_id is not None:
+        line_qset.filter(order__application_id=application_id)
```
noisecapella, line 58: I think you need a `line_qset = line_qset.filter(...)` here. Also, we should have a test for this case to make sure the filter is applied.

applications/api.py
```diff
@@ -57,7 +57,16 @@ def derive_application_state(bootcamp_application):  # pylint: disable=too-many-
...
     elif len(submissions) < bootcamp_application.bootcamp_run.application_steps.count():
         return AppStates.AWAITING_USER_SUBMISSIONS.value
-    if bootcamp_application.order is None or bootcamp_application.order.status != Order.FULFILLED:
+    total_paid = get_total_paid(
+        bootcamp_application.user,
+        bootcamp_application.bootcamp_run.run_key,
+        application_id=bootcamp_application.id,
+    )
+    if (
+        total_paid is None or
```
noisecapella, line 66: `total_paid` should never be `None`. If there are no payments made it returns a `Decimal(0)` instead

applications/api.py
```diff
@@ -57,7 +57,16 @@ def derive_application_state(bootcamp_application):  # pylint: disable=too-many-
         return AppStates.AWAITING_SUBMISSION_REVIEW.value
     elif len(submissions) < bootcamp_application.bootcamp_run.application_steps.count():
         return AppStates.AWAITING_USER_SUBMISSIONS.value
-    if bootcamp_application.order is None or bootcamp_application.order.status != Order.FULFILLED:
+    total_paid = get_total_paid(
```
noisecapella, line 60: Do we want to consider personal price here? If so we might want to use `if not is_paid_in_full(...)` as a shortcut

main/utils.py
```diff
@@ -256,6 +256,20 @@ def get_field_names(model):
...
+    """
+    Returns a filename without a directory path
+
+    Args:
+        filepath (str): The file path (e.g.: "/path/to/file.txt")
+
+    Returns:
+        str: The filename without the directory path (e.g.: "file.txt")
+    """
+    file_parts = filepath.split("/")
```
noisecapella, line 269: I think this can just be `return os.path.split(filepath)[1]`

PR #515, opened 2020-05-14
Fix support links

### LINKED ISSUE #510: No support_email in JS
On the forgot your password page you can see that the link to the Customer Support Center at the bottom links to `mailto:undefined`. It looks like the code is referencing `SETTINGS.support_email` but this is not set in the JS bundle from the view.

mbertrand:
Fixes #510 

#### What's this PR do?
Replaces support email links with zendesk links

- Go to `../create-account` and submit an email.  Click the 'Customer Support Center' link on the following page, it should open up the zendesk support contact page.
- Go to `../forgot-password` and repeat the above.
- set `DEBUG=False` and go to a nonexistent URL, the 404 message should have a link to the zendesk page.

### DIFF
--- a/app.json
+++ b/app.json
@@ -124,6 +124,10 @@
       "description": "Email address listed for customer support",
       "required": false
     },
+    "BOOTCAMP_SUPPORT_URL": {
+      "description": "URL for customer support",
+      "required": false
+    },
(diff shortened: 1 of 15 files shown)


noisecapella:
Can we also update 404.html and 500.html to replace the support email with the link?

Issue #492, opened 2020-05-08
API: Add API endpoint for fetching checkout data for an application


odlbot:
This should include serialized information about the following: 
- All orders having to do with the given bootcamp application, including amounts paid (if orders exist already)
  - *This assumes that we are maintaining the current concept of installments, i.e.: a user can pay for a bootcamp over multiple orders. If we are maintaining that concept, then we will want to remove the `order` field from `BootcampApplication` and add a foreign key from `Order` to `BootcampApplication`*
- Installments (the due date of the next installment, the amount left to be paid on the active installment, and the total amount left to be paid for the bootcamp run)
- Personal pricing for the bootcamp run being applied to (if any personal prices exist)
- The bootcamp run that the user is applying to

PR #525, opened 2020-05-14
Move ecommerce-related views into ecommerce app

noisecapella:
Part of #492 

#### What's this PR do?
Moves a couple views with ecommerce data into the ecommerce app

Nothing should be changed as far as functionality goes.


klasses/api.py
```diff
@@ -28,58 +24,3 @@ def payable_bootcamp_run_keys(user):
                     Q(personal_prices__user=user))
             .values_list("run_key", flat=True)
     )
```
mbertrand, line 26: This function was written as a temporary replacement for the removed `BootcampAdmissionClient`. But like the other functions moved to `ecommerce/api.py` in this PR, it is ecommerce-related and is now the only function left here.  Should it be moved as well so this file can be deleted?

noisecapella: Yes, just made the change

PR #526, opened 2020-05-14
Add API for submitting review for application submissions

### LINKED ISSUE #463: Admin API: Add endpoint to set the review state of an application submission
[Wireframe](https://projects.invisionapp.com/share/3HWYOFP6R8B#/screens/414529299)

Some technical details:
1. An admin should only be able to set the review state of an application submission if the application is in the "awaiting review" state.
1. After review state is set on the submission, the state of the application should be transitioned. If the submission was approved, the state should be set to "awaiting payment" if the submission is the last one required by the application steps, or "awaiting user submissions" otherwise. If the submission was rejected, the state should be set to "rejected". See [django-fsm usage](https://github.com/viewflow/django-fsm#usage) for details on how to implement those transitions

annagav:
Fix #463

#### What's this PR do?
Add endpoint to set the review state of an application submission

Create two un-reviewed submissions, call `submit_review_for_application_submission` on one of the submissions, and make sure the `BootcampApplication` and `ApplicationStepSubmission` get updated accordingly.

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -14,8 +14,8 @@ from applications.constants import (
     VALID_SUBMISSION_TYPE_CHOICES,
     AppStates,
     VALID_APP_STATE_CHOICES,
-    VALID_REVIEW_STATUS_CHOICES
-)
+    VALID_REVIEW_STATUS_CHOICES,
+    REVIEW_STATUS_REJECTED)
 from applications.utils import validate_file_extension
 from main.models import TimestampedModel, ValidateOnSaveMixin
 from main.utils import now_in_utc
@@ -146,11 +146,11 @@ class BootcampApplication(TimestampedModel):
     @transition(
         field=state,
         source=AppStates.AWAITING_SUBMISSION_REVIEW.value,
-        target=RETURN_VALUE(AppStates.AWAITING_SUBMISSION_REVIEW.value, AppStates.AWAITING_PAYMENT.value))
+        target=RETURN_VALUE(AppStates.AWAITING_USER_SUBMISSIONS.value, AppStates.AWAITING_PAYMENT.value))
     def approve_submission(self):
         """Approve application submission"""
         return (AppStates.AWAITING_PAYMENT.value
-                if self.all_submissions_are_reviewed() else AppStates.AWAITING_SUBMISSION_REVIEW.value)
+                if self.is_ready_for_payment() else AppStates.AWAITING_USER_SUBMISSIONS.value)
 
     @transition(
         field=state,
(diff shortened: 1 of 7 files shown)


applications/models.py
```diff
@@ -143,6 +143,27 @@ def upload_resume(self, resume_file):
...
+        field=state,
+        source=AppStates.AWAITING_SUBMISSION_REVIEW.value,
+        target=AppStates.REJECTED.value
+    )
+    def reject_submission(self):
+        """Reject application submission"""
+
+    def has_incomplete_submissions(self):
+        """Check if the application is waiting for some submissions to be reviewed"""
+        return self.submissions.filter(review_status__isnull=True).exists()
```
gsidebo, line 165: This will only work if there is one application step defined for the bootcamp. If there are multiple steps and the user has just submitted something for one of the steps, we would want to set the state to "awaiting submissions" if that first submission is approved. It might be better to flip this function to return True if (a) the user has submitted something for all the steps, and (b) they are all approved.  `applications.api.derive_application_state` does this; not sure if that's something we can abstract into a separate API function.

annagav:
@gsidebo I addressed all your comments. Thank you!

applications/models.py
```diff
@@ -143,6 +143,32 @@ def upload_resume(self, resume_file):
...
+    )
+    def reject_submission(self):
+        """Reject application submission"""
+
+    def all_submissions_are_reviewed(self):
+        """
+        Check if the application is waiting for some submissions to be reviewed. This is used only if
+        application is awaiting submissions review
+        """
+        if self.state != AppStates.AWAITING_SUBMISSION_REVIEW.value:
```
gsidebo, line 168: A couple things:
1. This function still won't give us the right value if the user has fewer submissions than there are application steps. Example: Bootcamp 1 has 2 application steps: a video interview and a quiz. So far a user has submitted neither of those. Right now your function would return `True` for that user since they have no submissions with a null review status (because they have no submissions at all). [This function](https://github.com/mitodl/bootcamp-ecommerce/blob/master/applications/api.py#L53-L60) checks for the number of steps and compares to the number of submissions. If it looks like you can use any of that same logic here, it might be useful to pull that out into a separate API function.
1. I don't think it's necessary to check for the application state here.

annagav: In the issue it is specified that this api should be called only when the application is in "AWAITING_SUBMISSION_REVIEW` state. The application state can only be in `AWAITING_SUBMISSION_REVIEW` when all applications steps have submissions. That's why there is no reason to use the same logic as in that function, because the state of the application already implies that all the application steps have been submitted.

gsidebo: We can only move on to `AWAITING_PAYMENT` when the user has (a) submitted something for all of the required steps, and (b) all of those submissions are approved. In your logic for `approve_submission`, it considers (b) but not (a) when deciding to transition the state to `AWAITING_PAYMENT`.

gsidebo: A user can be in `AWAITING_SUBMISSION_REVIEW` multiple times. If the application has 2 steps, and the user submits something for the first step, approving that first submission should move them back to `AWAITING_USER_SUBMISSIONS` because they need to submit something for step 2.

gsidebo: (Sorry, I really should have documented all of this more clearly 😬)

annagav: thanks, for the explanation! I made a wrong assumption.

gsidebo: Perfectly reasonable assumption given the documentation, which right now is just a bunch of docstrings. I just need to create a simple readme

annagav:
@gsidebo I updated my PR. Thanks!

applications/models.py
```diff
@@ -143,6 +143,38 @@ def upload_resume(self, resume_file):
...
+        """
+        return self.submissions.count() == self.bootcamp_run.application_steps.count()
+
+    def is_ready_for_payment(self):
+        """
+        Make sure all steps are submitted, reviewed and approved
+        """
+        if self.all_application_steps_submitted():
+            return not self.submissions.filter(
+                models.Q(review_status=REVIEW_STATUS_REJECTED) | models.Q(review_status__isnull=True)).exists()
```
gsidebo, line 175: We can flip this logic and make the query a bit simpler

```python
if self.all_application_steps_submitted():
  return not self.submissions.exclude(review_status=REVIEW_STATUS_APPROVED).exists()
```

annagav:
@gsidebo I addressed you comment. Thanks!

PR #528, opened 2020-05-15
Add checkout data API

### LINKED ISSUE #492: API: Add API endpoint for fetching checkout data for an application

noisecapella:
Fixes #492 

#### What's this PR do?
Adds a new API `/api/checkout` to provide checkout data regarding applications, what they cost and what the user has already paid. There is an optional query parameter `application=<pk>` which will filter the list to just the application listed, if it exists and if it's awaiting payment.

Go to `/api/checkout` and verify that the data looks appropriate. If you add `?application=<pk>` to the url it should show only that application. Also, make sure nothing has changed in the payment page.

#### Context
This is intended to replace `/api/v0/bootcamps` in the near future, so there is a lot of overlapping functionality. In a future PR we will remove that API and migrate frontend logic to use this API instead.

### DIFF
--- a/ecommerce/serializers.py
+++ b/ecommerce/serializers.py
@@ -1,7 +1,11 @@
 """Serializers for ecommerce"""
+from decimal import Decimal
 from rest_framework import serializers
 
+from applications.models import BootcampApplication
+from ecommerce.api import get_total_paid
 from ecommerce.models import Order, Line
+from klasses.serializers import BootcampRunSerializer, InstallmentSerializer
 
 
 class PaymentSerializer(serializers.Serializer):
@@ -56,3 +60,49 @@ class LineSerializer(serializers.ModelSerializer):
             'price',
             'description',
         )
+
+
+class CheckoutDataSerializer(serializers.ModelSerializer):
+    """Serializer for ecommerce information for a BootcampApplication"""
+    bootcamp_run = BootcampRunSerializer()
+    total_price = serializers.SerializerMethodField()
+    total_paid = serializers.SerializerMethodField()
+    payments = serializers.SerializerMethodField()
+    installments = serializers.SerializerMethodField()
+
+    def get_total_price(self, application):
+        """The personal price for the user, or the full price for the run"""
+        return application.bootcamp_run.personal_price(self.context["request"].user) or Decimal(0)
+
+    def get_total_paid(self, application):
+        """The total paid by the user for this application so far"""
+        return get_total_paid(
+            user=self.context["request"].user,
+            application_id=application.id,
+            run_key=application.bootcamp_run.run_key,
+        )
+
+    def get_payments(self, application):
+        """Serialized payments made by the user"""
+        return LineSerializer(
+            Line.for_user_bootcamp_run(self.context["request"].user, application.bootcamp_run.run_key),
+            many=True,
+        ).data
+
+    def get_installments(self, application):
+        """Installments with prices and due dates"""
+        return InstallmentSerializer(
+            application.bootcamp_run.installment_set.order_by('deadline'),
+            many=True,
+        ).data
+
+    class Meta:
+        model = BootcampApplication
+        fields = [
+            "id",
+            "bootcamp_run",
+            "total_price",
+            "total_paid",
+            "payments",
+            "installments",
+        ]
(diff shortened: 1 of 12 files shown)


ecommerce/views.py
```diff
@@ -202,3 +205,27 @@ def get(self, request, username, *args, **kwargs):  # pylint: disable=unused-arg
         )
 
         return Response(serialize_user_bootcamp_runs(user=user))
+
+
+class CheckoutDataViewSet(ReadOnlyModelViewSet):
```
gsidebo, line 210: I know our existing endpoint returns all bootcamps that a user can pay for, but I'm pretty positive that we're not sticking with that approach. Users will only need to get this checkout data for a specific application, and you can probably just return a 404 if no application id was provided

ecommerce/serializers.py
```diff
@@ -56,3 +60,49 @@ class Meta:
...
+        """The total paid by the user for this application so far"""
+        return get_total_paid(
+            user=self.context["request"].user,
+            application_id=application.id,
+            run_key=application.bootcamp_run.run_key,
+        )
+
+    def get_payments(self, application):
+        """Serialized payments made by the user"""
+        return LineSerializer(
```
gsidebo, line 87: I think we need to move away from querying for `Line`s that have a matching key, and instead fetch the `Line` from the `application` foreign key in `Order`. Something like `LineSerializer(instance=[order.lines.first() for order in application.orders.all()`. If a user applies to some bootcamp run multiple times for whatever reason, we would only want to see the payments for the specific application in the request. On top of that, this approach would save us some queries, particularly if we prefetched on the initial queryset.

ecommerce/views.py
```diff
@@ -202,3 +205,27 @@ def get(self, request, username, *args, **kwargs):  # pylint: disable=unused-arg
...
+    """
+    authentication_classes = (
+        SessionAuthentication,
+    )
+    permission_classes = (
+        IsAuthenticated,
+    )
+    serializer_class = CheckoutDataSerializer
+
+    def get_queryset(self):
```
gsidebo, line 222: We should select/prefetch tables that we know we're going to need. Bootcamp run, installments, and orders/lines come to mind

noisecapella:
Review comments addressed, but there's a substantial amount of refactoring so you may want to take another pass

ecommerce/serializers.py
```diff
@@ -56,3 +58,37 @@ class Meta:
...
+    bootcamp_run = BootcampRunSerializer()
+    total_price = serializers.DecimalField(max_digits=20, decimal_places=2, read_only=True, source="price")
+    total_paid = serializers.DecimalField(max_digits=20, decimal_places=2, read_only=True)
+    payments = serializers.SerializerMethodField()
+    installments = serializers.SerializerMethodField()
+
+    def get_payments(self, application):
+        """Serialized payments made by the user"""
+        return LineSerializer(
+            (order.line_set.first() for order in application.orders.filter(status=Order.FULFILLED)),
```
gsidebo, line 74: If we change this to `(order.line_set.first() for order in application.orders.all() if order.status == Order.FULFILLED)`, that will take advantage of the prefetch. I think this will generate another query with that `filter()`.

gsidebo: Also I think we can prefetch `orders__line_set` in `get_queryset` instead of just `orders`

noisecapella: Sounds good

ecommerce/views.py
```diff
@@ -202,3 +205,33 @@ def get(self, request, username, *args, **kwargs):  # pylint: disable=unused-arg
         )
 
         return Response(serialize_user_bootcamp_runs(user=user))
+
+
+class CheckoutDataViewSet(ReadOnlyModelViewSet):
```
gsidebo, line 210: This view should just return a single item now, shouldn't it? It's returning the correct data for me, but it's in a single item list and I don't think it needs to be

noisecapella: Sure, just made the change

PR #534, opened 2020-05-18
Add API for available bootcamp runs

### LINKED ISSUE #447: API: Add endpoint that returns bootcamps that are available for application

noisecapella:
Fixes #447 

#### What's this PR do?
Creates an API at `/api/bootcampruns/` which shows all bootcamp runs to logged in users. A user can also add `?available=true` to the URL and see only runs which are available for enrollment, which are runs which haven't started yet and which aren't already enrolled in by the user.

Go to `/api/bootcampruns/` and `/api/bootcampruns/?available=true` and verify that the output makes sense. Create a `BootcampRunEnrollment` and note that the run is not in the available runs list anymore.

### DIFF
--- a/klasses/views.py
+++ b/klasses/views.py
@@ -1,0 +1,21 @@
+"""views for bootcamps"""
+from django.db.models.functions import Now
+from rest_framework.mixins import ListModelMixin
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.viewsets import GenericViewSet
+
+from klasses.serializers import BootcampRunSerializer
+from klasses.models import BootcampRun
+
+
+class BootcampViewSet(ListModelMixin, GenericViewSet):
+    """Viewset for bootcamps"""
+    permission_classes = (IsAuthenticated,)
+    serializer_class = BootcampRunSerializer
+
+    def get_queryset(self):
+        """Make a queryset which optionally shows what runs are available for enrollment"""
+        queryset = BootcampRun.objects.all().select_related("bootcamp")
+        if self.request.query_params.get("available") == "true":
+            queryset = queryset.filter(start_date__lt=Now()).exclude(enrollments__user=self.request.user)
+        return queryset.order_by("id")
--- a/klasses/views.py
+++ b/klasses/views.py
@@ -1,4 +1,5 @@
 """views for bootcamps"""
+from django.db.models import Q
 from django.db.models.functions import Now
 from rest_framework.mixins import ListModelMixin
 from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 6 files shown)


klasses/views.py
```diff
@@ -0,0 +1,21 @@
...
+class BootcampViewSet(ListModelMixin, GenericViewSet):
+    """Viewset for bootcamps"""
+    permission_classes = (IsAuthenticated,)
+    serializer_class = BootcampRunSerializer
+
+    def get_queryset(self):
+        """Make a queryset which optionally shows what runs are available for enrollment"""
+        queryset = BootcampRun.objects.all().select_related("bootcamp")
+        if self.request.query_params.get("available") == "true":
+            queryset = queryset.filter(start_date__lt=Now()).exclude(enrollments__user=self.request.user)
```
gsidebo, line 20: I think we'll also want to filter out bootcamps runs for which the user already has an application, active or complete/rejected. We might just want to remove the enrollments check from the query, but I'll leave that up to you

klasses/views_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+@pytest.fixture
+def runs(user):
+    """Make some test runs"""
+    run = BootcampRunFactory.create()
+    unavailable_run = BootcampRunFactory.create(start_date=now_in_utc()+timedelta(days=1))
+    already_enrolled_run = BootcampRunFactory.create()
+    BootcampRunEnrollment.objects.create(user=user, bootcamp_run=already_enrolled_run)
+
+    return run, unavailable_run, already_enrolled_run
```
gsidebo, line 28: Please use `SimpleNamespace` or something like that instead of returning tuples (https://github.com/mitodl/bootcamp-ecommerce/blob/master/applications/serializers_test.py#L35-L38)

klasses/views.py
noisecapella: I removed the enrollments check and added an application check in its place. It looks like all application states are either awaiting or complete so I think we can just check for the presence of an application.

PR #536, opened 2020-05-18
Fixed application list view to only return applications that belong to the logged-in user

gsidebo:
No ticket. Noticed during RC testing

#### What's this PR do?
See title. Previously the view was returning all applications, not just the ones owned by the logged-in user

Create an application for some user, and another belonging to a different user. The list endpoint should not return the latter application in the results

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -29,7 +29,7 @@ class BootcampApplicationViewset(
         if self.action == "retrieve":
             return BootcampApplication.objects.prefetch_state_data()
         else:
-            return BootcampApplication.objects.all()
+            return BootcampApplication.objects.filter(user=self.request.user).all()
 
     def get_serializer_class(self):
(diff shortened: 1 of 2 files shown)


applications/views.py
```diff
@@ -29,7 +29,7 @@ def get_queryset(self):
         if self.action == "retrieve":
             return BootcampApplication.objects.prefetch_state_data()
         else:
-            return BootcampApplication.objects.all()
+            return BootcampApplication.objects.filter(user=self.request.user).all()
```
noisecapella, line 32: Should we do this filtering for `retrieve` as well?

gsidebo: That's handled through the `main.permissions.UserIsOwnerPermission` permission class (the view returns a 403 if you're requesting an application that isn't yours)

PR #541, opened 2020-05-19
add site-wide notification

umarmughal824:
https://trello.com/c/p0Wds42c/135-site-wide-notifications

#### What's this PR do?
Add site wide notification in the cms

Add side notification and inspect the code in your page source.


gsidebo:
Hey Umar. I mentioned in Slack that we won't need this notification, but that wasn't accurate. We do want to be able to allow admins to create a site-wide notification in the CMS, but we do not want to use the xpro approach of rendering it in its own element outside of the React app. For now, let's limit this PR to putting a hidden notification in the Django template. That means you can get rid of all of the Javascript. The notification will be rendered in the React app. I'll create a separate issue for that

main/templates/partials/notifications.html
```diff
@@ -0,0 +1,8 @@
+<div class="notifications site-wide d-none">
+  {% if notification %}
+    <div class="notification" data-notification-id="{{ notification.id }}">
+      <a href="#" class="close-notification"></a>
```
gsidebo, line 4: We can get rid of this element. We'll only be using this element to extract the HTML and show it in the React component, which will have its own close button

main/templates/partials/notifications.html
```diff
@@ -0,0 +1,8 @@
+<div class="notifications site-wide d-none">
```
gsidebo, line 1: I know this is probably copied over from xpro, but I'm not sure we need this outer element. We only render a single notification (or none at all). We can probably just keep the element inside the `if` tag and move the `site-wide d-none` class names down to that element

main/templates/partials/notifications.html
umarmughal824: ok

main/templates/partials/notifications.html
umarmughal824: ok

PR #552, opened 2020-05-20
Update application state after Jobma webhook

### LINKED ISSUE #449: API: Add endpoint for uploading a video interview submission for a bootcamp application
An API endpoint should be created to support adding a video interview to a user's bootcamp application

Some technical details:

1. **The current `VideoInterviewSubmission` model currently has a single field – `video_file`. This field may need to be removed and other fields added. What we need for this model will probably be indicated by the third party app we use for video interviews (Jobma)**
1. A user should only be able to upload a video interview if (a) they don't already have a reviewed video interview for the current bootcamp run application, and (b) they have filled out their profile and uploaded a resume (in other words, their application is in the "awaiting submissions" state
1. When the interview is uploaded, the user's application state should transition to "awaiting review". See [django-fsm usage](https://github.com/viewflow/django-fsm#usage) for details on how to implement those transitions

noisecapella:
Fixes #449 

#### What's this PR do?
 - Updates application state after webhook from Jobma is received
 - Removes video_file from `VideoInterviewSubmission`, adds a foreign key to `Interview` in its place

Not sure there is a way to do this yet


applications/admin.py
```diff
@@ -181,7 +181,7 @@ class VideoInterviewSubmissionAdmin(SubmissionTypeAdmin):
     model = models.VideoInterviewSubmission
 
     def get_list_display(self, request):
-        return tuple(super().get_list_display(request) or ()) + ("video_file",)
+        return tuple(super().get_list_display(request) or ())
```
gsidebo, line 184: I think we would want to include at least the interview id in the list display here. Even better would be including a link to the interview record in admin

PR #559, opened 2020-05-21
Fix DEFAULT_FILE_STORAGE value for S3 backend

ahmed-belal:
N/A

#### What's this PR do?
Fixes the value of `DEFAULT_FILE_STORAGE` when S3 is in use as a backend. There was a typo in the module name string.

Try to upload media, it should serve as expected.


ahmed-belal:
@odlbot

PR #567, opened 2020-05-26
Get user info from API on payment page

### LINKED ISSUE #538: Authentication issue on redirect to payment page after cybersource payment
A similar issue cropped up with #416 & #426 but was fixed, it is now broken again.  Django seems to think the user is not authenticated when returning from cybersource.

mbertrand:
Fixes #538

#### What's this PR do?
- Removes `user` from js SETTINGS
- Use `state.entities.currentUser` on the `PaymentPage` instead

- Make a payment on the payment page, nothing should break.

### DIFF
--- a/static/js/components/Payment.js
+++ b/static/js/components/Payment.js
@@ -13,6 +13,7 @@ import {
 } from "../util/util"
 import type { UIState } from "../reducers/ui"
 import type { InputEvent } from "../flow/events"
+import type { HttpAuthResponse, User } from "../flow/authTypes"
 
 export const PAYMENT_CONFIRMATION_DIALOG = "paymentConfirmationDialog"
 const isVisible = R.propOr(false, PAYMENT_CONFIRMATION_DIALOG)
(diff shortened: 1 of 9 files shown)


static/js/components/Payment.js
```diff
@@ -28,7 +29,8 @@ export default class Payment extends React.Component<*, void> {
     sendPayment: () => void,
     setPaymentAmount: (event: InputEvent) => void,
     setSelectedBootcampRunKey: (event: InputEvent) => void,
-    showDialog: (dialogKey: string) => void
+    showDialog: (dialogKey: string) => void,
+    currentUser: () => Promise<HttpAuthResponse<User>>
```
noisecapella, line 33: I think this should just be `?CurrentUser` since it's there's no `Promise` being resolved and the code below is using the object directly and not via `.body`

static/js/pages/PaymentPage.js
```diff
@@ -47,7 +49,8 @@ type Props = {
   setSelectedBootcampRunKey: (runKey: number) => void,
   setTimeoutActive: (active: boolean) => void,
   setToastMessage: (payload: any) => void,
-  showDialog: (dialogKey: string) => void
+  showDialog: (dialogKey: string) => void,
+  currentUser: () => Promise<HttpAuthResponse<User>>
```
noisecapella, line 53: Should this be `?CurrentUser`?

mbertrand: Thanks, changes made.

PR #577, opened 2020-05-28
Add Linkedin option for resume

### LINKED ISSUE #529: API: Add linkedin option for resume
Related to https://github.com/mitodl/bootcamp-ecommerce/issues/448 
[Design](https://projects.invisionapp.com/d/main#/console/19928472/417192803/preview)

We want to accept a LinkedIn profile URL in addition to a resume file. The user can provide one or both.

#### Acceptance criteria
- URL field added to the `BootcampApplication` model to store a linkedin url
- `resume_upload_date` updated if a user sets that linkedin url (we might want to rename the field too)
- API endpoint added to accept a linkedin URL for an application (or the existing resume endpoint updated to also accept a linkedin URL)
- `BootcampApplication` is transitioned from "awaiting resume" to "awaiting submissions" if the user has added a resume _or_ a linkedin URL (one or both)

annagav:
Fix #529

#### What's this PR do?
Add a new field to the BootcampApplication model, process the resume or linkedin url and transition the application to the next stage.

Create a bootcamp application, set the state to awaiting resume, try sending a post request with a `linkedin_url`
This should update the corresponding field on the application model, update the `resume_upload_date` field, and transition the application to next stage.


applications/api.py
```diff
@@ -98,7 +98,10 @@ def process_upload_resume(resume_file, bootcamp_application):
     """
     if bootcamp_application.state == AppStates.AWAITING_PROFILE_COMPLETION.value:
         raise InvalidApplicationException("The BootcampApplication is still awaiting profile completion")
-    bootcamp_application.upload_resume(resume_file)
+    if resume_file:
+        bootcamp_application.upload_resume(resume_file)
```
gsidebo, line 102: A user can provide either or both of the resume file and linkedin URL, and we only want to change the application state once if they provide both, so I think it makes the most sense to combine this into a single method that accepts both items as kwargs. Signature could be something like `add_work_history(self, *, resume_file=None, linkedin_url=None)`

applications/api_test.py
```diff
@@ -138,6 +138,24 @@ def test_get_or_create_bootcamp_application(mocker):
     assert created is False
 
 
+@pytest.mark.django_db
+@pytest.mark.parametrize("has_resume,has_linkedin,expected", [
+    (True, False, AppStates.AWAITING_USER_SUBMISSIONS.value),
+    (True, True, AppStates.AWAITING_USER_SUBMISSIONS.value)
```
gsidebo, line 144: You don't need the `AppStates.AWAITING_USER_SUBMISSIONS.value` in the test parameters since you're expecting the same value in all cases

applications/views.py
```diff
@@ -94,3 +94,27 @@ def patch(self, request, *args, **kwargs):
...
+    """
+    View for uploading resume and linkedin link
+    """
+    authentication_classes = (SessionAuthentication,)
+    permission_classes = (IsAuthenticated, UserIsOwnerPermission,)
+    lookup_field = "pk"
+    owner_field = "user"
+    queryset = BootcampApplication.objects.all()
+
+    def post(self, request, *args, **kwargs):
```
gsidebo, line 109: I think the request should be a `PATCH` rather than a `POST`

applications/views.py
```diff
@@ -94,3 +94,27 @@ def patch(self, request, *args, **kwargs):
...
+        """
+        Update the application with resume
```
gsidebo, line 111:
```suggestion
        Update the application with resume and/or linkedin URL
```

applications/views.py
```diff
@@ -94,3 +94,27 @@ def patch(self, request, *args, **kwargs):
...
+        """
+        application = self.get_object()
+        linkedin_url = request.data.get("linkedin_url", False)
+        resume_file = request.FILES.get('file', False)
+        if not (linkedin_url or resume_file):
```
gsidebo, line 116: Nit:
```suggestion
        linkedin_url = request.data.get("linkedin_url")
        resume_file = request.FILES.get('file')
        if linkedin_url is None and resume_file is None:
```

applications/api.py
```diff
@@ -91,20 +91,21 @@ def get_required_submission_type(application):
     )
 
 
-def process_upload_resume(resume_file, bootcamp_application):
+def process_upload_resume(resume_file, linkedin_url, bootcamp_application):
```
gsidebo, line 94: Looking at this again, I'm leaning toward getting rid of this API function and moving this logic into the view itself. It's just doing a single bit of validation (which I'm pretty sure is redundant), calling a model method, and saving. Until we have a another use case for running this same logic, I don't think we get much out of splitting this into a helper function. This would simplify tests for us as well

applications/models.py
```diff
@@ -167,10 +168,14 @@ def complete(self):
         ],
         target=AppStates.AWAITING_USER_SUBMISSIONS.value,
     )
-    def upload_resume(self, resume_file):
+    def add_resume(self, *, resume_file=None, linkedin_url=None):
         """Save resume and make sure that the state can be transitioned to a new state"""
-        validate_file_extension(resume_file)
-        self.resume_file = resume_file
+        if resume_file:
```
gsidebo, line 173: Please update the docstring: `Add resume and/or linkedin URL to the application`

applications/views.py
annagav: in the django documentation it said that for uploading a file is has to be POST resquest.

annagav:
@gsidebo I addressed all your comments. Thank you!

PR #611, opened 2020-06-02
Add application submissions admin api

### LINKED ISSUE #462: Admin API: Add endpoint to fetch bootcamp application submissions that need review
[Wireframe](https://projects.invisionapp.com/share/3HWYOFP6R8B#/screens)

Technical details:
- Ideally this endpoint should accept the filters indicated by the wireframe: bootcamps, bootcamp runs, and review status. It would also be ideal to allow multiple values for each filter. It may end up making more sense to implement this in a separate issue, though.
- There should be one row for each `ApplicationStepSubmission`.

rhysyngsun:
Closes #462 

#### What's this PR do?
Adds an API to view bootcamp submissions:
- Submissions can be filtered by review status and/or the bootcamp
- The response can be sorted by the created_data (asc or desc) of the bootcamp submission

- Go to `/api/submissions/`, you should see existing application submission steps listed there.
- You'll see a "Filters" button in the top right of the browseable API UI that will let you test out the filtering and sorting capabilities added here.
- You should see "facet" data in the response that updates as you filter results..
- The detail `/api/submissions/:id/` API should return a single result


applications/api.py
```diff
@@ -60,7 +58,9 @@ def derive_application_state(
     ]
     if any([status == REVIEW_STATUS_REJECTED for status in submission_review_statuses]):
         return AppStates.REJECTED.value
-    elif any([status is None for status in submission_review_statuses]):
+    elif any(
+        [status == REVIEW_STATUS_PENDING for status in submission_review_statuses]
+    ):
```
mbertrand, line 63: I think there will only be 1 video submission required at first, but wondering if this should be handled differently at all in the future if/when multiple submissions are required.  For instance, if 3 submissions are required and 2 have been submitted and are pending, is the correct application state `AWAITING_SUBMISSION_REVIEW` or `AWAITING_USER_SUBMISSIONS`?  I'm fine with kicking this can down the road though.

applications/serializers.py
```diff
@@ -94,3 +98,46 @@ class BootcampApplicationSerializer(serializers.ModelSerializer):
...
         fields = ["id", "state", "created_on", "bootcamp_run"]
+
+
+class SubmissionReviewSerializer(SubmissionSerializer):
+    """ApplicationStepSubmission serializer for reviewers"""
+
+    bootcamp_application = BootcampApplicationSerializer(required=False, read_only=True)
+    learner = UserSerializer(
+        source="bootcamp_application.user", required=False, read_only=True
+    )
```
mbertrand, line 109: Including the entire Learner serializer here maybe returns more data for the submissions list than necessary, but will save an additional API call on the submission detail UI to get that info.  With paging set to 10 the response size shouldn't be an issue though.

applications/views_test.py
```diff
@@ -122,29 +125,201 @@ def test_app_detail_view_permissions(client):
...
+                for submission in sorted(submissions, key=lambda s: s.review_status)
+            ],
+        },
+    }
+
+
+@pytest.mark.django_db
+def test_review_submission_list_query_bootcamp_id(admin_drf_client):
+    """
+    The review submission list view should return a list of submissions
```
mbertrand, line 200: "should return a list of submissions filtered by bootcamp id"

applications/views_test.py
```diff
@@ -122,29 +125,201 @@ def test_app_detail_view_permissions(client):
...
+            ],
+        },
+    }
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("review_status", ALL_REVIEW_STATUSES)
+def test_review_submission_list_query_review_status(admin_drf_client, review_status):
+    """
+    The review submission list view should return a list of submissions
```
mbertrand, line 232: "filtered by review status"

applications/views_test.py
```diff
@@ -122,29 +125,201 @@ def test_app_detail_view_permissions(client):
...
+        [REVIEW_STATUS_APPROVED, REVIEW_STATUS_REJECTED],
+        [REVIEW_STATUS_REJECTED, REVIEW_STATUS_PENDING],
+        [REVIEW_STATUS_REJECTED, REVIEW_STATUS_APPROVED, REVIEW_STATUS_PENDING],
+    ],
+)
+def test_review_submission_list_query_review_status_in(
+    admin_drf_client, review_statuses
+):
+    """
+    The review submission list view should return a list of all submissions
```
mbertrand, line 281: "filtered by multiple review statuses"

applications/api.py
rhysyngsun: I only touched this code because of the migration from a `None` value to the constant `REVIEW_STATUS_PENDING`, so I think this is something for another issue if it's something that needs to be done.

applications/serializers.py
rhysyngsun: yeah, I originally tried to pull in https://django-rest-framework-serializer-extensions.readthedocs.io/ to make this cleaner, but that ended being a tug of war between having it work for both reads here and writes elsewhere in the app. Decided to take the performance hit for the sake of not spending more time on it... :slightly_frowning_face:

PR #614, opened 2020-06-03
add `repl.py` from open-discussions

aliceriot:
none

#### What's this PR do?

I just noticed we didn't have the `repl.py` script over here and thought we might want it.

you should be able to run the script in the web container and it should open a django shell and automatically import a bunch of models for you, like `User` and so on.

I just copied the file from open-discussions without changing anything.


rhysyngsun:
Could you `chmod +x repl.py` it? Otherwise looks good

PR #618, opened 2020-06-03
Payment drawer

### LINKED ISSUE #572: Add payment UI

noisecapella:
Fixes #572 

#### What's this PR do?
Adds `PaymentDisplay` to allow people to pay from the application page. Some API changes were also made to do this.

First, set up Cybersource environment variables. I can send you the variables for testing via keybase. Go to `/application` and expand one of the applications. Click 'Make a Payment'. Type in a number in the textbox and click 'Pay Now'. You should be directed to Cybersource.

Note: Once you complete a payment it will redirect to `/pay/`. I'm not sure what should happen at this point so I left the behavior as is.


applications/models.py
```diff
@@ -90,9 +90,17 @@ class BootcampApplicationQuerySet(models.QuerySet):
...
 
     def prefetch_state_data(self):
-        """Prefetches models that inform the state of bootcamp applications"""
-        return self.select_related("user__profile").prefetch_related(
-            "submissions", "orders", "bootcamp_run__application_steps__application_step"
+        """
+        Prefetches models that inform the state of bootcamp applications,
+        and filters to only include fulfilled orders.
+        """
+        return self.select_related("user__profile", "bootcamp_run").prefetch_related(
```
noisecapella, line 97: `bootcamp_run` is added here because it improves performance. From [the docs](https://docs.djangoproject.com/en/3.0/ref/models/querysets/), specifically this line: "Of course, the best_pizza relationship could also be fetched using select_related to reduce the query count to 2:"

static/scss/ecommerce.scss
```diff
@@ -0,0 +1,70 @@
+.payment-drawer {
+  max-width: 600px;
+  padding: 25px;
+
+  .drawer-title {
```
gsidebo, line 5: Can we make an effort to reuse existing classes/definitions, and if they don't exist, create new ones that are likely to be reused? Examples:
- The `h2` definition in `_shared.scss` is very close to this `.drawer-title` definition. This has wider letter spacing, but I'm willing to bet it won't matter if we use that already-defined letter spacing.
- `.bootcamp` and `.payment-amount` have the same `font-size` and `letter-spacing` .
- `button` probably has some styles that we would want to share with other buttons. Maybe the designs show a slight discrepancy with this button and other buttons, but I think our approach should be to use the same styling if any of these elements look moderately similar. If the discrepancies really were intentional, Abdul can let us know before we go live. For now I think it's much more likely that we're going to have barely-controlled expansion of css without a real need for it

noisecapella:
@annagav FYI I just committed a change to parse the amount in the textbox and only submit if it is a valid non-zero number

static/scss/ecommerce.scss
noisecapella: I made some changes. Let me know if you spot other places we can consolidate CSS. I'm not really sure what the threshold is for new CSS vs reusing slightly different CSS. 
Screenshot from 2020-06-04 16-10-57

noisecapella:
@annagav FYI another minor commit to make sure we use the personal price if it exists

annagav:
I am getting this error: 
``` File "./hubspot/serializers.py", line 82, in to_representation
web_1     |     if amount_paid >= instance.bootcamp_run.personal_price(instance.user):
web_1     | TypeError: '>=' not supported between instances of 'decimal.Decimal' and 'NoneType'``

File: static/js/components/PaymentDisplay.js [hunk 3]
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  const run = application.bootcamp_run
+  const totalSpent = sum(
+    application.orders.map(order => order.total_price_paid)
+  )
+  const cost = application.price
+
+  return (
+    <div className="container">
```
gsidebo, line 39: Any reason you're not using Formik here? We need validation messages for an empty field, a non-number value, and an invalid amount at the very least. We're already using Formik elsewhere

noisecapella:
@annagav I think you need to create at least one `Installment` for the `BootcampRun`.

static/js/components/PaymentDisplay.js
noisecapella: No reason, I'll look into it

annagav:
@noisecapella Am I supposed to see this when I click the Pay Now button?
Screen Shot 2020-06-07 at 10 21 34 PM

noisecapella:
@annagav can you double check your Cybersource environment variables? Also, what is the amount in the textbox when you click "Pay Now"?

annagav:
The entered the amount that was determined by the installment  I created, is that correct?

noisecapella:
@annagav Yes it should be. As long as it's a valid number it should work.

When you set the environment variables, did you restart the docker container afterwards?

annagav:
Yes, I did restart the docker container.

noisecapella:
@annagav I think I found and fixed the bug... can you `git pull` and try again?

annagav:
Functionality looks good! I'll take another look once the tests are passing.

gsidebo:
@noisecapella 

> Note: Once you complete a payment it will redirect to /pay/. I'm not sure what should happen at this point so I left the behavior as is.

If this is still true, and the change is easy enough, can you just change that to redirect to the application dashboard (`/applications/`)? The `/pay/` page is not going to exist anymore, and it's going to add unnecessary difficulty to functional testing if we leave that as-is.

noisecapella:
@gsidebo Already made the change to `applications`. I'd prefer to leave the drawer actions as is just because that code is already written

noisecapella:
@annagav Tests should be passing now :crossed_fingers:

PR #622, opened 2020-06-04
asadiqbal08/ProgramElements Under Home Page

asadiqbal08:
[Ticket](https://trello.com/c/ospegPH6/165-home-page-program-elements)

#### What's this PR do?
Introduce Program Elements in HomePage

Add child of `Program Description` under `HomePage`. 

screencapture-bc-odl-local-8099-homepage-2020-06-04-15_08_13

screencapture-bc-odl-local-8099-homepage-2020-06-04-15_09_10


cms/migrations/0015_add_banner_image.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 2.2.10 on 2020-06-04 12:30
```
umarmughal824, line 1: NIT: better to rename that file with `0015_add_banner_image_in_programdescriptionpage` or so, so that it could be understood better and quicker by just viewing the file name. As of now, it does not speak for where we are adding that field.

cms/models.py
```diff
@@ -54,10 +80,13 @@ def get_context(self, request, *args, **kwargs):
             "js_settings_json": json.dumps(_serialize_js_settings(request)),
             "site_name": settings.SITE_NAME,
             "title": self.title,
+            "program_description_section": self.program_description_section,
         }
 
+    subpage_types = ["ProgramDescriptionPage"]
```
umarmughal824, line 86: why not `ThreeColumnImageTextPage ` and `LearningResourcePage ` are provided as the subpage_types in it?

umarmughal824: Please add these two in subpage_types.

asadiqbal08: Not need for this PR. When we will work over these two pages for `HomePage` then add accordingly.

asadiqbal08:
sure. renamed the migration file.

umarmughal824:
@asadiqbal08 could you please address that deployment error, before I do the final review.

asadiqbal08:
There was some migration issue during CI deployment that I fixed manually, Now you can create `Program Elements` under HomePage over CI

PR #625, opened 2020-06-05
Finished application detail UI in dashboard

### LINKED ISSUE #558: Add detailed application state UI
**DEPENDS ON #523**

[Design](https://projects.invisionapp.com/d/main#/console/19928472/417192801/preview)

#### Acceptance criteria
- When a card for a bootcamp application is expanded in the dashboard (#523), a detail UI is revealed that matches the design
  - When a card is first expanded, the application detail API is called to fetch the details for the expanded view (#450)
- Expand/collapse works for each card

gsidebo:
Closes #558 

#### What's this PR do?
Finishes the detail UI in the application dashboard ([design](https://projects.invisionapp.com/d/main#/console/19928472/418545905/preview))

Open the dashboard, expand the detail section, and experiment with updating your application to different states

If we _really_ want the gradient depicted in the progress bar, we can implement that later. 

fireshot_2020-06-05
ss 2020-06-05 at 12 19 55

### DIFF
--- a/static/js/components/Drawer.js
+++ b/static/js/components/Drawer.js
@@ -4,6 +4,7 @@ import * as React from "react"
 import { useDispatch, useSelector } from "react-redux"
 import { Drawer as RMWCDrawer, DrawerContent } from "@rmwc/drawer"
 import { Theme } from "@rmwc/theme"
+import * as R from "ramda"
 
 import { drawerSelector } from "../lib/selectors"
 import { setDrawerOpen } from "../reducers/drawer"
@@ -13,21 +14,24 @@ import EditProfileDisplay from "./EditProfileDisplay"
 import ViewProfileDisplay from "./ViewProfileDisplay"
 import PaymentDisplay from "./PaymentDisplay"
 
-const renderDrawerContents = (drawerState: string): ?React$Element<*> => {
+const renderDrawerContents = (
+  drawerState: string,
+  drawerMeta?: Object
+): ?React$Element<*> => {
   switch (drawerState) {
   case PROFILE_EDIT:
     return <EditProfileDisplay />
   case PROFILE_VIEW:
     return <ViewProfileDisplay />
   case PAYMENT:
-    return <PaymentDisplay />
+    return <PaymentDisplay application={R.prop("application", drawerMeta)} />
   default:
     return null
   }
 }
 
 export default function Drawer() {
-  const { drawerOpen, drawerState } = useSelector(drawerSelector)
+  const { drawerOpen, drawerState, drawerMeta } = useSelector(drawerSelector)
 
   const dispatch = useDispatch()
   const closeDrawer = useCallback(() => {
@@ -38,7 +42,7 @@ export default function Drawer() {
     <Theme>
       <RMWCDrawer open={drawerOpen} onClose={closeDrawer} dir="rtl" modal>
         <DrawerContent dir="ltr">
-          {renderDrawerContents(drawerState)}
+          {renderDrawerContents(drawerState, drawerMeta)}
         </DrawerContent>
       </RMWCDrawer>
     </Theme>
--- a/static/js/components/applications/ProgressDetailRow.js
+++ b/static/js/components/applications/ProgressDetailRow.js
@@ -1,0 +1,40 @@
+// @flow
+/* global SETTINGS: false */
+import React from "react"
+
+type DetailRowProps = {
+  className: string,
+  fulfilled: boolean,
+  children?: React$Node
+}
+
+/*
+ * Renders a section of the application progress bar (with a checkbox if this step is complete) along with the
+ * detail content of the application step.
+ */
+export default function ProgressDetailRow(
+  props: DetailRowProps
+): React$Element<*> {
+  const { children, className, fulfilled } = props
+
+  return (
+    <div
+      className={`detail-section ${className}${fulfilled ? " fulfilled" : ""}`}
+    >
+      <div className="d-flex">
+        <div className="progress-meter">
+          <div className="top-line" />
+          <div className="check">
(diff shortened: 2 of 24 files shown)


rhysyngsun:
@gsidebo you have a conflict, could you rebase?

gsidebo:
@rhysyngsun Ready for review

static/js/components/Drawer.js
```diff
@@ -13,21 +14,26 @@ import EditProfileDisplay from "./EditProfileDisplay"
...
+  let props
   switch (drawerState) {
   case PROFILE_EDIT:
     return <EditProfileDisplay />
   case PROFILE_VIEW:
     return <ViewProfileDisplay />
   case PAYMENT:
-    return <PaymentDisplay />
+    props = { application: R.prop("application", drawerMeta) }
+    return <PaymentDisplay {...props} />
```
rhysyngsun, line 29: Can simplify this down to:
```suggestion
    return <PaymentDisplay application={R.prop("application", drawerMeta)} />
```

static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -72,57 +78,277 @@ export class ApplicationDashboardPage extends React.Component<Props, State> {
...
-      allApplicationDetail,
-      setDrawerOpen,
-      setDrawerState,
-      setDrawerMeta
-    } = this.props
+  /*
+   * Returns an object mapping an application step id to the user's submission (if anything was submitted
+   * for that step).
+   */
+  getAppStepSubmissions = (
```
rhysyngsun, line 85: It would be good to refactor this out to a library/util method since it doesn't access anything on the class.

static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -72,57 +78,277 @@ export class ApplicationDashboardPage extends React.Component<Props, State> {
...
+    )
 
-    if (!allApplicationDetail || !allApplicationDetail[String(applicationId)]) {
-      return null
-    }
+  /*
+   * Renders a section of the application progress bar (with a checkbox if this step is complete) along with the
+   * detail content of the application step.
+   */
+  renderDetailRow = (
```
rhysyngsun, line 103: General feedback that this class is doing too much, each of the `renderX` method should be their own independent components. It'll make the class less complex and easier to test.

static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -72,57 +78,277 @@ export class ApplicationDashboardPage extends React.Component<Props, State> {
...
+              "review",
+              submissionApproved
+            )}
+          </React.Fragment>
+        )
+      }
+    )
+
+    const paymentRow = this.renderDetailRow(
+      <React.Fragment>
```
rhysyngsun, line 320: Passing a fragment into a function is a bit of an antipattern here, implementing `renderDetailRow` as an HOC and then wrapping this fragment (which should be its own component too) with that HOC as a concrete component will make this a lot cleaner and easier to reason about.

static/js/reducers/drawer.js
```diff
@@ -3,19 +3,37 @@ import { createAction, createReducer } from "@reduxjs/toolkit"
...
-    [setDrawerMeta]: (state, action) => {
-      state.drawerMeta = action.payload
+    [openDrawer]: (
+      state: DrawerState,
+      action: { payload: DrawerChangePayload }
+    ) => {
+      state.drawerState = action.payload.type
+      if (action.payload.meta) {
+        state.drawerMeta = action.payload.meta
+      }
```
rhysyngsun, line 35: I think you probably want to reset `drawerMeta` to `{}` if its not passed so you don't accidentally leak state from a previous drawer open into another one.

static/js/pages/applications/ApplicationDashboardPage.js
gsidebo: Yeah, looking at it with fresh eyes this definitely got away from me

static/js/reducers/drawer.js
gsidebo: Great catch

gsidebo:
@rhysyngsun Much cleaner now. Not sure how I forgot that I could create components 🤷‍♂️ 

Ready for review again

static/js/components/applications/ProgressDetailRow.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+export default function ProgressDetailRow(
+  props: DetailRowProps
+): React$Element<*> {
+  const { children, className, fulfilled } = props
+
+  return (
+    <div
+      className={`detail-section ${className}${fulfilled ? " fulfilled" : ""}`}
+      key={className}
```
rhysyngsun, line 23: I don't think `key` here does anything since there's no siblings

static/js/components/applications/detail_sections.js
```diff
@@ -0,0 +1,206 @@
...
+  ready: boolean,
+  fulfilled: boolean,
+  openDrawer: (actionPayload: DrawerChangePayload) => void
+}
+
+type ProfileDetailProps = DetailSectionProps & {
+  user: User
+}
+
+export const ProfileDetail = (props: ProfileDetailProps): React$Element<*> => {
```
rhysyngsun, line 28: I think each of these components should be in their own source files for consistency

gsidebo: They're all very small components, they're all very closely related, and it's somewhat likely that a change to one would need to be applied to all. I know the normal practice is 1 component 1 file, but given what I mentioned + the fact that this would result in 12 separate files when you include test files, I'd prefer to keep it as-is if you don't feel that strongly about it.

Issue #641, opened 2020-06-11
UI: Specify success, error, and loading behaviors for pages that make HTTP requests


gsidebo:
We have a bunch of UI elements that make HTTP requests either upon page load or in response to a button click et. al. We need to decide on what happens when those requests fail, and when certain requests succeed.

#### Specific questions to answer:
- For a page/component that makes a request for data as soon as it loads or after a button click, what happens if that request fails? Examples: the application dashboard page loading a user's applications, the new application form loading all available bootcamp runs, the admin review list page loading all user submissions that need review.
  - Option(s) - We could have show a standard error component in the body of each page with some meaningful text. This would be shown instead of the normal page contents. We could also use a toast message.
- What should happen in general when a request to create/update something succeeds? Examples: creating a new application via the form, successfully completing a payment, submitting a review for an application in the admin review page.
  - Option(s) - This seems like a good use case for a toast message. Regardless of the UI element we use, I assume that a new application/successful payment should simply close the drawer and remain on the application dashboard, and the review should send the user back to the review list page.
- What should happen when a request to create/update something fails? Same examples as above.
  - Option(s) - Show a simple error message near the submit button and keep the user on the same page. We could also use a toast.
- Should we have any standard UI effect when a request to create/update something is in progress? 
  - Option(s) - Replace button contents with a spinner, replace the entire button with a spinner, put a spinner next to the button, simply set the button to disabled and grey it out
- Should we have any standard UI effect when a request to load some data is in progress? Examples: loading a page of submissions in the admin review list page, loading application details when the "expand" button is clicked in the application dashboard

abdulkdawson:
We have a bunch of UI elements that make HTTP requests either upon page load or in response to a button click et. al. We need to decide on what happens when those requests fail, and when certain requests succeed.

#### Specific questions to answer:
- For a page/component that makes a request for data as soon as it loads or after a button click, what happens if that request fails? Examples: the application dashboard page loading a user's applications, the new application form loading all available bootcamp runs, the admin review list page loading all user submissions that need review.
  - Option(s) - We could have show a standard error component in the body of each page with some meaningful text. This would be shown instead of the normal page contents. We could also use a toast message.

**I think a toast message here would work.**

- What should happen in general when a request to create/update something succeeds? Examples: creating a new application via the form, successfully completing a payment, submitting a review for an application in the admin review page.

**We could also apply a toast message with success or confirmation text.**

[...]

PR #620, opened 2020-06-03
Mb/submissiondetail

mbertrand:
#### What are the relevant tickets?
Closes #550 

#### What's this PR do?
(Required)

#### How should this be manually tested?
(Required)

#### Where should the reviewer start?
(Optional)

#### Any background context you want to provide?
(Optional)

#### Screenshots (if appropriate)
(Optional)

#### What GIF best describes this PR or how it makes you feel?
(Optional)

PR #652, opened 2020-06-15
Resume Drawer: upload file or linkedin url

### LINKED ISSUE #548: Add resume upload UI
**DEPENDS ON #529**

[Design](https://projects.invisionapp.com/d/main#/console/19928472/417192803/preview)

#### Acceptance criteria
- Component(s) and form added to React to upload a resume and/or add a linkedin profile URL
  - UI matches the design
  - Either or both inputs can be provided, and an error message is shown if neither one is submitted.
  - Upon successful submission, the user's application is updated in the database
- Form is responsive so that it will render correctly within a drawer

#### Nice to have
- Drag and drop for the resume file. We have components in other projects that allow for an image drop and drop, but if for whatever reason this is too complicated for the resume use case, it's more important to just get file upload working in some capacity.
- Validation on the front end for valid file resume types (like we already have on the back end)

annagav:
Fix #548

#### What's this PR do?
Implement a Resume Upload Drawer where user can upload a file or add linkedin url to their application.

Screen Shot 2020-06-16 at 1 57 31 PM

### DIFF
--- a/applications/utils.py
+++ b/applications/utils.py
@@ -13,6 +13,6 @@ def validate_file_extension(value):
         value (File): file name
     """
     ext = os.path.splitext(value.name)[1]
-    valid_extensions = [".pdf", ".doc", ".docx", "odt"]
+    valid_extensions = [".pdf", ".doc", ".docx", ".odt"]
     if not ext.lower() in valid_extensions:
         raise ValidationError("Unsupported file extension.")
--- a/static/js/components/UploadResumeView.js
+++ b/static/js/components/UploadResumeView.js
@@ -1,0 +1,154 @@
+// @flow
+/* global SETTINGS: false */
+import React from "react"
+import Dropzone from "react-dropzone-uploader"
+import { compose } from "redux"
+import { connect } from "react-redux"
+import { mutateAsync } from "redux-query"
+import { values, join, keys } from "ramda"
+
+import applications from "../lib/queries/applications"
+import { ErrorMessage, Field, Form, Formik } from "formik"
+import FormError from "./forms/elements/FormError"
+import { DEFAULT_POST_OPTIONS } from "../lib/redux_query"
+import { ALLOWED_FILE_EXTENTIONS } from "../constants"
+import type { ApplicationDetail } from "../flow/applicationTypes"
+
+const doneBtnStyle = {
+  color:           "#fff",
+  backgroundColor: "#dc3545",
+  borderColor:     "#d43f3a"
+}
+
+type UploaderProps = {
+  application: ApplicationDetail
+}
+export const ResumeUploader = (props: UploaderProps) => {
+  const { application } = props
+
+  const getUploadParams = () => {
+    return {
+      url: `/api/applications/${application.id}/resume/`,
+      ...DEFAULT_POST_OPTIONS
+    }
+  }
+
+  const handleSubmit = (files, allFiles) => {
+    allFiles.forEach(f => f.remove())
+  }
+
+  const inputContent = () => {
+    return (
+      <div className="input-label">
+        <h3>Drag file here or click to Browse</h3>
+        {application.resume_filepath ? (
+          <React.Fragment>
+            <p className="uploaded-resume">
+              Your uploaded resume:{" "}
+              <strong>
+                {application.resume_filepath.substr(
+                  application.resume_filepath.indexOf("_") + 1
+                )}
+              </strong>
+            </p>
+            <p>This will replace your existing resume.</p>
+          </React.Fragment>
+        ) : null}
+        <p>Allowed file types: {join(",", keys(ALLOWED_FILE_EXTENTIONS))}</p>
+      </div>
+    )
+  }
+
+  return (
+    <Dropzone
+      getUploadParams={getUploadParams}
+      inputContent={inputContent}
+      onSubmit={handleSubmit}
+      submitButtonContent="Done"
+      classNames="drop-outline"
+      maxFiles={1}
+      styles={{
+        dropzone:     { minHeight: 200, maxHeight: 250 },
+        submitButton: doneBtnStyle
+      }}
+      accept={join(",", values(ALLOWED_FILE_EXTENTIONS))}
+    />
+  )
+}
+type Props = {
+  application: ?ApplicationDetail,
+  uploadResume: (linkedinUrl: string, applicationId: number) => Promise<void>
+}
+
+export class UploadResumeView extends React.Component<Props> {
+  render() {
+    const { uploadResume, application } = this.props
+
+    if (!application) {
+      return null
+    }
+    return (
+      <div className="container">
+        <div className="drawer-wrapper">
+          <h2 className="drawer-title">Resume or LinkedIn Profile</h2>
+          <p>Please upload resume or a link to your LinkedIn profile</p>
+          <h3>Resume</h3>
+          <ResumeUploader application={application} />
+
+          <h3 className="margin-top">LinkedIn URL</h3>
+          <Formik
+            onSubmit={async ({ linkedinUrl }, actions) => {
+              const {
+                body: errors
+              }: // $FlowFixMe
+              { body: Object } = await uploadResume(linkedinUrl, application.id)
+
+              if (errors && errors.length > 0) {
+                actions.setErrors({
+                  linkedinUrl: errors
+                })
+              }
+              actions.setSubmitting(false)
+            }}
+            initialValues={{
+              linkedinUrl: application.linkedin_url || null
+            }}
+            render={({ isSubmitting }) => (
+              <Form>
+                <div className="form-group">
+                  <Field
+                    name="linkedinUrl"
+                    className="form-control"
+                    placeholder="LinkedIn URL"
+                  />
+                  <p className="example-input">
+                    Example: https://www.linkedin.com/in/User-Name-7472b48/
+                  </p>
+                  <ErrorMessage name="linkedinUrl" component={FormError} />
+                </div>
+                <button
+                  className="btn btn-danger"
+                  type="submit"
+                  disabled={isSubmitting}
+                >
+                  Submit
+                </button>
+              </Form>
+            )}
+          />
+        </div>
+      </div>
+    )
+  }
+}
+
+const uploadResume = (linkedinUrl: string, applicationId: number) =>
+  mutateAsync(
+    applications.applicationUploadResumeMutation(linkedinUrl, applicationId)
+  )
+
+const mapDispatchToProps = {
+  uploadResume: uploadResume
+}
+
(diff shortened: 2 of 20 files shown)


static/js/components/UploadResumeView.js
```diff
@@ -0,0 +1,153 @@
...
+                {
+                  resume_filename: file,
+                  linkedin_url:    linkedin_url
+                },
+                application.id
+              )
+
+              if (errors && errors.length > 0) {
+                setErrors({
+                  linkedin_url: errors
```
mbertrand, line 96: Not seeing error messages when something goes wrong (for example when the application is in the wrong state and the response returned is `{errors: "Cannot upload a resume in this application state"}`.  Would be good to also have a notification when the submission is successful.  [I used bootstrap alerts for this in another PR](https://github.com/mitodl/bootcamp-ecommerce/pull/620/files#diff-de0f43dd282cf73824d967d8e2bb1fe5R167-R177).

gsidebo: @annagav @mbertrand - We didn't have error/success behavior defined for us initially. Abdul provided feedback on how this should generally work right here: https://github.com/mitodl/bootcamp-ecommerce/issues/641. That doesn't cover every single case where an error might occur from an HTTP request or we have to alert the user about a successful request, but it provides some guidelines. As indicated in that issue, in the case of a success message, we want to bring up a Toast. I think it's fine to completely ignore the error/success behavior in this PR, and instead we can implement it all at once as part of this issue (which I'm going to fill in details for): https://github.com/mitodl/bootcamp-ecommerce/issues/641. We're at risk here of having a whole bunch of custom implementations of error/success messages since we didn't have that behavior spec'ed out initially. It might be better just to ignore it for now and knock it all out in one PR, but I'll leave it to you @annagav to decide if some of that behavior is simple enough that it makes sense to include in this PR

annagav:
Updated:
Screen Shot 2020-06-18 at 8 52 24 AM

static/js/flow/applicationTypes.js
```diff
@@ -64,14 +64,19 @@ export type ApplicationDetail = {
...
+  review_status: ?string,
+  review_status_date: ?string,
+  bootcamp_application: Application,
+  learner: User,
+  interview_url: ?string
+}
+
+export type ApplicationResumeDetail = {
+  resume_file:       ?Object,
+  linkedin_url:          ?string,
```
gsidebo, line 79: Nit: can we go with either single space after the colon, or line up the values? It's neither one here

static/js/lib/queries/applications.js
```diff
@@ -126,5 +128,18 @@ export default {
...
     }
+  }),
+  applicationUploadResumeQuery: (
+    applicationDetail: ApplicationResumeDetail,
+    applicationId: number
+  ) => ({
+    queryKey: "resume",
+    url:      `/api/applications/${applicationId}/resume/`,
+    body:     {
+      ...applicationDetail
```
gsidebo, line 139: I think you can just do `body: applicationDetail`

static/js/lib/queries/applications.js
```diff
@@ -126,5 +128,18 @@ export default {
         "X-CSRFTOKEN": getCookie("csrftoken")
       }
     }
+  }),
+  applicationUploadResumeQuery: (
```
gsidebo, line 132: Convention is to call non-GET query configs "mutations" instead of "queries". So this should be `applicationUploadResumeMutation`

static/js/lib/queries/applications.js
```diff
@@ -126,5 +128,18 @@ export default {
...
+    applicationDetail: ApplicationResumeDetail,
+    applicationId: number
+  ) => ({
+    queryKey: "resume",
+    url:      `/api/applications/${applicationId}/resume/`,
+    body:     {
+      ...applicationDetail
+    },
+    options: {
```
gsidebo, line 141: If we add a return value to the serializer, we can configure this query in such a way that the resume filename will show up immediately in the drawer after it successfully uploads. We'll need to do 2 things:

1. In `applications.views.UploadResumeView.post`, return all of the resume-relevant properties in the response body (`resume_filepath`, `linkedin_url`, `resume_upload_date`). For `resume_filepath` you can move that logic out of the serializer and just add it as a property on the `Application` model itself. We may also want a python test to make sure that those property names are the same in the response from this endpoint and from the application detail endpoint.
1. Add `transform` and `update` properties on this query config. Basically we would just need to update the `applicationDetail` entity state. That state has application ids mapped to an `ApplicationDetail`-type object. I _think_ something like this would work:
    ```javascript
    transform: (json: ResumeUploadResponse) => {
      return {
        [applicationDetailKey]: {
          // The json is not a full `ApplicationDetail` object. Just has some
          // of the keys. That's hopefully ok since we're using mergeDeepRight below in update()
          [applicationId]: json
        }
      }
    },
    update: {
      [applicationDetailKey]: (
        prev: ApplicationDetailState,
        transformed: Object // we could create a type for this too if needed
      ) => (
        R.mergeDeepRight(
          prev,
          transformed
        ) 
      )
    }
    ```
After that, `state.entities.applicationDetail` should look the same, except the resume properties for the logged-in user's detail should be updated, and the uploaded resume file path should be viewable in the drawer right away.

static/js/components/UploadResumeView_test.js
```diff
@@ -0,0 +1,112 @@
...
+    const { inner } = await renderPage()
+    assert.equal(
+      inner.find(".drawer-title").text(),
+      "Resume or LinkedIn Profile"
+    )
+
+    assert.equal(
+      inner.find(".uploaded-resume").text(),
+      `Your uploaded resume: ${
+        application.resume_filepath ? application.resume_filepath : ""
```
gsidebo, line 43: We should make sure that we're explicitly testing when the application has a resume and when it doesn't.

annagav: are you talking specifically about this test? or about the implementation of the component?

static/js/components/UploadResumeView.js
```diff
@@ -0,0 +1,158 @@
...
+type UploaderProps = {
+  application: ApplicationDetail
+}
+export const ResumeUploader = (props: UploaderProps) => {
+  const { application } = props
+  const getUploadParams = () => {
+    return {
+      url: `/api/applications/${application.id}/resume/`,
+      ...DEFAULT_POST_OPTIONS
+    }
```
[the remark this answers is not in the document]
annagav: this is `react-dropzpne-uploader`

static/js/components/UploadResumeView.js
```diff
@@ -0,0 +1,158 @@
...
+      <div className="container">
+        <div className="drawer-wrapper">
+          <h2 className="drawer-title">Resume or LinkedIn Profile</h2>
+          <p>Please upload resume or a link to your LinkedIn profile</p>
+
+          <ResumeUploader application={application} />
+          {application.resume_filepath ? (
+            <p className="uploaded-resume">
+              Your uploaded resume:{" "}
+              {last(split("_", application.resume_filepath))}
```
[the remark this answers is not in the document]
annagav: I am getting a file name like this: 3a3a5e78-511e-4b01-825b-210b0fbdafd0_AnnaInvoinceAugust9.pdf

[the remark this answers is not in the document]
gsidebo: Ah, I see. We would probably need to save the original filename when the resume is uploaded. No need to do that in this PR. Can you at least make sure that file names with underscores show up correctly?

annagav:
@gsidebo I updated the functionality. Please check it out:

Screen Shot 2020-06-19 at 12 20 35 PM

annagav:
Screen Shot 2020-06-19 at 1 17 52 PM

static/js/components/UploadResumeView.js
```diff
@@ -0,0 +1,168 @@
...
+      <div className="container">
+        <div className="drawer-wrapper">
+          <h2 className="drawer-title">Resume or LinkedIn Profile</h2>
+          <p>Please upload resume or a link to your LinkedIn profile</p>
+          <h3>Resume</h3>
+          <ResumeUploader application={application} />
+
+          {application.resume_filepath ? (
+            <p className="uploaded-resume">
+              {getUploadedFileMassage(application.resume_filepath)}
```
gsidebo, line 109: Sorry about the earlier file name suggestion. I forgot that the file names are altered when we save them. We'll probably want to save the original filenames upon upload, but that doesn't need to be part of this PR. For the time being, I can't imagine we're going to want the altered filename being showed here. Can you reimplement the split function you had here before, and just make sure the filename still shows correctly if it has underscores in it?

applications/utils.py
```diff
@@ -13,6 +13,6 @@ def validate_file_extension(value):
         value (File): file name
     """
     ext = os.path.splitext(value.name)[1]
-    valid_extensions = [".pdf", ".doc", ".docx", "odt"]
+    valid_extensions = [".pdf", ".doc", ".docx", ".odt"]
```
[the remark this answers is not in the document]
annagav: this is not related to this PR, and is not going to cause any problem right now. I do not want to touch it because it is not a single line change and it involves tests.

static/js/components/UploadResumeView.js
```diff
@@ -0,0 +1,168 @@
...
+
+const getUploadedFileMassage = (resume_filepath: string) => {
+  return `Your uploaded resume: ${getFilenameFromPath(resume_filepath)}`
+}
+
+const done_btn_style = {
+  color: "#fff",
+  backgroundColor: "#d9534f",
+  borderColor: "#d43f3a",
+}
```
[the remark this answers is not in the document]
annagav: if we decide to get new styles or get a new component all together, `button.dzu-submitButton` is going to stay in the repo forever.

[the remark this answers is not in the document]
gsidebo: If we wanted to change the styles for this, we would want to do it in CSS. We define standard colors/borders/etc. in the CSS, and if we want to make a change to look and feel, it's done in the CSS or, occasionally, using bootstrap classes. If we wanted to change some standard button color, for example, we'd want to do it by changing a CSS variable and have it apply everywhere. We can't do that here. It's rare and unexpected to have styling defined in JS like this. Anyway, not very important. You can leave as-is if you prefer

annagav:
@gsidebo I addressed your comments

annagav:
@gsidebo, I agree about the css comment. However, here we are overriding styles of an existing third party component, and the classes and naming is really non-intuitive and would be hard to find in the code base. We did this a lot in MM, redefining and tweaking all those classes from material-ui. And this is what made upgrading `material-ui` so difficult. Maybe we should think about more efficient practices in this situation.

gsidebo:
@annagav Sorry, didn't see you question. Go ahead and merge if you've tested it locally

PR #670, opened 2020-06-16
asadiqbal08/moved resource pages under homepage

asadiqbal08:
#### What's this PR do?
Moved resource pages under HomePage.

Try to create a resource page under HomePage.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -130,6 +130,7 @@ class HomePage(Page, CommonProperties):
         "HomeAlumniPage",
         "BootcampIndexPage",
         "LearningResourcePage",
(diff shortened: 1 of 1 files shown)


umarmughal824:
@asadiqbal08 don't you think we need to remove from the parent level where it can be added as the sibling?

umarmughal824:
@asadiqbal08 I meant the resource page can be at same level like the home page. See the screenshots for the clarity.

Screenshot 2020-06-17 at 18 11 47

asadiqbal08:
Thanks @umarmughal824 and @ahmed-belal

PR #692, opened 2020-06-18
View statement link should only show up if the user has made a payment

noisecapella:
None

#### What's this PR do?
Hides the "View Statement" link if the user has not yet made any payments for that application

 - Go to `/applications`. Get to a state where you have made one payment. You should see the "View Statement" link in the UI.
 - Find the `Order` for that `BootcampApplication` and set `status='created'`. Go to `/applications` again and the view statements link should not show up. The rest of the UI should look the same as before.

### DIFF
--- a/applications/serializers.py
+++ b/applications/serializers.py
@@ -10,6 +10,7 @@ from applications.constants import (
 )
 from applications.exceptions import InvalidApplicationStateException
 from applications.models import VideoInterviewSubmission
+from ecommerce.models import Order
 from ecommerce.serializers import ApplicationOrderSerializer
 from klasses.serializers import BootcampRunSerializer
 from main.utils import now_in_utc
@@ -105,10 +106,19 @@ class BootcampApplicationSerializer(serializers.ModelSerializer):
     """BootcampApplication serializer"""
 
     bootcamp_run = BootcampRunSerializer()
+    has_payments = serializers.SerializerMethodField()
+
+    def get_has_payments(self, application):
+        """Return true if the user has any fulfilled orders for this application"""
+        # IMPORTANT: orders has already been prefetched and filtered status=Order.FULFILLED
+        # If you are using this outside of BootcampApplicationViewset, you will need to
+        # use prefetch_state_data() or similarly prefetch orders with status=Order.FULFILLED.
+        # If you don't, this will count all orders instead of just the fulfilled ones.
+        return bool(application.orders.all())
 
     class Meta:
         model = models.BootcampApplication
-        fields = ["id", "state", "created_on", "bootcamp_run"]
+        fields = ["id", "state", "created_on", "bootcamp_run", "has_payments"]
 
 
 class SubmissionReviewSerializer(SubmissionSerializer):
--- a/applications/serializers.py
+++ b/applications/serializers.py
@@ -110,11 +110,14 @@ class BootcampApplicationSerializer(serializers.ModelSerializer):
 
     def get_has_payments(self, application):
         """Return true if the user has any fulfilled orders for this application"""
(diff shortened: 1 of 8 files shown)


applications/serializers.py
```diff
@@ -105,10 +106,19 @@ class BootcampApplicationSerializer(serializers.ModelSerializer):
     """BootcampApplication serializer"""
 
     bootcamp_run = BootcampRunSerializer()
+    has_payments = serializers.SerializerMethodField()
+
+    def get_has_payments(self, application):
+        """Return true if the user has any fulfilled orders for this application"""
+        # IMPORTANT: orders has already been prefetched and filtered status=Order.FULFILLED
```
gsidebo, line 113: Can we just pass in a context variable to the serializer that indicates that the application is prefetched? Then in the body of this method we can have a conditional around the value of that context variable. If true, use this `bool()` statement. If false, run the actual query and filter on FULFILLED. It will be useful to have a comment to explain it, but you'll probably want to edit the comments you have here now

applications/views.py
```diff
@@ -44,7 +45,9 @@ def get_queryset(self):
             return BootcampApplication.objects.prefetch_state_data()
         else:
             return (
-                BootcampApplication.objects.filter(user=self.request.user)
+                BootcampApplication.objects.prefetch_state_data()
+                .filter(user=self.request.user)
+                .annotate(num_orders=Count("orders", status=Order.FULFILLED))
```
gsidebo, line 50: Are you using this annotation anywhere?

static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -254,15 +254,17 @@ export class ApplicationDashboardPage extends React.Component<Props, State> {
...
-                <a
-                  className="btn-link"
-                  href={reverse(routes.applications.paymentHistory.self, {
-                    applicationId: application.id
-                  })}
-                >
-                  <span className="material-icons">printer</span>
-                  View Statement
-                </a>
+                {application.has_payments ? (
```
gsidebo, line 257: Nit, take or leave: if theres just a single predicate, the ternary can be written like this...

```
{predicate && (
  <div>jsx</div>
)}
```

static/js/pages/applications/ApplicationDashboardPage_test.js
```diff
@@ -349,13 +349,22 @@ describe("ApplicationDashboardPage", () => {
     })
   })
 
-  it("renders a view statement link", async () => {
-    const { wrapper } = await renderPage()
+  ;[true, false].forEach(hasPayments => {
+    it(`${hasPayments ? "has" : "doesn't have"} a view statement link if hasPayments=${String(hasPayments)}`, async () => {
```
gsidebo, line 353:
```suggestion
    it(`${shouldIf(hasPayments)} have a view statement link if hasPayments=${String(hasPayments)}`, async () => {
```

applications/views.py
noisecapella: No, it was a different attempt at counting fulfilled orders but I couldn't figure how to get this to the serializer.

applications/views.py
```diff
@@ -44,15 +45,23 @@ def get_queryset(self):
             return BootcampApplication.objects.prefetch_state_data()
         else:
             return (
-                BootcampApplication.objects.filter(user=self.request.user)
+                BootcampApplication.objects.prefetch_related(
+                    Prefetch(
+                        "orders",
+                        queryset=Order.objects.filter(
+                            status=Order.FULFILLED
+                        ).select_related("user__profile"),
```
gsidebo, line 53: I don't think this `select_related` is necessary. I tested this without it and it worked.

PR #730, opened 2020-06-23
Add receipt email

### LINKED ISSUE #557: Send receipt emails after successful payment
Email and text-only version of the payment receipts.

┆Attachments: <a href="https://invis.io/MDXEADT3GHQ#/418697202_Bootcamp_Receipt_Email">Bootcamp Receipt Email</a> | <a href="https://trello-attachments.s3.amazonaws.com/5e448428fb73a343e4eeac2c/5ec6dd15ef30f08fce97a545/7c63f45449cdc9a470562240bf487d60/418697202.png">https:&#x2F;&#x2F;invis.io&#x2F;MDXEADT3GHQ#&#x2F;418697202_Bootcamp_Receipt_Email</a> | <a href="https://invis.io/MDXEADT3GHQ#/418697203_Bootcamp_Receipt_Email_Copy">Bootcamp Receipt Email Copy</a> | <a href="https://invis.io/MDXEADT3GHQ#/418697204_Bootcamp_Receipt_Email_Text_Only">Bootcamp Receipt Email Text Only</a> | <a href="https://invis.io/MDXEADT3GHQ#/418697205_Bootcamp_Receipt_Email_Text_Only_Mulitple_Payments">Bootcamp Receipt Email Text Only Mulitple Payments</a>

rhysyngsun:
Closes #557 

#### What's this PR do?
- Adds a receipt email when a payment is successfully made
- Updated/cleaned up the email debugger to be more robust

To fully test this, you'd need to get cybersource up and running locally, but for the sake of keeping testing simple, i'd suggest setting up an application with one or two `Order` records manually and then running this code:

```python
from applications.models import *
from ecommerce import api
app = BootcampApplication.objects.prefetch_state_data().last()
api.send_receipt_email(app.id)
```

I originally approached this wanting to roll up the computations for running balance and other values into the model/query, but that ended up looking like a larger footprint, particularly on the regression testing side. The approach here means we have similar code in both React and django templates, but it delivers this faster and without the possibility of regressions (thus requiring more testing).

Firefox_Screenshot_2020-06-23T01-50-19 769Z

### DIFF
--- a/mail/templates/receipt/body.html
+++ b/mail/templates/receipt/body.html
@@ -1,0 +1,141 @@
+{% extends "email_base.html" %}
+{% load collections dollar_format countries mathfilters parse_date %}
+
+{% block styles %}
+  <style type="text/css">
+    table.receipt-table td {
+      padding: 10px;
+      vertical-align: top;
+    }
+    tr.row-dark {
+      background: #f5f5f5;
+    }
+    td.row-value {
+      color: #000000;
+    }
+    td.total-label {
+      font-weight: bold;
+      color: #000000;
+    }
+  </style>
+{% endblock %}
+
+{% block content %}
+<!-- 1 Column Text + Button : BEGIN -->
+  <tr>
+      <td style="background-color: #ffffff;">
+          <table role="presentation" cellspacing="0" cellpadding="0" border="0" width="100%">
+              <tr>
+                  <td class="row-text">
+                      <p>Dear {{ application.user.profile.name }},</p>
+                      <p>
+                        You have been accepted into the {{ application.bootcamp_run.title }} Bootcamp,
+                        {{ application.bootcamp_run.start_date|date:"F j, Y" }} - {{ application.bootcamp_run.end_date|date:"F j, Y" }}.
+                        You can access your payment statements here on your <a href="{% url "applications" %}">{{ site_name }} Dashboard</a>.
+                      </p>
+                      <p>Below you will find a copy of your receipt:</p>
+                  </td>
+              </tr>
+              <tr>
+                  <td class="row-text">
+                    <h1>Receipt</h1>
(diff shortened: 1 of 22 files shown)


mail/templates/receipt/body.html
```diff
@@ -0,0 +1,141 @@
...
+                        {{ application.bootcamp_run.start_date|parse_iso_datetime|date:"F j, Y" }} -
+                        {{ application.bootcamp_run.end_date|parse_iso_datetime|date:"F j, Y" }}</td>
+                    </tr>
+                    <tr class="row-dark">
+                      <td colspan="2">
+                        <h3>Order Information</h3>
+                      </td>
+                    </tr>
+                    {% for order in application.orders %}
+                    {% cumulative_field_total application.orders "total_price_paid" forloop.counter as running_balance %}
```
mbertrand, line 68: The "Balance" is showing the cumulative amount paid at the time of each order, but I think it's supposed to show the total amount remaining that still needs to be paid.  For instance, "$3000 out of $5000" instead of "$2000".

https://projects.invisionapp.com/share/MDXEADT3GHQ#/screens/418697203_Bootcamp_Receipt_Email_Copy

rhysyngsun: d'oh, good catch

rhysyngsun:
@mbertrand feedback addressed

mail/templates/receipt/body.html
mbertrand: Balance is correct now, in design though it also includes " out of $<total_price>"

rhysyngsun: you'll have to ignore the design a bit - it wasn't specified in the issue, but this UI should be nearly identical to the UI in `static/js/pages/PaymentHistoryPage.js`

PR #738, opened 2020-06-23
Limit to one Job per BootcampRun

noisecapella:
https://sentry.io/organizations/mit-office-of-digital-learning/issues/1740233045/?environment=rc&project=159445&query=is%3Aunresolved

#### What's this PR do?
Adds a unique constraint so that there's only one `Job` per `BootcampRun`. This fixes a configuration issue where some code makes this assumption already but a user could add multiple `Job` objects without realizing this constraint exists. 

Migrations should run correctly


noisecapella:
@rhysyngsun I did already for RC, but I'll check with CI as well. Production should not have any `Job` objects at the moment

rhysyngsun:
LGTM

PR #745, opened 2020-06-23
Fix import error in deployed environments

rhysyngsun:
N/A

#### What's this PR do?
Fixes an import error caused by importing `drf_datetime` in a view. This seemed innocuous enough on its face, but the `test_util` module imports `pytest` so that fails in deployment environments. 

Tests should pass, the `/__emaildebugger__` ui should still work


noisecapella:
Not a blocker for this PR but we should probably use strftime instead, for example: https://github.com/mitodl/bootcamp-ecommerce/blob/master/ecommerce/api.py#L153

PR #787, opened 2020-06-26
Update the link styling for all links to match InVision on both home and product page

### LINKED ISSUE #786: Review link styles and make them the same as [InVision]
- Product Page Feedback: Review link styles and make them the same as [InVision](https://projects.invisionapp.com/d/main?origin=v7#/console/19950477/417931532/preview?scrollOffset=0)
- Home Page Feedback: Update the link styling for all links to match [InVision](https://projects.invisionapp.com/d/main?origin=v7#/console/19950477/418712140/preview?scrollOffset=102) (keep existing hover state)

umarmughal824:
#786 

#### What's this PR do?
fixes #786, Update the link styling for all links to match InVision on both home and product page

Just see the links in dodger-blue color as given in the Invision for both Product and Home Page

**Home Page**

screencapture-bc-odl-local-8099-2020-06-26-14_54_00

**Product Page**

screencapture-bc-odl-local-8099-enterpreneur-bootcamp-bootcamp-13-2020-06-26-15_02_33

### DIFF
--- a/cms/templates/home_page.html
+++ b/cms/templates/home_page.html
@@ -9,6 +9,7 @@
 {% endblock %}
 
 {% block content %}
+<div class="cms-page">
   <div class="home-page">
     <div class="header">
         <div class="container">
@@ -31,4 +32,5 @@
   {% if page.three_column_image_text_section %} {% include "partials/three-column-image-text-section.html" with page=page.three_column_image_text_section %} {% endif %}
   {% if page.alumni %} {% include "partials/homepage-alumni.html" with page=page.alumni %} {% endif %}
   {% if page.learning_resources %} {% include "partials/learning-resources.html" with page=page.learning_resources %} {% endif %}
+</div>
 {% endblock %}
--- a/cms/templates/product_page.html
+++ b/cms/templates/product_page.html
@@ -9,6 +9,7 @@
 {% endblock %}
 
 {% block content %}
+<div class="cms-page">
   <div class="product-page">
     <div class="header">
       {% image page.header_image fill-1920x1080 %}
@@ -39,5 +40,6 @@
   {% if page.instructors %} {% include "partials/instructor.html" with page=page.instructors %} {% endif %}
   {% if page.alumni %} {% include "partials/alumni.html" with page=page.alumni %} {% endif %}
   {% if page.learning_resources %} {% include "partials/learning-resources.html" with page=page.learning_resources %} {% endif %}
-  </div>
+  </>
+</div>
 {% endblock %}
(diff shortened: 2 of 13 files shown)


pdpinch:
When I use inspect mode in InVision, the blue is #0084FF which I don't think is the same as dodger-blue. This is confirmed on the [InVision stylesheet](https://projects.invisionapp.com/d/main?origin=v7#/console/19950477/419499153/preview?scrollOffset=264)

umarmughal824:
@pdpinch I have used that same color code again that name `$dodger-blue`, I don't remember the colors by the names. I got the name from the Google search, I guess that might be wrong but I am not sure. 

Screenshot 2020-06-26 at 17 37 31

gsidebo:
@umarmughal824 If we have the hex value, we should just use that instead of this color name

umarmughal824:
@gsidebo sir, but we need some variable name to use that hex value.

gsidebo:
OK, fair enough. We can leave the color name. My comment about setting a new variable for a standard link color still applies

umarmughal824:
@gsidebo sure, thanks :)

static/scss/cms/admissions-section.scss
```diff
@@ -24,7 +24,7 @@
   }
 
   .admissions-heading > a {
-    color: $crusta-orange;
+    color: $dodger-blue;
```
gsidebo, line 27: If we still need this definition at all, please use `$standard-link-color`

static/scss/cms/program-description.scss
```diff
@@ -112,6 +112,7 @@
 
           &[aria-expanded="true"] {
             font-weight: bold;
+            color: $dodger-blue;
```
gsidebo, line 115: If we still need this definition at all, please use `$standard-link-color`

cms/templates/home_page.html
```diff
@@ -9,6 +9,7 @@
 {% endblock %}
 
 {% block content %}
+<div class="cms-page">
```
gsidebo, line 12: You can just add `cms-page` to the top level element instead of adding another wrapper element, right? `<div class="cms-page home-page">`

cms/templates/product_page.html
```diff
@@ -9,6 +9,7 @@
 {% endblock %}
 
 {% block content %}
+<div class="cms-page">
```
gsidebo, line 12: Same comment as above. `<div class="cms-page product-page">`

cms/templates/home_page.html
umarmughal824: That's the contrainer for one section of the product or home page so that's why had to add a parent contrainer which could aplly it for the whole page.

Earlier I had it besides the home-page so then it applies to the hero section of the page.

cms/templates/product_page.html
umarmughal824: Same as above

PR #789, opened 2020-06-26
update footer styling, background-color etc.

### LINKED ISSUE #788: Update link styling, background color and add a horizontal rule to the footer, as it in in [InVision]
- Update link styling, background color and add a horizontal rule to the footer, as it in in [InVision](https://projects.invisionapp.com/d/main?origin=v7#/console/19950477/417931532/preview?scrollOffset=0)

umarmughal824:
#788 

#### What's this PR do?
fixes #788

Just see the footer of the page

**Desktop Large Size**
Screenshot 2020-06-26 at 14 47 39
**Desktop Medium Size**
Screenshot 2020-06-26 at 14 47 53
**Tablet**
Screenshot 2020-06-26 at 14 48 03
**Mobile**
Screenshot 2020-06-26 at 14 48 14

### DIFF
--- a/static/scss/_variables.scss
+++ b/static/scss/_variables.scss
@@ -30,7 +30,6 @@ $input-border-color: $light-gray;
 $dusty-gray: #979797;
 $very-light-gray: #f5f5f5;
 $soft-light-gray: #d8d8d8;
-$white-smoke: #f1f1f1;
 $link-blue: #0983ff;
 $light-blue: #15729b;
 $cornflower-blue: #579cf9;
@@ -50,9 +49,11 @@ $alabaster: #f9f9f9;
 $gray-80: #ccc;
 $wewak: #eb654d;
 $white-smoke: #f5f5f5;
+$white-smoke-dark: #f0f0f0;
 $trolley-gray: #808080;
 $dodger-blue: #0084ff;
 $night-rider: #303030;
+$dim-gray: #696969;
 
 // Social Icons Colours.
 $facebook: #42599e;
--- a/static/scss/footer.scss
+++ b/static/scss/footer.scss
@@ -10,15 +10,6 @@
   padding-bottom: 20px;
   position: relative;
 
-  &:after {
-    position: absolute;
-    left: -9999px;
-    right: -9999px;
-    bottom: 0;
-    display: block !important;
-    border-bottom: 1px solid black;
-  }
-
   .links-info {
     @include media-breakpoint-down(sm) {
       margin-bottom: 20px;
@@ -35,18 +26,17 @@
     }
 
     a {
-      color: black;
+      color: $matterhorn;
+      text-decoration: underline;
     }
   }
 
   .hbspt-form {
-
     label {
       display: none;
     }
 
     .hs-email {
-
       .hs-input {
         border: 1px solid gray;
         padding: 5px 10px;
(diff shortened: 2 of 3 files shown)


static/scss/_variables.scss
```diff
@@ -53,6 +53,7 @@ $white-smoke: #f5f5f5;
 $trolley-gray: #808080;
 $dodger-blue: #0084ff;
 $night-rider: #303030;
+$white-smoke-dark: #f0f0f0;
```
gsidebo, line 56: We have a definition for `$white-smoke: #f1f1f1;`, which was overwritten by a new definition for `$white-smoke: #f5f5f5;`. Can you delete the first `$white-smoke` definition, then move this to be defined directly under the second `$white-smoke` definition?

static/scss/footer.scss
```diff
@@ -71,15 +61,26 @@
 }
 
 .footer-bottom {
-  padding-top: 20px;
+  padding-top: 40px;
   padding-bottom: 20px;
+  background: $white-smoke-dark;
+  padding-left: 999px;
```
gsidebo, line 67: Can you include a comment about why these padding/margin definitions are necessary?

umarmughal824: That's for carrying the footer background to the corners on both left and right sides.

umarmughal824: Try to remove that it and test it on your system. You will get it pretty easily.

gsidebo: If we need it, please include that as a comment in the code

gsidebo: @umarmughal824 If you're just trying to make sure the background of the `footer-bottom` element stretches to the edge, you can just add a `p-0` class to the `container` element in the footer, then add normal padding to `footer-bottom`. Please try that out and let me know if that doesn't work

umarmughal824: And the reason of giving the big number is that It should not break on bigger screens.

static/scss/_variables.scss
umarmughal824: sure

static/scss/footer.scss
umarmughal824: sure, let me try it.

umarmughal824: that's not going to work what are you suggesting.
on bigger screens, it would not work. I just tested in on the left side.
Screenshot 2020-06-26 at 19 51 52

umarmughal824: this is how it currently looks like on that big size

Screenshot 2020-06-26 at 19 55 48

pdpinch:
I'm not sure where the submit button is coming from, but we'd like it to be inline with the form field. In general, we want it to look similar to [xPRO](https://xpro.mit.edu), but with the colors reversed. 

image

pdpinch:
While we're working on this, the contrast between, for example "About MIT Bootcamps" and the background is too low for accessibility requirements.

pdpinch:
I believe changing the footer links from `Gray` (`#808080`) to `DimGray` (`#696969`) is good enough to pass WCAG AA.

umarmughal824:
@pdpinch that submit button is coming from the Hubspot as it's not marked as Raw html. It will get fixes after marking it as a Raw html.

asadiqbal08:
Adding this [ticket](https://github.com/mitodl/bootcamp-ecommerce/issues/811) here for reference here.

static/scss/footer.scss
```diff
@@ -106,6 +108,11 @@
     line-height: 20px;
     color: gray;
     margin: 0 0 10px;
+
+    .institue-name {
```
gsidebo, line 112: Please remove this definition

Issue #592, opened 2020-05-29
Refunds


odlbot:
As a learner, I would like to ask for a refund, so I can get my money back, net of any non-refundable deposits [[RFC](https://docs.google.com/document/d/1qY5raXyuM91hPP4KgjKZXO7y8nfaQXV_Dw2Jam1ZZCQ/edit)]

┆Attachments: <a href="https://docs.google.com/document/d/1qY5raXyuM91hPP4KgjKZXO7y8nfaQXV_Dw2Jam1ZZCQ/edit">RFC</a>

PR #806, opened 2020-06-26
Refund enrollments command

mbertrand:
Partially addresses #592 

#### What's this PR do?
Creates a management command to issue a refund for an enrollment.

- Apply for a bootcamp and make a full payment 
- In django admin, run `ecommerce.api.complete_successful_order(order)` on the order you made.  There should now be a `BootcampEnrollment` object afterward.
- Run this command and enter an amount greater than what you paid, you should get an error indicating that the refund amount exceeds the amount paid.:
```python manage.py refund_enrollment --user <username> --run <bootcamp_run.id> --amount <too_big_amount>```
- Run this command again but enter an amount that is half of what you paid, the command should complete successfully and there should be a new fulfilled `Order` with a payment amount equal to the negative of what you entered.
- Run it again for the same amount, it should succeed again and there should be an additional order.
- Run it one more time, it should fail with an error indicating that the refund amount exceeds the amount paid.

- Entering an amount is necessary because some bootcamps will have non-refundable deposits that are not stored in the database.
- The code needs to handle refunds for enrollments that may not have applications (which may be possible in the future via bulk purchases).

### CODE REVIEW

File: applications/models.py [hunk 1]
```diff
@@ -246,6 +246,18 @@ def reject_submission(self):
...
         """Revert previous submission decision"""
 
+    @transition(
+        field=state,
+        source=[
+            AppStates.COMPLETE.value,
+            AppStates.AWAITING_PAYMENT.value,
+            AppStates.REFUNDED.value,
+        ],
+        target=AppStates.REFUNDED.value,
```
noisecapella, line 256: Do we need to change anything like `complete` to include `REFUNDED` or others as sources?

ecommerce/management/commands/refund_enrollment.py
```diff
@@ -0,0 +1,49 @@
...
+
+        try:
+            refund_enrollment(user=user, enrollment=enrollment, amount=amount)
+            self.stdout.write(
+                "Refunded enrollment for user: {} ({})\nEnrollment affected: {}".format(
+                    user.username, user.email, enrollment.bootcamp_run.title
+                )
+            )
+        except EcommerceException as exc:
+            self.stderr.write(str(exc))
```
noisecapella, line 49: We should `sys.exit(1)` here or similar as well

klasses/api.py
```diff
@@ -0,0 +1,20 @@
...
+        run_enrollment (BootcampRunEnrollment): The bootcamp run enrollment to deactivate
+        change_status (str): The change status to set on the enrollment when deactivating
+
+    Returns:
+        BootcampRunEnrollment: The updated enrollment
+    """
+    run_enrollment.active = False
+    run_enrollment.change_status = change_status
+    run_enrollment.save()
+    return run_enrollment
```
noisecapella, line 20: It looks like we don't use this return value, and the caller will have it anyway since they are passing it in as an argument. Can we remove this line here if not needed?

klasses/models.py
```diff
@@ -216,6 +220,13 @@ class BootcampRunEnrollment(TimestampedModel):
     bootcamp_run = models.ForeignKey(
         BootcampRun, on_delete=models.CASCADE, related_name="enrollments"
     )
+    change_status = models.CharField(
+        choices=ENROLL_CHANGE_STATUS_CHOICES, max_length=20, null=True, blank=True
+    )
+    active = models.BooleanField(
```
noisecapella, line 226: Do we need to add any code to filter on `active=True`? I'm not sure if we are just storing the enrollment for now or if we are querying it somewhere.

applications/models.py
mbertrand: I guess that depends on whether we want to allow the user to pay for it again and re-enroll using the same application - @pdpinch?

klasses/models.py
mbertrand: Good point, there so API query for enrollments yet but it would probably be best to add that filter [here](https://github.com/mitodl/bootcamp-ecommerce/blob/00dbe5378a3ee68643462ba3e93bab323ecc8519/ecommerce/api.py#L328).

klasses/api.py
mbertrand: 👍

ecommerce/management/commands/refund_enrollment.py
mbertrand: 👍

ecommerce/api.py
```diff
@@ -325,7 +325,9 @@ def complete_successful_order(order):
     run = order.get_bootcamp_run()
     application = order.application
     if application.is_paid_in_full:
-        BootcampRunEnrollment.objects.get_or_create(user=order.user, bootcamp_run=run)
+        BootcampRunEnrollment.objects.get_or_create(
+            user=order.user, bootcamp_run=run, active=True
```
noisecapella, line 329: I think this would break the unique constraint if there was already an enrollment with active=False. We probably need to just get_or_create it, then set `active=True` and `enrollment.save()`

applications/models.py
mbertrand: @pdpinch any thoughts on whether a user should be able to use the same application to make another payment and re-enroll after receiving a refund?

ecommerce/api.py
mbertrand: How about:
```
BootcampRunEnrollment.objects.update_or_create(
            user=order.user,
            bootcamp_run=run,
            defaults={"active": True, "change_status": None},
        )
```

applications/models.py
pdpinch: If it happens that a user requests a refund and then chooses to make more payments, it's going to be such an edge case that I don't think we should plan for it.

PR #814, opened 2020-06-29
fix some accessibility issues on application dashboard page

aliceriot:

none

#### What's this PR do?

this adds keyboard accessibility to a few elements on the application dashboard page. In particular, you can now expand and collapse the sections via the keyboard and tab to the application step 'action' buttons (like 'upload resume' and so on).

I think just try navigating through that page using the keyboard and try to see if there's things you get caught up on.

One thing I noticed is that our dropzone / upload component is not accessible. There is an issue here: https://github.com/fortana-co/react-dropzone-uploader/issues/10 it seems unresolved. So we may want to see about switching to a different component.

Screenshot from 2020-06-29 11-54-14
Screenshot from 2020-06-29 11-54-09

### DIFF
--- a/static/js/components/AccessibleAnchor.js
+++ b/static/js/components/AccessibleAnchor.js
@@ -1,0 +1,30 @@
+// @flow
+import React from "react"
+
+// this component is inteded to take care of accessibility concerns
+// when using an anchor tag as a click target without an href
+
+type Props = {
+  onClick: Function,
+  className: string,
+  children: React$Node
+}
+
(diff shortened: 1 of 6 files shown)


asadiqbal08:
@aliceriot is that fix anyone from this [ticket](https://trello.com/c/uzr6YbZD/197-accessibility) list ?

aliceriot:
@asadiqbal08 yeah thanks for pointing out the list there - the corresponding github issue has no items on it. this just fixes the keyboard navigability.

PR #817, opened 2020-06-29
Make take interview link open in a separate tab

### LINKED ISSUE #816: Video Interview: open Jobma link in a new tab
As a user, I lose my place in the application process when I go to the Jobma site, and there's now way for me to link back. 

It would be convenient if Jobma opened in a new tab.

noisecapella:
Fixes #816 

#### What's this PR do?
Removes the view to create a Jobma interview and the frontend logic which uses it. Instead, the interview will be created via an API call in the backend, with a celery task, when the `BootcampApplication` is first created

First, make sure the relevant models are created:
 - `Job` with the right template id linking to a bootcamp run you want to test with
 -  `ApplicationStep` for the bootcamp with type  "Video Interview"
 -  `BootcampRunApplicationStep` linking to it and a bootcamp run

Create a new `BootcampApplication` for that run (or try deleting and reenrolling via the UI). The application should start a celery task which populates the `Interview` object, and `Interview.interview_url`, for the application.

On the dashboard, click the link to open the video interview tab. You should see a "Take Video Interview" link. Click it and you should have a new tab or window open with a link to the Jobma interview to be started.

### DIFF
--- a/applications/api.py
+++ b/applications/api.py
@@ -1,4 +1,5 @@
 """API for bootcamp applications app"""
+from django.contrib.contenttypes.models import ContentType
 from django.db import transaction
 
 from applications.constants import (
@@ -6,10 +7,17 @@ from applications.constants import (
     REVIEW_STATUS_REJECTED,
     REVIEW_STATUS_PENDING,
 )
-from applications.models import BootcampApplication
+from applications.models import (
+    ApplicationStepSubmission,
+    BootcampApplication,
+    BootcampRunApplicationStep,
+    VideoInterviewSubmission,
+)
+from applications import tasks
+from jobma.api import create_interview_in_jobma
+from jobma.models import Interview, Job
 
 
-@transaction.atomic
 def get_or_create_bootcamp_application(user, bootcamp_run_id):
     """
     Fetches a bootcamp application for a user if it exists. Otherwise, an application is created with the correct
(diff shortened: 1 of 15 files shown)


applications/api.py
```diff
@@ -89,3 +102,34 @@ def get_required_submission_type(application):
...
+
+
+def populate_interviews_in_jobma(application):
+    """
+    Go over each ApplicationStep for the application and create the interviews in Jobma.
+
+    Args:
+        application (BootcampApplication): A bootcamp application
+    """
+    for run_step in BootcampRunApplicationStep.objects.filter(
```
gsidebo, line 114: We need to filter this to only include run steps with `application_step__submission_type=SUBMISSION_VIDEO`

applications/serializers.py
```diff
@@ -41,14 +41,24 @@ class SubmissionSerializer(serializers.ModelSerializer):
     """ApplicationStepSubmission serializer"""
 
     interview_url = serializers.SerializerMethodField(required=False, allow_null=True)
+    take_interview_url = serializers.SerializerMethodField(
+        required=False, allow_null=True
+    )
 
     def get_interview_url(self, submission):
-        """ Get the interview url for a video submission if of that type"""
+        """Return the results URL for the reviewer or others to view the interview"""
```
gsidebo, line 49: Can you rename this to `get_result_interview_url` to disambiguate more clearly?

noisecapella:
@gsidebo `submission_status` matches `review_status` which is also non-nullable. `Order.status` and `BootcampApplication.state` are also non-nullable. Is there a meaningful difference between having `ApplicationStepSubmission.status == PENDING` and `ApplicationStepSubmission.status is None`?

applications/api.py
noisecapella: Thanks, good catch

gsidebo:
@noisecapella I guess that works. Would be nice to have a status that differentiates submissions that have not been started, and ones that have but cannot be completed. We can make that change later if we need to I guess

applications/serializers.py
noisecapella: I agree this is confusing but I'd like to address this later in the interest of time

noisecapella:
@gsidebo any other comments? @pdpinch created a new item #831 which depends on this so we should merge soon if everything looks good

gsidebo:
@noisecapella Testing now

noisecapella:
@gsidebo looks like I need to tweak the CSS. The link is turning white on hover.

PR #834, opened 2020-06-30
Add 'static' to letter template signature url

mbertrand:
N/A

#### What's this PR do?
Fixes the signature link in acceptance/rejection letters

- Accept, then reject, one of your applications.
- You should get an email for both decisions.  On each, check that the signature image url at the bottom works correctly.

### DIFF
--- a/mail/templates/applicant_letter/body.html
+++ b/mail/templates/applicant_letter/body.html
@@ -1,4 +1,5 @@
 {% extends "email_base.html" %}
+{% load static %}
 
 {% block content %}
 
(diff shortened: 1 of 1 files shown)


mail/templates/applicant_letter/body.html
```diff
@@ -14,7 +15,7 @@
                   <td style="padding: 20px; font-family: sans-serif; font-size: 15px; line-height: 20px; color: #555555;">
                       <p style="margin: 0 0 10px;">
                         Sincerely,<br />
-                        <img src="{{ base_url }}/images/signature-mariah-rawding.png" alt="Signature for Mariah Rawding" /><br />
+                        <img src="{{ base_url }}{% static "images/signature-mariah-rawding.png" %}" alt="Signature for Mariah Rawding" /><br />
```
noisecapella, line 18: I think we need to keep `/static` since this could become an absolute URL in production and so `base_url` would interfere with it

mbertrand: Do you mean `"{{ base_url }}/static/images/.....`?

PR #842, opened 2020-06-30
Fixed address factory (which caused flaky username test)

gsidebo:
No ticket

#### What's this PR do?
Fixes `LegalAddressFactory`, which occasionally produces addresses that fail our own address validation on the back end. This was causing the occasional failure of the username generation tests.

None needed, but if you prefer, you can go into a shell and run the following:

```python
from profiles.factories import LegalAddressFactory
# Test that US addresses look right
LegalAddressFactory.build(country="US").__dict__
# Test that CA addresses look right
LegalAddressFactory.build(country="CA").__dict__
# Test addresses that are not limited to a specific country
LegalAddressFactory.build().__dict__
```

### DIFF
--- a/profiles/factories.py
+++ b/profiles/factories.py
@@ -1,9 +1,18 @@
 """
 Factory for Profiles
 """
-import string
+import random
+import pycountry
 from django.contrib.auth.models import User
-from factory import Faker, Sequence, SubFactory, Trait, RelatedFactory
+from factory import (
+    Faker,
+    Sequence,
+    SubFactory,
+    Trait,
+    RelatedFactory,
+    lazy_attribute,
+    random,
+)
 from factory.django import DjangoModelFactory
 from factory.fuzzy import FuzzyText, FuzzyChoice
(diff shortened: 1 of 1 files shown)


profiles/factories.py
```diff
@@ -44,16 +45,29 @@ class LegalAddressFactory(DjangoModelFactory):
...
+
+        NOTE: There didn't seem to be a straightforward way to produce a state/territory using faker,
+        and using random letters for that value results in addresses that fail our own address validation.
+        This implementation is here to ensure that we produce legit-looking addresses that our own app
+        will consider valid.
+        """
+        if self.country not in {"US", "CA"}:
+            return ""
+        subdivisions = pycountry.subdivisions.get(country_code=self.country)
+        subdivision = random.sample(subdivisions, 1)[0]
```
noisecapella, line 68: Can we use Faker's random here so we can use the same seed value? https://faker.readthedocs.io/en/master/#accessing-the-random-instance

PR #865, opened 2020-07-01
Simplify review submission serializer

### LINKED ISSUE #822: Submissions api has many n+1 queries
We should see what we can prefetch to improve performance of this API, and also see what pieces of data are unused by the review interface and can be cut

noisecapella:
Fixes #822 

#### What's this PR do?
 - Separates the review serializer from the application detail serializer
 - Cuts out unused parts of it
 - Add select_related and prefetch_related as appropriate for the review viewset
 - 

Nothing should break. You should be able to view `/review` and update the status of a submission without any problem.

Also, populate 100 submissions on your local instance (I did this with the `ApplicationStepSubmissionFactory`). Go to `/api/submissions/?limit=1000`. It takes about 2 seconds to load for me on master. Switch to this PR and try again, and it should be much faster (about 400 milliseconds for me).

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -158,7 +158,14 @@ class ReviewSubmissionViewSet(
     authentication_classes = (SessionAuthentication,)
     serializer_class = SubmissionReviewSerializer
     permission_classes = (IsAdminUser,)
-    queryset = ApplicationStepSubmission.objects.all()
+    queryset = (
+        ApplicationStepSubmission.objects.all()
+        .select_related(
+            "bootcamp_application__user__profile",
+            "bootcamp_application__user__legal_address",
+        )
+        .prefetch_related("content_object__interview")
+    )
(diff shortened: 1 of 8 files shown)


applications/views.py
```diff
@@ -158,7 +158,14 @@ class ReviewSubmissionViewSet(
     authentication_classes = (SessionAuthentication,)
     serializer_class = SubmissionReviewSerializer
     permission_classes = (IsAdminUser,)
-    queryset = ApplicationStepSubmission.objects.all()
+    queryset = (
+        ApplicationStepSubmission.objects.all()
```
mbertrand, line 162: If submissions with `status=pending` should be filtered out, this is the place to do it

noisecapella: Sounds good, but I'll leave it to a separate issue to fix this

PR #885, opened 2020-07-07
Update test requirements to fix master - #884

ahmed-belal:
#884 

#### What's this PR do?
Pins `isort==4.3.21` in order to have the tests pass.

Build should not break.


pdpinch:
Is there a reason why we don't have a test_requirements.in like we have requirements.in?

ahmed-belal:
Dunno, I think we should!

PR #895, opened 2020-07-07
legal address requirement

### LINKED ISSUE #878: Add legal address requirement to bootcamp application
We should not let a user proceed in their bootcamp application until they have filled out their address. This is an implicit requirement, since user's need to fill out an address to successfully create an account, but we should make this an explicit requirement when determining a user's initial application state (`applications.api.derive_application_state`)

mbertrand:
Closes #878 

#### What's this PR do?
Adds legal address check to the `applications.api.derive_application_state` function

Using django admin, save a blank city or street address to a user's `LegalAddress`.  Then start a new application as that user.  You shouldn't be able to advance beyond "View/Edit Profile" until you enter a complete address.

#### Screenshot
Screen Shot 2020-07-08 at 8 37 39 AM

### DIFF
--- a/applications/api.py
+++ b/applications/api.py
@@ -62,6 +62,8 @@ def derive_application_state(
     if (
         not hasattr(bootcamp_application.user, "profile")
         or not bootcamp_application.user.profile.is_complete
+        or not hasattr(bootcamp_application.user, "legal_address")
+        or not bootcamp_application.user.legal_address.is_complete
     ):
         return AppStates.AWAITING_PROFILE_COMPLETION.value
     if not bootcamp_application.resume_file:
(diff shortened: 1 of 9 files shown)


noisecapella:
Just double checking, will `LegalAddress` always exist for a `User`?

mbertrand:
@noisecapella `LegalAddress` should always exist for a `User` in theory, but anything can happen, especially in a shell or django-admin 😉 

I think the JS warning should probably be dealt with in a separate issue/PR, that code isn't touched in this one, and it happens on master too.  Also it only seems to happen under artificial circumstances (manually deleting and recreating the address) and not under normal conditions when creating an account or editing an existing address.

PR #905, opened 2020-07-08
Paging for submission reviews

### LINKED ISSUE #802: Application Review feedback
See the cheklist at https://trello.com/c/Yq11a7jh/198-application-review-feedback

mbertrand:
Closes #802 

#### What's this PR do?
Adds paging controls and URL parameters to the submission reviews page.

- Go to the `/review` page and try out the paging controls.  If you have <=10 submissions, the paging controls shouldn't appear.  Otherwise, they should work, the url should be updated on each page change, and a change in selected facets should clear out the URL paging parameters.
- Create lots of fake submissions:
```python
ApplicationStepSubmissionFactory.create_batch(105)
```
- Go to the `/review` page and try out the paging controls again.  They should work, the url should be updated on each page change, and a change in selected facets should clear out the URL paging parameters.

I didn't implement the usual endless scrolling so that reviewers can easily navigate back to where they were after finishing an individual review.

Screen Shot 2020-07-08 at 4 15 57 PM

(Optional)

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "react-hot-loader": "^4.12.20",
     "react-meta-tags": "^0.4.1",
     "react-page-click": "^4.0.4",
+    "react-paginate": "^6.3.2",
     "react-redux": "7.2.0",
     "react-router": "^5.1.2",
     "react-router-dom": "^5.1.2",
--- a/static/js/components/SubmissionFacets.js
+++ b/static/js/components/SubmissionFacets.js
@@ -52,13 +52,13 @@ export function Option({ option, facetKey }: OptionProps) {
     e => {
       e.preventDefault()
 
-      const updatedParams = facetIsActive ?
+      let updatedParams = facetIsActive ?
         omit([qsKey], qs.parse(location.search)) :
         {
           ...qs.parse(location.search),
           [qsKey]: getQSValue(option)
         }
-
+      updatedParams = omit(["limit", "offset"], updatedParams)
       const url = urljoin(location.pathname, `/?${qs.stringify(updatedParams)}`)
       history.push(url)
     },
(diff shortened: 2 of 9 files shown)


mbertrand:
@pdpinch @abdulkdawson latest screenshot for review paging, including paging parameters in the URL (`?limit=X&offset=Y`):

Screen Shot 2020-07-08 at 4 15 57 PM

abdulkdawson:
@mbertrand I know this is a screenshot but I'm concerned with the position of the pagination.

mbertrand:
@abdulkdawson happy to make whatever position changes you recommend.  Is it the horizontal alignment that is the main problem?  Currently the paging is centered within the table's containing div, but the positioning/alignment of the status column kind of makes it seem like it's aligned to the right.  Here's a screenshot with borders added to illustrate that.

Screen Shot 2020-07-09 at 11 32 40 AM

abdulkdawson:
Can we increase the spacing between the last name on the list and the pagination? You 10-15px would be great.

mbertrand:
@abdulkdawson here's the latest screenshot with 15px of top padding for the paging controls div: 

Screen Shot 2020-07-09 at 3 37 43 PM

abdulkdawson:
@mbertrand this looks a lot better.

aliceriot:
one initial comment - I think it would be better for the links in the pagination controls to be either react-router links with correct `to` props set or buttons. it's a bit weird and doesn't really fit in with a react-router based app to be manually pushing these onto the `history`.

is it a lot of work to change that? if so, at the minimum we should set `pointer: cursor` on the buttons so it's obvious they're clickable

mbertrand:
@aliceriot the pagination controls are for the most part a 3rd-party app (react-paginate); I used `history.push` to be consistent with how the facets on this page currently work.  Could use state and `history.replace` instead of `history.push` like search does in open, but in that case it seems the facets here should do the same?

aliceriot:
ah, yeah if it's a third-party library I think that's probably ok

PR #906, opened 2020-07-09
Implemented consistent error and success behavior

### LINKED ISSUE #633: Finalize site-wide notifications
**DEPENDS ON https://github.com/mitodl/bootcamp-ecommerce/issues/591**

There are a couple remaining issues with site-wide notifications:
1. They are not shown at all in the Django-driven part of the app (i.e.: the CMS pages). This is due to the fact that we only render them in the React app. 
  - We can take our cues from xpro here. There are two bundles in that app - a header bundle, which just contains the navbar and notification container and is loaded in the CMS pages, and a full-application bundle which contains everything in the header bundle and the rest of the react app. (https://github.com/mitodl/mitxpro/blob/master/static/js/containers/HeaderApp.js, https://github.com/mitodl/mitxpro/blob/master/static/js/containers/App.js, https://github.com/mitodl/mitxpro/tree/master/static/js/entry)
1. Notifications are rendered in the DOM in such a place where they are pushed down by the padding of the body section of the page.
1. The notifications in the CMS support HTML, but we only show text content in the actual notification. We will probably want to support links in these notifications at the very least.

#### Acceptance criteria:
- ~~Notifications are rendered in the CMS pages just as they are in the React app~~ Done
- ~~Notifications are shown immediately below the nav bar~~ Done
- HTML content is properly shown in the notifications

#### Nice to have:
- The title of these notifications in the CMS is just the raw HTML content. If there's an easy way to make that prettier, we should do it.

### LINKED ISSUE #658: Streamline success/error behavior for front end requests and forms
_Based on design feedback gathered in this issue: https://github.com/mitodl/bootcamp-ecommerce/issues/641_

We have a bunch of UI elements that make HTTP requests either upon page load or in response to a button click et. al. We also have several forms that require validation. This issue covers the UI side effects of the success or failure of those requests and form submissions.

### Failed request to load some data, either on page load or in response to a button click

**Examples:** application dashboard page loading a user's applications, loading application details when a user clicks "Expand" in the dashboard, loading submissions that need review in the review listing page.

**Behavior:** Show a dismissable user notification with some error text. Example: "Failed to load (applications/application details/etc.). Please try refreshing the page or [contact support](mailto:BOOTCAMP_SUPPORT_EMAIL) if the problem persists." 

**NOTE:** If it's easy enough to implement, we should probably either make it so the notification area is anchored to the top of the browser window, or force scroll to the top of the page whenever we show one of these notifications. If the user is scrolled down, they will not see these notifications.

### Request to create/update something succeeds

**Examples:** creating a new application via the form, successfully completing a payment, uploading a resume

**Behavior:** Show a user notification with a success message. Ideally this notification would dismiss on its own after a short duration. Example text: "Application started for <bootcamp title>", "Your resume was successfully uploaded", "Your payment was successfully submitted"

**NOTE:** If the given request represents the completion of a form that is presented in a drawer (e.g.: create new application, resume upload, payment), the drawer should be closed when that request succeeds (in addition to showing the user notification)

**ANOTHER NOTE:** We don't have standard behavior for what happens after a button is clicked and a request is in progress. For now, we should set those buttons to `disabled` and add some styling that greys out the button (`static/js/components/NewApplication.js` and `static/scss/_inputs.scss` have an example of this)

### Request to create/update something fails

**Examples:** same as above

**Behavior:** Show some error text directly beneath the button (these requests are all triggered by button click). This text can should have the same styling as form validation text (see below). Example text: "(Resume upload/Payment/Application) failed. Please refresh the page and try again, or [contact support](mailto:BOOTCAMP_SUPPORT_EMAIL) if the problem persists."

**NOTE:** Under no circumstances should we show the user some error text that says something about the "application state". That's a concept that's not relevant to users.

### Form fails validation

Just match what our auth forms do (error text beneath the input, change the input border styling)
ss 2020-06-17 at 15 40 01

### Specific error handling/validation

- The resume upload needs to show an error if the request to upload fails for any reason
- If the user can't upload a resume or submit something in the given application state, we should just disable the input(s)
- The payment drawer form does not validate empty input correctly

### LINKED ISSUE #919: Remove obsolete error handling in UploadResumeView
https://github.com/mitodl/bootcamp-ecommerce/blob/master/applications/views.py#L204-L206

gsidebo:
Closes #658 
Closes #633 
Closes #919 

#### What's this PR do?
- Adds UI artifacts for instances where certain requests succeed and fail
- Re-enables HTML in site-wide notifications

##### Payment success/failure messages
- Create some order for an in-progress application in django admin. Set the status to "created" and note the id
- Apply this patch
    ```
    diff --git a/main/views.py b/main/views.py
    index afbf03f..51cf424 100644
    --- a/main/views.py
    +++ b/main/views.py
    @@ -60,6 +60,13 @@ def react(request, **kwargs):
                     "purchase_date_utc": request.POST["signed_date_time"],
                 }
             )
    +    context["CSOURCE_PAYLOAD"] = json.dumps(
    +        {
    +            "decision": "ACCEPT",
    +            "bootcamp_run_purchased": "<YOUR BOOTCAMP RUN TITLE>",
    +            "purchase_date_utc": "2021-07-08T00:00:00Z",
    +        }
    +    )
         return render(request, "bootcamp/react.html", context)
 
    ```
- Visit your application dashboard with some qs params: `/applications/?status=receipt&order=<your_order_id>`. You should see a notification thanking you for submitting a payment 
- At this point the app should be polling our API until the order status goes to "fulfilled" or "failed", or the request limit is reached.
    - If you set the status to "fulfilled" within the allotted time, the notification should be the same as the bottom screenshot
    - If you set the status to "failed" within the allotted time, the notification should be the same as the second-to-last screenshot
    - If you let the time elapse without changing the status, the notification should be the same as the second-to-last screenshot

NOTE: These notifications should only come up if (a) those querystring params exist and point to an actual order, and (b) the `purchase_date_utc` value in the context is a maximum of 15 min old.

##### Request failures

- Test the new application form
    - Edit `applications.views.BootcampApplicationViewset.create` to immediately raise an exception or return an error response
    - Use the new application form and click "Continue". You should see an error message
- Test the resume/LinkedIn form
    - Edit `UploadResumeView` to immediately raise an exception or return an error response
    - Using an application with no resume, try to add/edit a LinkedIn url and click "submit". You should see an error message

- There are some notes in the issue about what to do when certain requests are successful. What I found while I was working on this was that we didn't need them so much because there was already some user feedback. When a new application is created, the drawer closes and the user can see that application immediately. When profile info is edited, the form closes and shows the read-only profile info UI. When a resume/LinkedIn URL are added, the drawer closes and the user's application state will be updated right away in the dashboard. Adding a notification on top of those bits of feedback seems unnecessary for now

ss 2020-07-08 at 17 15 33
ss 2020-07-08 at 17 16 32
ss 2020-07-08 at 17 17 29

### DIFF
--- a/static/js/components/EditProfileDisplay.js
+++ b/static/js/components/EditProfileDisplay.js
@@ -12,14 +12,17 @@ import EditProfileForm from "./forms/EditProfileForm"
 import { PROFILE_VIEW } from "../constants"
 import users, { currentUserSelector } from "../lib/queries/users"
 import queries from "../lib/queries"
+import { getFirstResponseBodyError, isErrorResponse } from "../util/util"
 import { setDrawerState } from "../reducers/drawer"
 
 import type {
   Country,
   CurrentUser,
   User,
-  HttpAuthResponse
+  HttpAuthResponse,
+  EditProfileResponse
 } from "../flow/authTypes"
+import SupportLink from "./SupportLink"
 
 type StateProps = {|
   countries: ?Array<Country>,
@@ -50,14 +53,16 @@ export class EditProfileDisplay extends React.Component<Props> {
         {})
     }
     try {
-      const {
-        body: { errors }
-      }: // $FlowFixMe
-      { body: Object } = await editProfile(payload)
-
-      if (errors && errors.length > 0) {
+      const response: EditProfileResponse = await editProfile(payload)
+      if (isErrorResponse(response)) {
+        const responseBodyError = getFirstResponseBodyError(response)
         setErrors({
-          email: errors[0]
+          general: responseBodyError || (
+            <span>
+              Something went wrong while updating your profile. Please refresh
+              the page and try again, or <SupportLink />
+            </span>
+          )
         })
       } else {
         updateDrawer(PROFILE_VIEW)
--- a/static/js/components/ResumeLinkedIn.js
+++ b/static/js/components/ResumeLinkedIn.js
@@ -9,6 +9,7 @@ import { values, join, keys } from "ramda"
 import { ErrorMessage, Field, Form, Formik, FormikActions } from "formik"
 import * as yup from "yup"
 
+import SupportLink from "./SupportLink"
 import FormError from "./forms/elements/FormError"
 import { closeDrawer } from "../reducers/drawer"
 import applications, {
@@ -16,7 +17,12 @@ import applications, {
 } from "../lib/queries/applications"
 import { DEFAULT_NON_GET_OPTIONS } from "../lib/redux_query"
 import queries from "../lib/queries"
-import { getFilenameFromMediaPath, isNilOrBlank } from "../util/util"
+import {
+  getFilenameFromMediaPath,
+  getFirstResponseBodyError,
+  isErrorResponse,
+  isNilOrBlank
+} from "../util/util"
 import { appResumeAPI } from "../lib/urls"
 import { ALLOWED_FILE_EXTENSIONS } from "../constants"
 
@@ -102,7 +108,7 @@ type ResumeLinkedInForm = {
 type Props = {
   application: ?ApplicationDetail,
   closeDrawer: () => void,
-  fetchAppDetail: (applicationId: number, force: boolean) => Promise<void>,
+  fetchAppDetail: (applicationId: string, force: boolean) => Promise<void>,
   addLinkedInUrl: (
     applicationId: number,
     linkedInUrl: string
@@ -126,16 +132,17 @@ export class ResumeLinkedIn extends React.Component<Props> {
       return
     }
 
-    const { body: errors }: ResumeLinkedInResponse = await addLinkedInUrl(
-      application.id,
-      linkedInUrl
-    )
+    const linkedInResponse = await addLinkedInUrl(application.id, linkedInUrl)
 
-    // $FlowFixMe: errors can only be null or an array
-    const hasErrors = errors && errors.length > 0
-    if (hasErrors) {
+    if (isErrorResponse(linkedInResponse)) {
+      const responseBodyError = getFirstResponseBodyError(linkedInResponse)
       actions.setErrors({
-        linkedInUrl: errors
+        linkedInUrl: responseBodyError || (
+          <span>
+            Something went wrong while adding your LinkedIn profile. Please try
+            again, or <SupportLink />
+          </span>
+        )
       })
       actions.setSubmitting(false)
     } else {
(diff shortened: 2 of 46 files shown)


static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -69,23 +114,168 @@ const getAppStepSubmissions = (
...
+        } else if (prevState.pollingCount !== pollingCount) {
+          await this.checkForOrderCompletion(orderId)
+        }
+      }
+    }
   }
 
-  onCollapseToggle = (applicationId: number): void => {
+  getOrderIdFromCybersourceParams = (): ?string => {
+    const { location } = this.props
```
mbertrand, line 171: Lots of new order-related code in this PR, is there another issue besides the two mentioned in the PR that it addresses?

static/js/components/ResumeLinkedIn.js
```diff
@@ -25,6 +29,17 @@ import type {
...
 } from "../flow/applicationTypes"
 
+const defaultLinkedInErrorMsg = (
+  <span>
+    Something went wrong while adding your LinkedIn profile. Please try again,
+    or{" "}
+    <a href={SETTINGS.support_url} target="_blank" rel="noopener noreferrer">
+      contact support
+    </a>{" "}
+    to resolve this issue.
```
mbertrand, line 39: This part of the error message is repeated in a few places, maybe put it in a util function?

static/js/pages/applications/ApplicationDashboardPage.js
gsidebo: In a sense, yes. The order-related code handles the case where a user is returning from Cybersource and we want to show them a message if their payment succeeded or failed. We never ended up creating a discrete issue for that, though we should have. I just took care of it in this PR

static/js/components/EditProfileDisplay.js
```diff
@@ -50,11 +51,9 @@ export class EditProfileDisplay extends React.Component<Props> {
...
     }
     try {
-      const {
-        body: { errors }
-      }: // $FlowFixMe
-      { body: Object } = await editProfile(payload)
-
+      const response: EditProfileResponse = await editProfile(payload)
+      const errors =
+        response.body && response.body.errors ? response.body.errors : null
```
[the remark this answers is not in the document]
gsidebo: Great catch

static/js/components/ResumeLinkedIn.js
gsidebo: Makes sense

gsidebo:
@mbertrand re: `ReviewDetailPage`, I think that should be a separate issue if we work on it at all. That's part of the admin app, normal users will never see it, and the actual UI artifact is different from anything we have in the user-facing app. I don't think there's a need to change it at this point

PR #910, opened 2020-07-09
Allow refunds for users without enrollments (partial payments)

### LINKED ISSUE #908: Allow refunds for partial payments without enrollments
If a user has made some partial payments for a bootcamp but hasn't paid in full (and therefore does not have a  `BootcampEnrollment`), allow the user to be refunded anyway.

mbertrand:
Closes #908 

#### What's this PR do?
Allows refunds for users/runs that don't have a corresponding `BootcampEnrollment` yet

- Apply for a bootcamp and make a partial payment 
- In django admin, run `ecommerce.api.complete_successful_order(order)` on the order you made.
- Run this command and enter an amount equal to or less than what you paid it should work:
```python manage.py refund --user <username> --run <bootcamp_run.id> --amount <amount>```

- Apply for a bootcamp and make a full payment 
- In django admin, run `ecommerce.api.complete_successful_order(order)` on the order you made.  There should now be a `BootcampEnrollment` object afterward.
- Run this command and enter any amount, it should succeed:
```python manage.py refund --user <username> --run <bootcamp_run.id> --amount <amount>```
- Check the `BootcampEnrollment`, it should have attributes `active=False` and `change_status=refunded`

### DIFF
--- a/ecommerce/management/commands/refund.py
+++ b/ecommerce/management/commands/refund.py
@@ -27,9 +27,11 @@ class Command(BaseCommand):
             required=True,
         )
         parser.add_argument(
-            "--run", type=str, help="The id for an enrolled BootcampRun"
+            "--run", type=int, help="The id for an enrolled BootcampRun", required=True
         )
-        parser.add_argument("--amount", type=str, help="The amount of the refund")
+        parser.add_argument(
+            "--amount", type=Decimal, help="The amount of the refund", required=True
+        )
         super().add_arguments(parser)
 
     def handle(self, *args, **options):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: ecommerce/management/commands/refund.py [hunk 1]
```diff
@@ -35,16 +35,14 @@ def add_arguments(self, parser):
     def handle(self, *args, **options):
```
noisecapella, line 35: A little out of scope but can we make the other two arguments above required as well? And we can set the `type` of `--run` to `int` and `--amount` to `Decimal`

mbertrand: 👍

Issue #901, opened 2020-07-08
Fix dropzone issues


gsidebo:
There are a few issues with the dropzone library we're using (). 

#### Accessibility

From Alice:
> the react-dropzone-uploader library we're using is not keyboard accessible. see this issue: fortana-co/react-dropzone-uploader#10

We probably need to supply a custom component to the library to fix this ([doc](https://react-dropzone-uploader.js.org/docs/customization#default-components))

#### React warning
A react warning comes up in the console when you open the resume drawer

> Each child in a list should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
    in label
    ...
    in AppContainer

Similar to the accessibility issue, we probably need to supply a custom component to the library to avoid this ([doc](https://react-dropzone-uploader.js.org/docs/customization#default-components))

#### Message clarity
- If an upload fails, it's not clear to the user that anything happened
- There is no text indication that says only PDFs are allowed
- If a user tries to upload a non-PDF, the file isn't uploaded, but no message is shown

### Acceptance Criteria
- Dropzone inputs are accessible
- Dropzone no longer causes React warning in locally running app
- Error message is clearer when an upload fails
- Message added to indicate that only PDF files are accepted
- Some error/indicator shown when a user tries to upload a non-PDF file

aliceriot:
for the input on the dropzone I think I found a CSS only way to do it. Basically, input elements should normally be tab-able, but if they have `display: none` set on them they get taken out of the tab order. So basically I removed `display: none` and simulated it by using `position: fixed` and `opacity: 0` - this makes the input tab-able

gsidebo:
@aliceriot Regarding error messages when something fails to upload: This library does not give you a way to access the server response from a file upload request. Someone opened a PR to add this feature – _gulp_ – almost a year ago, but it still hasn't been reviewed or merged by the library author(s): https://github.com/fortana-co/react-dropzone-uploader/pull/34. Judging from the comments it seems like it works. We could potentially use a fork of that repo with that feature implemented (or create our own fork), but obviously we'll need to consider whether or not we'll want to replace this library since it seems to be abandoned by its maintainer. We get a bunch of features out-of-the-box with this library that aren't available in `react-dropzone`, so we need to consider that.

mbertrand:
[...]

PR #912, opened 2020-07-09
fix dropzone accessibility issue

aliceriot:

part of #901

#### What's this PR do?

This just makes some CSS tweaks to allow the user to tab into the hidden input which is used for the resume upload in the dropzone.

Go to the application page and navigate to the resume upload drawer. Initially the 'x' should be selected, and then if you hit tab you'll be on the input. you should then be able to press 'enter' to open a file dialog.

### DIFF
--- a/static/scss/resume.scss
+++ b/static/scss/resume.scss
@@ -21,6 +21,12 @@
     text-align: center;
     color: $black;
 
+    input.dzu-input {
+      position: fixed;
+      opacity: 0;
+      display: initial;
+    }
+
     .file-types {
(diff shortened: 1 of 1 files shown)


aliceriot:
I was looking into that and didn't find anything _great_ that we can do - we could turn off the opacity when the item is focused, but that just sort of makes it pop out of nowhere. it works, but the problem is that the `input` tag also receives focus when you click on it, so something to make it visible when you're tabbing around will also flash into view when the user clicks or drops a file...kinda yucky.

PR #917, opened 2020-07-10
Conditional hubspot/zendesk JS

### LINKED ISSUE #907: Gracefully handle blocked external javascript
If external JS for hubspot or zendesk are blocked or fail to load for whatever reason, make sure that the embedded script variables that depend on them are not undefined before use:

```javascript
if (typeof hbspt !== "undefined") {
          hbspt.forms.create({......});
}
```

```javascript
if (typeof zE !== "undefined") {
        zE('webWidget', 'prefill', {....})
}
```

mbertrand:
Closes #907 

#### What's this PR do?
Places embedded hubspot/zendesk JS code inside `if` statements to prevent running them if the variables are undefined (most likely due to ad blockers).

- Populate the following using values from CI:
  ```
  HUBSPOT_FOOTER_FORM_GUID
  HUBSPOT_PORTAL_ID
  ZENDESK_HELP_WIDGET_ENABLED=True
  ```
- Load any page, both the zendesk widget and hubspot email form should appear near the bottom right.
- Add the following to your ad blocker plugin blocklist:
  ```
  *hubspot*
  *hsform*
  *.zendesk.*
  ```
- Reload the web page with the JS console in view.  The zendesk widget and hubspot email form should not be visible, and there should not be any console errors about either `hbspt` or `zE` being undefined.

Screen Shot 2020-07-10 at 2 35 23 PM

### DIFF
--- a/main/templates/base.html
+++ b/main/templates/base.html
@@ -42,16 +42,18 @@
     {% endblock %}
     {% if zendesk_config.help_widget_enabled %}
       <script type="text/javascript">
-        zE('webWidget', 'prefill', {
-            name: {
-                value: '{{user.name}}',
-            },
-            email: {
-                value: '{{user.email}}',
-            }
-        });
-        // set page title as a search term
-        zE('webWidget', 'helpCenter:setSuggestions', { search: document.title });
+        if (typeof zE !== 'undefined') {
+            zE('webWidget', 'prefill', {
+                name: {
+                    value: '{{user.name}}',
+                },
+                email: {
+                    value: '{{user.email}}',
+                }
+            });
+            // set page title as a search term
+            zE('webWidget', 'helpCenter:setSuggestions', { search: document.title });
+        }
       </script>
     {% endif %}
     {% render_bundle 'third_party' %}
--- a/main/templates/footer.html
+++ b/main/templates/footer.html
@@ -19,11 +19,13 @@
         <![endif]-->
         <script charset="utf-8" type="text/javascript" src="//js.hsforms.net/forms/v2.js"></script>
         <script>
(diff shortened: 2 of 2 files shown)


main/templates/base.html
```diff
@@ -42,16 +42,18 @@
...
-            name: {
-                value: '{{user.name}}',
-            },
-            email: {
-                value: '{{user.email}}',
-            }
-        });
-        // set page title as a search term
-        zE('webWidget', 'helpCenter:setSuggestions', { search: document.title });
+        if (typeof zE !== 'undefined') {
```
gsidebo, line 45: I think we can simplify this:
```suggestion
        if (!!zE) {
```

main/templates/footer.html
```diff
@@ -19,11 +19,13 @@ <h2><a href="https://mitbootcamps.zendesk.com/hc/en-us/requests/new">Contact Us<
         <![endif]-->
         <script charset="utf-8" type="text/javascript" src="//js.hsforms.net/forms/v2.js"></script>
         <script>
-          hbspt.forms.create({
-            portalId: '{{ hubspot_portal_id }}',
-            formId: '{{ hubspot_footer_form_guid }}',
-            target: '#hs_footer_form'
-          });
+          if (typeof hbspt !== 'undefined') {
```
gsidebo, line 22: Same as above
```suggestion
          if (!!hbspt) {
```

mbertrand: I get `Uncaught ReferenceError: hbspt is not defined` again if I change the code to `!!hbspt` - probably because the variable is not declared if the external JS is blocked.

PR #922, opened 2020-07-14
Various improvements to django admin classes

gsidebo:
No ticket. Noticed this while checking off RC boxes and such

#### What's this PR do?
Makes a bunch of improvements to Django admin classes to make them more usable and readable

(Required)

Click through the affected Django admin classes, check out list pages, detail pages, and new object forms. Make sure everything looks correct

### DIFF
--- a/applications/admin.py
+++ b/applications/admin.py
@@ -35,33 +35,29 @@ class BootcampRunApplicationStepAdmin(admin.ModelAdmin):
     model = models.BootcampRunApplicationStep
     list_display = (
         "id",
-        "get_bootcamp_title",
-        "get_run_title",
-        "due_date",
+        "get_run_display_title",
         "get_submission_type",
         "get_step_order",
+        "due_date",
     )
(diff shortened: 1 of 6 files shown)


applications/admin.py
```diff
@@ -166,43 +165,44 @@ def get_queryset(self, request):
...
+    get_run_display_title.short_description = "App Bootcamp Run"
+    get_run_display_title.admin_order_field = (
         "app_step_submission__bootcamp_application__bootcamp_run__title"
     )
 
     def app_step_submission_link(self, obj):
         """Returns a link to the related bootcamp application"""
-        app_step_submission = obj.app_step_submissions.first()
-        if app_step_submission is None:
+        app_step_submissions = obj.app_step_submissions.all()
```
annagav, line 202: why this change?

gsidebo: Takes advantage of the prefetched queryset. If you use `first()`, that runs an extra query.

PR #944, opened 2020-07-29
add cache-control header to hash.txt and api urls

### LINKED ISSUE #952: Add cache control headers to responses
When Fastly caches API responses it can lead to unexpected and unpredictable behavior on the client side. We have currently excluded all API routes from caching in the Fastly configuration, but it would be best to add the appropriate cache control headers in the application to ensure that we are enforcing the intended behavior everywhere.

mbertrand:
Closes #952

#### What's this PR do?
Adds cache-control headers to API responses (via new middleware) with a value of `private, no-store` [as recommended by fastly](https://docs.fastly.com/en/guides/configuring-caching#do-not-cache).
Also adds [the same cache-control headers as xpro](https://github.com/mitodl/mitxpro/pull/1658/files) to `/static/hash.txt` (via nginx.conf)

Make some requests to API and non-API URLs.  The API URL responses should have a `Cache-Control: private, no-store` header, other URL responses should not.  The heroku PR instance should also have cache-control headers for `/static/hash.txt`.

### DIFF
--- a/main/middleware.py
+++ b/main/middleware.py
@@ -1,0 +1,12 @@
+""" Middleware classes for the main app"""
+from django.utils.deprecation import MiddlewareMixin
+
+
+class CachelessAPIMiddleware(MiddlewareMixin):
+    """ Add Cache-Control header to API responses"""
+
+    def process_response(self, request, response):
+        """ Add a Cache-Control header to an API response """
+        if request.path.startswith("/api/"):
+            response["Cache-Control"] = "private, no-store"
+        return response
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -47,6 +47,11 @@ http {
             try_files /static/images/favicon.ico /favicon.ico;
         }
 
+        location ~* /static/hash.txt {
+            expires -1;
+            add_header Cache-Control private;
+        }
+
         location ~* /static/(.*$) {
             expires max;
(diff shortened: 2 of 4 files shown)


noisecapella:
Can we add support for async middleware? There's a snippet in the Django docs: https://django.readthedocs.io/en/stable/topics/http/middleware.html#asynchronous-support

main/middleware.py
```diff
@@ -0,0 +1,13 @@
...
+
+from django.utils.deprecation import MiddlewareMixin
+
+
+class CachelessAPIMiddleware(MiddlewareMixin):
+    """ Add Cache-Control header to API responses"""
+
+    def process_response(self, request, response):
+        """ Add a Cache-Control header to an API response """
+        if request.path.lower().startswith("/api/"):
```
noisecapella, line 11: The `.lower()` shouldn't be needed here

mbertrand:
@noisecapella I think async middleware support was added in Django 3.0, bootcamp is on version 2.2

noisecapella:
@mbertrand The sample code in the docs should work with the current Django version. It won't do anything useful yet but it will make the upgrade process a little easier once we do upgrade bootcamp to Django 3.x

mbertrand:
@noisecapella I can't include `from django.utils.decorators import sync_and_async_middleware` as shown in the example because it doesn't exist yet in 2.2

PR #965, opened 2020-08-10
Add signatory name and signature customization fields in cms letter template page

### LINKED ISSUE #818: Add customization options for acceptance/rejection letter
https://github.com/mitodl/bootcamp-ecommerce/pull/744

The above PR implemented the first version of the acceptance/rejection letter. In that implementation, the signatory name and the signature image were hard-coded into the letter template. We need to fix that (probably by adding those properties to the `LetterTemplatePage ` class). There may be other customization options we want to add to this issue as well

umarmughal824:
#818 

#### What's this PR do?
Add signatory name and signature customization options for cms letter template page

Just see cms template page has signatory field and check all the pages should show Mariah Rawding, signatory, for preview, acceptance/rejection letter and application letter

#### Screenshots
Screenshot 2020-08-13 at 11 30 35
Screenshot 2020-08-13 at 11 30 26
Screenshot 2020-08-13 at 11 32 34
Screenshot 2020-08-13 at 11 32 46
Screenshot 2020-08-13 at 11 31 48


umarmughal824:
@pdpinch This is designed while keep one thing in mind that we will have the letter template already and these two are required fields. So if we will have that template then we will have to provide that both.

noisecapella:
`ApplicantLetter` keeps a record of the messages sent to applicants. Should we include the signature there too or is that unnecessary?

pdpinch:
Is there a wagtail-friendly way to deploy with default values that will work?

The default values could be the ones we already use, Mariah’s name and signature. 

> @pdpinch <https://github.com/pdpinch> This is designed while keep one thing in mind that we will have the letter template already and these two are required fields. So if we will have that template then we will have to provide that both.

umarmughal824:
@pdpinch yeah I will add it the current values as the default ones.

noisecapella:
I think it would be simpler to break this PR into two parts
 - Add the fields to the `LetterTemplatePage`. Release just that change, then adjust the defaults on production. Mariah's signature would still be used.
 - Switch to use the new fields for signatures

Any thoughts?

umarmughal824:
@pdpinch @noisecapella  now, on that PR deployment to prod, it will still display default "Mariah" signatory as long as we don't provide the customized signatory on the cms letter template page.

noisecapella:
@umarmughal824 did you see my comment about making this just a PR to add the fields, then having another PR to do the implementation? https://github.com/mitodl/bootcamp-ecommerce/pull/965#issuecomment-672224775

umarmughal824:
I thought that would be acceptable. Sure, changes it as suggested.

noisecapella:
@umarmughal824 Thanks, this PR looks good to me. I'll take a look at the other PR when it's ready for review

PR #968, opened 2020-08-12
Updated configure_cms command to create resource pages and letter template - #882

### LINKED ISSUE #882: Add data migrations for CMS resource pages and letters
We should have data migrations that do the following:

1. If they don't exist, add the necessary pages/settings for an initial version of the resource pages we expect in the app (terms and conditions, about us, etc.)
2. If they don't exist, add an initial version of the acceptance/rejection letters

The site does not work correctly without these items in place. This will aid local development and testing in deployed environments

**NOTE:** Keep an eye on this issue – #900 – when working on this. If we move away from CMS data migrations entirely, then the acceptance criteria above will be the same except it should be implemented in a management command(s) instead of data migrations

ahmed-belal:
Closes #882 

#### What's this PR do?
Updates the `configure_cms` command to create the required resource pages and letter template page in CMS so that the dev environment can start working without much manual content creation.

Run the `configure_cms` command. Multiple runs of the command on the same environment should not matter as the command is intended to be idempotent.


umarmughal824:
@ahmed-belal could you please rebase it to get the latest migration for letter template so that I could review it?

PR #969, opened 2020-08-12
Allow user to retry cybersource compliance validation

### LINKED ISSUE #909: Provide feedback to user for cybersource compliance errors
Cybersource compliance checks against OFAC/SDN sometimes return non-fatal errors (meaning the request could be retried). Particulary there are at lest 2 that solvable by the user themselves:

- Code `101` - a field is blank
- Code `102` - a field had invalid info
  - we've seen this when there are (presumably) characters in a field that does not pass cybersource validation
   - Example: one or both of `+`, `_` in a name field

I'd propose we make the following changes to make this more user-friendly:

- Propagate cybersource validation errors back to the UI and display them to the user. These probably will be a generic "invalid input" or something less curt, since we don't get specifics of _why_ it failed, just that it did.
- For existing users, we'd need to prepopulate the legal address fields with the user's info so the validation errors actually have context. We probably should do the same for step 2 as well if it's not a huge lift
   - We can't use the existing API unfortunately because it requires a session, which the user doesn't have yet, so this needs to come back on the auth api response

mbertrand:
Closes #909

#### What's this PR do?
- Allows users to modify their name/address and resubmit on registration if cybersource compliance validation returns certain error codes, or raises a `UserTryAgainLaterException`.

(Required)

(Optional)

(Optional)

##### Error code 150 or `UserTryAgainLaterException` exception:

Screen Shot 2020-08-12 at 10 22 05 AM

##### Error code 102

Screen Shot 2020-08-12 at 10 22 45 AM

##### Error code 101

Screen Shot 2020-08-12 at 10 22 58 AM


static/js/constants.js
```diff
@@ -222,3 +222,16 @@ export const FACET_OPTION_LABEL_KEYS = {
...
+export const CS_101 = "CS_101"
+export const CS_102 = "CS_102"
+export const CS_150 = "CS_150"
+export const CS_0 = "CS_0"
+
+export const CS_ERROR_MESSAGES = {
+  [CS_101]:
+    "Missing required fields. Please modify your name/address and try again.",
+  [CS_102]: "Invalid data. Please modify your name/address and try again.",
+  [CS_150]: "Please try again.",
```
pdpinch, line 235:
```suggestion
  [CS_150]: "Please try again. If this happens more than once, please contact customer support",
```

Ideally, "contact customer support" would be linked to https://mitbootcamps.zendesk.com/hc/en-us/requests/new -- however, if it takes more than 15m to get HTML to work in the error message, it's not worth iit.

static/js/constants.js
```diff
@@ -222,3 +222,16 @@ export const FACET_OPTION_LABEL_KEYS = {
...
+  [CS_0]:   "Please try again."
```
pdpinch, line 236: What's `CS_0`? Is that a real error, or a catch-all?  

```suggestion
  [CS_0]:   "Please try again. If this happens more than once, please contact customer support."
```

mbertrand: catch all

static/js/constants.js
mbertrand: @pdpinch I updated the error messages & screenshots

authentication/pipeline/compliance.py
```diff
@@ -5,19 +5,33 @@
...
 
+def set_user_session_data(strategy, user):
+    """
+    Save the user data in session.
+
+    Args:
+        strategy (social_django.strategy.DjangoStrategy): the strategy used to authenticate
+        user(User): the user to save to the session
+    """
+    strategy.request.session["COMPLIANCE_DATA"] = UserSerializer(user).data
```
rhysyngsun, line 29: I don't think we should be storing this data on the user's session and writing it out into the html, it should be incorporated as part of the social auth api responses instead.

Here's how i'd do it:
- Keep the redirect as-is
- When the view loads, the frontend queries the /retry endpoint immediately with no data, which will cause it to get the same response but in JSON form.
- Update the /retry endpoint to return the value of `UserSerializer` in the `extra_data` field, similar to how the code handling `RequirePasswordException` does.
  - This may get a little tricky here, because we don't want this whole data structure coming back in the response for `RequirePasswordException` for security reasons. That might require moving `extra_data` into `PartialException` instead of having it in the serializer.

static/js/components/forms/RegisterDetailsForm.js
```diff
@@ -31,35 +35,47 @@ const INITIAL_VALUES = {
...
-              Continue
-            </button>
+const RegisterDetailsForm = ({
+  onSubmit,
+  countries,
+  includePassword
+}: Props) => {
+  if (!includePassword && INITIAL_VALUES.password) {
+    // $FlowFixMe
+    delete INITIAL_VALUES.password
```
rhysyngsun, line 45: This shouldn't be mutating a `const`, it's probably better to have two consts, one without `password, the other:

```js
const INITIAL_WITH_PASSWORD = {...INITITAL_WITHOUT_PASSWORD, password: ""}
```

authentication/pipeline/compliance.py
mbertrand: @rhysyngsun Not sure I understand.  Keep which redirect as is?  When which view loads?  The `RequirePasswordException` handling [here](https://github.com/mitodl/bootcamp-ecommerce/blob/mb/compliance/authentication/serializers.py#L282-L287)?

rhysyngsun: Sorry, I guess maybe I'm not understanding the code correctly, what's the purpose of `COMPLIANCE_DATA`? It looked like it prepopulates this data on app init, which I don't think we want to do. Instead, the auth APIs should return whatever the frontend needs in its response.

mbertrand: @rhyngsun I couldn't figure out how to pass along the serialized user data to the `/retry` page via social auth responses that worked for both new users registering for the first time and existing users logging in that had previously failed/skipped export compliance.  The session variable was a way around that.  It sounds like you know how to do it with social auth but I wasn't sure which redirect/view you're referring to and not sure looking at the code exactly how the `RequirePasswordException` leads to a populated `extra_data` field.

mbertrand: @rhysyngsun I couldn't figure out how to pass along the serialized user data to the `/retry` page via social auth responses that worked for both new users registering for the first time and existing users logging in that had previously failed/skipped export compliance.  The session variable was a way around that.  It sounds like you know how to do it with social auth but I wasn't sure which redirect/view you're referring to and not sure looking at the code exactly how the `RequirePasswordException` leads to a populated `extra_data` field.

mbertrand: I think I get it now - Modify the compliance serializer so that `extra_data` == `UserSerializer(user).data`.  When the new compliance retry page initially loads, make a post call to `api/register/email/retry` which will return an auth response that should include that `extra_data` field.  Use it to prepopulate the form instead of the session data.  Does that sound right?

rhysyngsun: yeah, exactly

mbertrand: Thanks, sometimes I'm a little slow and need to sleep on things! 😁

PR #987, opened 2020-08-25
Add loaders for API requests

### LINKED ISSUE #914: Add loading UI elements
Based on feedback gathered in https://github.com/mitodl/bootcamp-ecommerce/issues/847

We currently have no loading indicators in the app. If a button is clicked and it triggers a long-running request, or some request is made on page load, the user gets little or no indication that some work is being done. 

This library was put forward as a possible inspiration: https://loading.io/button/. Abdul indicated that we should use an indicator like the "ld-over" example at the top for buttons that trigger an HTTP request, and something like the "ld-over-full" example for requests that happen on page load. 

Some notes:
- We don't need to use this library, or if we do go with this library, we don't need to use those exact animations. We should do some browser testing with whatever we use, and whoever works on this should send some screenshots to Abdul before putting up a PR.
- When there is a page load, we probably only want to put an overlay over the main content section. In other words, we don't need to cover the navbar and footer while content is loading.
- We may want to think about delaying the loading animation on page load, or fading it slowly in. If a page load request completes very quickly, we may see some weird blinking if we show the loading indicator immediately when the page starts loading. We can experiment with this.

noisecapella:
Fixes #914 

#### What's this PR do?
Adds loader indications where API requests are taking place

TBD

TBD


noisecapella:
@mbertrand Thanks for spotting that. Should be fixed now. For the other two issues, it looks technically possible to solve by keeping the spinner active even after submit, but we would need to complicate the logic where the form submission happens and the work to do that looks non-trivial

mbertrand (approved):
👍

PR #989, opened 2020-08-27
Remove Payment component, used in the previous version of this application

### LINKED ISSUE #616: Remove deprecated pages and related code
We have at least 2 pages that are being deprecated by the site redesign.
- The `/pay/` page
- The old index page (`index.html`)

We should get rid of those pages and any associated views and styling.

**NOTE:** The `/pay/` page was removed in #852, but that PR did not get rid of at least one API endpoint, and the old `Payment` component is still there (`static/js/components/Payment.js`). Please review that PR and make sure that all of the components/templates/endpoints/styles/etc. that are related only to the old payment page are removed.

noisecapella:
Fixes #616 

#### What's this PR do?
Removes the `Payment` component, which has been replaced by `PaymentDisplay`

Nothing should break


gsidebo:
@noisecapella Can you check to see if this fully satisfies #616? If so, please add that to the relevant tickets. If not, could you add to this PR so that it does?

PR #993, opened 2020-09-01
Add support for sticky notifications

### LINKED ISSUE #824: Enable fixed position and auto-dismiss user alerts
We have user notifications that appear directly beneath the nav bar.

ss 2020-06-29 at 18 42 13

We're using this same notification element to show some error/success messages to users (#906). This is less than ideal for a few reasons:
1. These notifications are not in fixed position, so if the user is scrolled down on the page, they will not see this notification unless they scroll up, and they will not see it at all if they navigate to another page.
1. If the user has a drawer open, the overlay renders above the notification element and almost completely hides it.

We will implement fixed position notifications for these types of messages.

### Acceptance criteria
- Site-wide notifications (sourced from the CMS) still shown as they were before (under the nav bar, not in fixed position, rendered under the overlay if a drawer is open)
- Error/success messages that are rendered in an alert/notification now shown in fixed position at the bottom of the page (i.e.: they stay at the bottom of the page when scrolling), and are shown above all other elements. 

### Nice to have
- Success messages auto-dismiss after a short period of time

### LINKED ISSUE #914: Add loading UI elements

noisecapella:
Fixes #824 

#### What's this PR do?
Adds sticky footer notifications for success and error notifications. Site-wide notifications will still show up at the top like they did before. (Site-wide notification code was moved around but this is just a refactor.)

I started working on auto-dismiss success notifications but it looked non-trivial to implement in this PR so I left it out.

To fake success and failure messages, add this to `ApplicationDashboardPage.componentDidMount`:

      async componentDidMount() {
        const { applications, addSuccessNotification, addErrorNotification } = this.props
        if (applications) {
          const orderId = this.getOrderIdFromCybersourceParams()
          if (orderId) {
            await this.handleCybersourcePageLoad(orderId)
          }
        }
    
        // TODO: make sure to remove this in the PR
        addSuccessNotification({
          props: {
            text: "success!"
          }
        })
        addErrorNotification({
          props: {
            text: "oh no!"
          }
        })
      }

Make sure the notifications stick to the bottom as you scroll up and down. Also, open the drawer and verify that the notification is visible with the drawer behind it.

For CMS notifications, add a new `SiteNotification` from the Django admin. Then go to the dashboard to view the notifications.

#### Screenshots
Screenshot from 2020-09-01 17-21-01
Screenshot from 2020-09-01 17-28-21

Updated screenshot with newer font:
Screenshot from 2020-09-02 14-43-11


abdulkdawson:
The text in the sticky notification should be Helvetica Bold 18px. Other than that it looks good.

main/templates/footer.html
```diff
@@ -74,3 +74,4 @@ <h3>MIT Bootcamps</h3>
     </div>
   </div>
 </footer>
+<div id="footer-notifications" class="fixed-bottom"></div>
```
gsidebo, line 77: I think a better approach would be to add an empty block to `base.html` (called something like `postfooter`), then fill in that block with this element in `react.html`. As a bonus, you might be able to get rid of the DOMContentLoaded event listener if you do this

static/js/components/notifications/NotificationContainer_test.js
```diff
@@ -143,4 +145,28 @@ describe("NotificationContainer component", () => {
       1
     )
   })
+
+  //
+  ;[true, false].forEach(filterAlertType => {
+    it(`${shouldIf(filterAlertType)} based on alert types`, async () => {
```
gsidebo, line 151: The output of this will be something like `should based on alert types`. Probably missing a word or two

static/scss/notification.scss
```diff
@@ -47,3 +47,10 @@
     }
   }
 }
+
+#footer-notifications {
+  position: sticky;
+  top: 0;
+
+  font: bold 18px "Helvetica", sans-serif;
```
gsidebo, line 55: Helvetica is already the standard font for the site. I don't see any reason why we should specify the font family again here

static/js/components/notifications/FooterNotifications.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+import React from "react"
+import NotificationContainer from "./NotificationContainer"
+import { ALERT_TYPE_ERROR, ALERT_TYPE_SUCCESS } from "../../constants"
+
+const FooterNotifications = () => (
```
gsidebo, line 6: Is this component necessary? Could we just render `<NotificationContainer ... />` directly in root.js?

gsidebo (changes requested):

main/templates/footer.html
noisecapella: This div needs to go below the `footer` element, otherwise the footer notifications stop just above the footer when scrollin

gsidebo: It can be still be placed below the footer with the approach I suggested. I'm saying we should add a new block to `base.html`, which already defines a block for the footer.

```
# base.html
    {% block footer %}
      {% include "footer.html" %}
    {% endblock %}
    {% block postfooter %}
    {% endblock %}

# react.html
{% block content %}
<div id="app-container" class="top-level-page-body">
</div>
{% endblock %}
{% block postfooter %}
<div id="footer-notifications" class="fixed-bottom">
</div>
{% load render_bundle %}
{% render_bundle "root" %}
{% endblock %}
```

gsidebo:
Clarified my comment above

noisecapella:
@gsidebo I see, it looks like I had the render_bundle nested in the content block. Should work fine now.

gsidebo (approved):
Looks great! 👍

PR #996, opened 2020-09-03
Fixed react-dropzone-uploader bugs

### LINKED ISSUE #901: Fix dropzone issues

gsidebo:
Fixes #901 

#### What's this PR do?
- Gets rid of react warning when loading the dropzone component
- Clearly shows errors within the component, and shows error text from the server response if any came back

First, create/edit and application such that it is awaiting resume upload

Then:
- Set `MAX_RESUME_FILE_SIZE` to something like `1`, then try to upload a file. You should see an error saying the file is too large.
- Apply the patch below to force an error from the server when a resume upload is attempted, then use the dropzone to upload a resume. You should see the server error message displayed.
- Use the interactive elements of the dropzone and make sure everything looks correct.

```diff
diff --git a/applications/views.py b/applications/views.py
index a3e21130..602070df 100644
--- a/applications/views.py
+++ b/applications/views.py
@@ -193,6 +193,11 @@ class UploadResumeView(GenericAPIView):
         """
         Update the application with resume and/or linkedin URL
         """
+        return Response(
+            {"error": "This is an error from the server"},
+            status=status.HTTP_400_BAD_REQUEST
+        )
+
         application = self.get_object()
         linkedin_url = request.data.get("linkedin_url")
         resume_file = request.FILES.get("file")
```

#### Background info
I created a fork of the react-dropzone-uploader repo under the mitodl org, made some fixes and additions, and published the package to npm ([forked repo](https://github.com/mitodl/react-dropzone-uploader), [npm page](https://www.npmjs.com/package/@mitodl/react-dropzone-uploader))

ss 2020-09-03 at 14 22 19
ss 2020-09-03 at 14 25 43
ss 2020-09-03 at 14 26 56


gsidebo:
@abdulkdawson Can you take a quick look at the screenshots above? This is for the uploader component for resumes in bootcamps

static/js/constants.js
```diff
@@ -209,6 +209,8 @@ export const NEW_APPLICATION = "newApplication"
 export const TAKE_VIDEO_INTERVIEW = "TAKE_VIDEO_INTERVIEW"
 export const RESUME_UPLOAD = "resumeUpload"
 
+export const MAX_RESUME_FILE_SIZE = 15000000 // 15MB
```
mbertrand, line 212: The limit on the nginx/server side seems to be lower than this, when I try uploading a 10 MB file it fails with a generic `Your file failed to upload` message.  Do you think it might be worth setting this as an .env variable that gets passed to the front end to make it more configurable?

mbertrand (commented):
Overall works great, one question about the max size limit.

abdulkdawson:
@gsidebo This is looking pretty good to me.

static/js/constants.js
gsidebo: @mbertrand Great idea

mbertrand (approved):
👍 - probably should also open a devops/salt issue to set `MAX_FILE_UPLOAD_SIZE` to match whatever nginx allows in each environment.

PR #1015, opened 2020-09-22
Configured IdP for NovoEd to enable login via SAML

### LINKED ISSUE #1008: NovoEd SSO
We want to enable single sign-on with NovoEd so users can access a NovoEd course without extra authentication once they've logged in on the bootcamps app.

gsidebo:
Closes #1008 

#### What's this PR do?
Configure bootcamps as a SAML Identity Provider for NovoEd's Service Provider to allow single sign-on to NovoEd when a user logs in on bootcamps

This can be tested in the PR build: https://bootcamp-ecommerce-ci-pr-1015.herokuapp.com/

- Create a user
- Enroll that user in the `test-bootcamp` course using a Django shell (`novoed.api.enroll_in_novoed_course` - use `test-bootcamp` for the novoed stub value)
- Visit the SP-initiated login URL: https://app.novoed.com/saml/sso?provider=mitbootcamps

This should redirect you to bootcamps, then instantly redirect you to NovoEd since you're already logged in. You should see a dashboard when you land back on NovoEd's app with the `test-bootcamp` course displayed

Also try this in an incognito window to make sure that the redirection works when you are not already logged in


config/nginx.conf.erb
```diff
@@ -81,6 +81,9 @@ http {
             uwsgi_param SERVER_PORT $server_port;
             uwsgi_param SERVER_NAME $server_name;
             uwsgi_pass unix:/tmp/nginx.socket;
+            uwsgi_buffer_size 16k;
```
noisecapella, line 84: Is this just common settings which weren't added before, or is there a need specific to this PR?

main/settings.py
```diff
@@ -872,3 +876,77 @@ def get_all_config_keys():
...
+                        (f"{IDP_BASE_URL}/sso/post", saml2.BINDING_HTTP_POST),
+                        (f"{IDP_BASE_URL}/sso/redirect", saml2.BINDING_HTTP_REDIRECT),
+                    ],
+                    "single_logout_service": [
+                        (f"{IDP_BASE_URL}/slo/post", saml2.BINDING_HTTP_POST),
+                        (f"{IDP_BASE_URL}/slo/redirect", saml2.BINDING_HTTP_REDIRECT),
+                    ],
+                },
+                "name_id_format": [NAMEID_FORMAT_UNSPECIFIED, NAMEID_FORMAT_PERSISTENT],
+                "sign_response": False,
```
noisecapella, line 935: Does this need to be false? In the example code in the docs it's set to true, but maybe the requirements are different here

main/settings.py
```diff
@@ -872,3 +876,77 @@ def get_all_config_keys():
...
+                    "single_sign_on_service": [
+                        (f"{IDP_BASE_URL}/sso/post", saml2.BINDING_HTTP_POST),
+                        (f"{IDP_BASE_URL}/sso/redirect", saml2.BINDING_HTTP_REDIRECT),
+                    ],
+                    "single_logout_service": [
+                        (f"{IDP_BASE_URL}/slo/post", saml2.BINDING_HTTP_POST),
+                        (f"{IDP_BASE_URL}/slo/redirect", saml2.BINDING_HTTP_REDIRECT),
+                    ],
+                },
+                "name_id_format": [NAMEID_FORMAT_UNSPECIFIED, NAMEID_FORMAT_PERSISTENT],
```
noisecapella, line 934: Does the order matter here? I'm having trouble finding info about this in the docs

config/nginx.conf.erb
gsidebo: Yup, the SAML requests contain some rather long XML strings, so we needed to increase this to avoid nginx errors

noisecapella:
Functionality works great

main/settings.py
gsidebo: This setting might not actually be needed since this is an IdP config, not an SP. I'll test

main/settings.py
gsidebo: Doesn't matter. The requirement is just that any one of the values must match the name ID format requested by the SP

main/settings.py
gsidebo: @noisecapella Actually this is just a default. The `ServiceProvider` record has a property that overrides this if you specify it (and we do). No risk leaving this as `False` for the moment

gsidebo:
@noisecapella Anything else?

noisecapella (approved):

PR #1021, opened 2020-09-23
Add admin for WireTransferReceipt

### LINKED ISSUE #1020: Add admin for WireTransferReceipt
This was overlooked when implementing #924

noisecapella:
Fixes #1020 

#### What's this PR do?
Adds an admin for WireTransferReceipt. I largely copied this from the existing admin for Receipt

View the WireTransferReceipt in the django admin. If you don't have one, try creating one manually in the Django shell, or use the `import_wire_transfers` command.

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -171,6 +171,9 @@ class WireTransferReceiptAdmin(TimestampedModelAdmin):
     def has_delete_permission(self, request, obj=None):
         return False
 
+    def has_change_permission(self, request, obj=None):
+        return False
+
     def get_queryset(self, request):
         """Overrides base queryset"""
         return super().get_queryset(request).select_related("order__user")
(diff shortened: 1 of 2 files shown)


ecommerce/admin.py
```diff
@@ -156,7 +156,62 @@ def order_link(self, obj):
...
+    readonly_fields = get_field_names(WireTransferReceipt) + ["order_link"]
+    list_display = ("id", "get_user_email", "order_link", "get_order_status")
+    search_fields = ("order__user__email", "order__user__username")
+
+    def has_add_permission(self, request):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
+
```
mbertrand, line 173: Since there aren't any editable fields, maybe add this to remove the `Save` buttons:

```python
def has_change_permission(self, request, obj=None):
        return False
```

mbertrand (changes requested):
Looks great, one minor suggestion

mbertrand (approved):
👍

Issue #1016, opened 2020-09-23
Optimize product and home page context for section pages


ahmed-belal:
Accessing the section child pages in the template like:
`{% if page.learning_resources %} {% include 'partials/learning-resources.html' page=page.learning_resources %} {% endif %}`

executes two queries. Please populate the context with these attributes as new context vars so they can be directly accessed and the queries are run once per section like so:
`{% if learning_resources%} {% include ..... page=learning_resources %} {% endif %}`

For reference, this is how it's done in xPRO:
https://github.com/mitodl/mitxpro/blob/7d9b032ea669d7ef1980a5d12ddb8a417dc50ca1/cms/models.py#L436

PR #1023, opened 2020-09-24
Optimize Profile and Home page context

HamzaIbnFarooq:
Fixes: #1016 

#### What's this PR do?
Optimize Profile and Home page context

Navigate to Home and Product pages and check if the sequence of subsections is correct, along with its data.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -127,9 +127,15 @@ class HomePage(Page, CommonProperties):
     def get_context(self, request, *args, **kwargs):
         return {
             **super().get_context(request, *args, **kwargs),
+            "CSOURCE_PAYLOAD": None,
             "site_name": settings.SITE_NAME,
             "title": self.title,
-            "CSOURCE_PAYLOAD": None,
+            # properties
+            "alumni": self.alumni,
+            "catalog": self.catalog,
+            "learning_resources": self.learning_resources,
+            "program_description_section": self.program_description_section,
+            "three_column_image_text_section": self.three_column_image_text_section,
         }
 
     @property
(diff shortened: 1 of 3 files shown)


ahmed-belal:
@HamzaIbnFarooq It would be better to include "verify that there are no duplicate queries on the home and product page rendering" as that is the original issue being fixed.

ahmed-belal (approved):
LGTM

HamzaIbnFarooq:
@ahmed-belal, After profiling Home and Product pages, with current PR's fixes implemented, I could only find some image related duplicate queries which seems like a technical limitation. Other than that everything seems in-place.

cms/models.py
```diff
@@ -127,9 +127,14 @@ class HomePage(Page, CommonProperties):
     def get_context(self, request, *args, **kwargs):
         return {
             **super().get_context(request, *args, **kwargs),
+            "alumni": self.alumni,
+            "catalog": self.catalog,
+            "CSOURCE_PAYLOAD": None,
+            "learning_resources": self.learning_resources,
+            "program_description_section": self.program_description_section,
             "site_name": settings.SITE_NAME,
+            "three_column_image_text_section": self.three_column_image_text_section,
```
gsidebo, line 136: One very minor suggestion: can you put the variables that are not properties of this class above the other variables that are properties?
```suggestion
            "site_name": settings.SITE_NAME,
            "CSOURCE_PAYLOAD": None,
            "alumni": self.alumni,
            "catalog": self.catalog,
            "learning_resources": self.learning_resources,
            "program_description_section": self.program_description_section,
            "three_column_image_text_section": self.three_column_image_text_section,
```

gsidebo (changes requested):

cms/models.py
HamzaIbnFarooq: @gsidebo, Thanks for the review.
I have reordered the variables along with adding a comment, kindly have a look now.

gsidebo (approved):
Looks great! 👍

PR #1037, opened 2020-10-02
Added setting for overriding SESSION_ENGINE

gsidebo:
Related to #1008 

#### What's this PR do?
Changes the `SESSION_ENGINE` setting to be configurable via env var

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -185,7 +185,17 @@ AUTHENTICATION_BACKENDS = (
     "django.contrib.auth.backends.ModelBackend",
 )
 
-SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"
+SESSION_ENGINE_BACKEND = get_string(
+    "SESSION_ENGINE_BACKEND",
+    "signed_cookies",
+    description=(
+        "The backend that will support user sessions. This should be a module within django.contrib.sessions.backends. "
+        "Possible values: signed_cookies, db, cached_db, cache, file. "
+        "(https://docs.djangoproject.com/en/3.1/topics/http/sessions/#configuring-the-session-engine)"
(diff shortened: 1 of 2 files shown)


main/settings.py
```diff
@@ -185,7 +185,15 @@
...
 
-SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"
+SESSION_ENGINE = get_string(
+    "SESSION_ENGINE",
+    "django.contrib.sessions.backends.signed_cookies",
+    description=(
+        "The backend that supports user sessions "
+        "(https://docs.djangoproject.com/en/3.1/topics/http/sessions/#configuring-the-session-engine)"
+    ),
+    required=True,
```
rhysyngsun, line 195: Can you make this `False` instead? It'll cause the app to fail to boot otherwise and there's a reasonable default.

gsidebo: Misunderstanding on my end. Fixed

blarghmatey:
Should we make this a dictionary mapping? So, rather than the full module path we populate the env var with values such as `cookie`, `cache`, `db`?

gsidebo:
@blarghmatey I don't feel too strongly either way. I think we're only somewhat more likely to have a typo or something in the full module path than just the single module name

blarghmatey:
I don't have a strong conviction on that, just a suggestion to improve ergonomics to prevent having to look up the possible values. Might be worth at least adding a comment with some of the options, or a link to relevant docs?

gsidebo:
@blarghmatey Just updated with that suggestion and possible values

rhysyngsun (approved):

blarghmatey:
Looks good to me

PR #1041, opened 2020-10-06
Added link to NovoEd from application dashboard

### LINKED ISSUE #1031: Add NovoEd bootcamp links to dashboard
If a user is enrolled an a bootcamp that is served through NovoEd, we should provide a link for that user to access the course on NovoEd directly

https://projects.invisionapp.com/d/main?origin=v7#/console/19928472/424654864/preview?scrollOffset=738

gsidebo:
Closes #1031 

#### What's this PR do?
Adds a link from Bootcamps to NovoEd if a course is served through NovoEd and they are enrolled

- Add `NOVOED_SAML_LOGIN_URL=https://app.novoed.com/saml/sso?provider=mitbootcamps` to your `.env`
- Make sure you have an application started for some bootcamp run
- Add a `BootcampRunEnrollment` record for some user, then expand the detail section of that run in your app dashboard. You should see a "Start Bootcamp" link at the bottom going to that SAML URL above

- **These "start bootcamp" links won't work locally.** That app.novoed link is the login link on production, which is already known to work.
- It's OK if there are unchecked bubbles above the "Bootcamp Starts" section. We have some scenarios where users have an enrollment despite not having completed all of the standard application steps. We will correct this UI wonkiness in a separate issue.
- This link will bring users to their NovoEd dashboard, which has a link to the enrolled bootcamp. Ideally we would link our users directly to the bootcamp in NovoEd, but I need to hear back from NovoEd about how we can do that. This dashboard link is sufficient in the meantime.

ss 2020-10-06 at 09 01 07

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -872,6 +872,11 @@ NOVOED_API_SECRET = get_string(
 NOVOED_API_BASE_URL = get_string(
     "NOVOED_API_BASE_URL", None, description="The base URL of the NovoEd API"
 )
+NOVOED_SAML_LOGIN_URL = get_string(
+    "NOVOED_SAML_LOGIN_URL",
+    None,
+    description="The SP-initiated SAML login URL for NovoEd",
+)
 
 # Relative URL to be used by Djoser for the link in the password reset email
 # (see: http://djoser.readthedocs.io/en/stable/settings.html#password-reset-confirm-url)
--- a/static/js/pages/applications/ApplicationDashboardPage.js
+++ b/static/js/pages/applications/ApplicationDashboardPage.js
@@ -15,6 +15,7 @@ import qs from "query-string"
 import { reverse } from "named-urls"
 
 import {
+  BootcampStartDetail,
   PaymentDetail,
   ProfileDetail,
   QuizDetail,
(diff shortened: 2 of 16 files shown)


main/settings.py
```diff
@@ -872,6 +872,16 @@ def get_all_config_keys():
 NOVOED_API_BASE_URL = get_string(
     "NOVOED_API_BASE_URL", None, description="The base URL of the NovoEd API"
 )
+NOVOED_MIT_BASE_URL = get_string(
```
mbertrand, line 875: This doesn't seem to be used for anything?

static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -426,12 +427,27 @@ export class ApplicationDashboardPage extends React.Component<Props, State> {
...
+    const isNovoEdCourse = !!applicationDetail.bootcamp_run.novoed_course_stub
+    const novoEdEnrolled =
+      isNovoEdCourse &&
+      !!SETTINGS.novoed_login_url &&
+      !!applicationDetail.enrollment &&
+      !!applicationDetail.enrollment.novoed_sync_date
+    const bootcampStartRow = isNovoEdCourse ? (
+      <BootcampStart
+        ready={novoEdEnrolled}
+        fulfilled={novoEdEnrolled}
```
mbertrand, line 439: Looks like `ready` and `fulfilled` will always be the same, so one of them could be removed as a prop.  Or is there a future condition/use-case where that might not be true?

mbertrand (changes requested):

static/js/pages/applications/ApplicationDashboardPage.js
gsidebo: Yup they'll always be the same (for now), but the rest of the detail sections follow a pattern where they receive those two props and they can be different values. To me it makes sense to keep the props consistent even if at the moment they will always be given the same value

main/settings.py
gsidebo: Thought I was going to need this initially. Removed 👍

gsidebo:
@mbertrand Those annoying `caniuse-lite is outdated` warnings finally turned into an error. I fixed that in [this PR](https://github.com/mitodl/bootcamp-ecommerce/pull/1042) and rebased this PR on it. If you have an extra minute, you could review that PR as well

mbertrand (approved):
👍

Issue #1050, opened 2020-10-13
Support SAML login with staging config


gsidebo:
Related to #1015 

Logging into NovoEd via SAML in two different environments leaves us open to issues where user ids/usernames exist in both environments. We need to alter the config so that properties coming from RC (our SAML staging environment) and production are reliably unique

pdpinch:
@gsidebo did #1051 take care of this?

gsidebo:
@pdpinch It did for new users. I'll close this and open a new issue for handling users that have an email that has already been used to log into NovoEd previously

PR #1051, opened 2020-10-13
Changed SAML config to use different identifiers for staging

gsidebo:
Hopefully addresses #1050

#### What's this PR do?
Changes SAML config so that we send a prefixed unique identifier when in the staging environment (in our case that's RC)

No manual testing. This will be formally tested in RC. If it doesn't fix the issue, further PRs will be opened


mbertrand (approved):
👍

PR #1052, opened 2020-10-14
Prevented payment if bootcamp run start date is in the past

gsidebo:
No ticket

#### What's this PR do?
Fixes issue where users are still able to pay after the bootcamp run has started

Set your application to the awaiting payment state, and set the bootcamp run start date in the future. You should see a link to submit payment. Then set the start date to the past and reload the page. That link should no longer be shown

There's a code comment that explains why we're using bootcamp run start date instead of installment deadlines to determine whether or not a user should be shown a payment link

### DIFF
--- a/klasses/models.py
+++ b/klasses/models.py
@@ -7,6 +7,7 @@ from django.contrib.auth.models import User
 from django.db import models
 
 from main.models import TimestampedModel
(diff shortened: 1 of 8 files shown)


mbertrand (approved):
👍

Issue #1053, opened 2020-10-15
node-fetch dependency upgrade to 2.6.1 or above


HamzaIbnFarooq:
[update node-fetch on bootcamps](https://github.com/mitodl/bootcamp-ecommerce/network/alert/yarn.lock/node-fetch/open)

PR #1054, opened 2020-10-15
node-fetch dependency upgrade to version 2.6.1

HamzaIbnFarooq:
Fixes: https://github.com/mitodl/bootcamp-ecommerce/issues/1053

#### What's this PR do?
Upgrades node-fetch dependency to 2.6.1

#### Note
`formik 1.5.8` requires `node-fetch@^1.0.1` and if we upgrade `formik` to its `latest version 2.x.x our code starts breaking`.
`flow-typed` is at `3.2.1` (which is the latest version) and requires `node-fetch@^2.3.0`.


umarmughal824:
@HamzaIbnFarooq the ideal way to for updating the parent dependencies first. If we don't find that way fix, then we could go for resolution. So my question is that how you tried the first approach?

HamzaIbnFarooq:
@umarmughal824 In the PR you will see that I have tried to increase the version of the relevant parent dependencies which could be upgraded. But I found some limitations which I mentioned in the description of this PR (under the **Note** section). So the resolution is meant to handle those limitations.

umarmughal824:
@HamzaIbnFarooq in my opinion we should go for addressing those issues for `formik`. Let me see how many errors it's generating.

umarmughal824 (changes requested):
Please address my comments

HamzaIbnFarooq:
> @HamzaIbnFarooq in my opinion we should go for addressing those issues for `formik`. Could you please explore how it got fixed in `mitxpro` without updating the `formik`. In that way, we might save ourself from updating `formik` and its errors in the latest version.

The node-fetch dependency related to formik is not resolved in xpro as well. You may check xpro's master branch.

umarmughal824:
> dependency

It is resolved, if it had not been resolved we had the security alert for it.

umarmughal824 (changes requested):
I tried that locally, by only updating the `codcov` to `3.8.0` that is the latest version has fixed the security alert. So please remove that resolution part of the PR. 

The rest updates of the packages are fine if they don't break any of our functionality.

Steps to perform:

- [x] 1. remove that resoultion part
- [ ] 2. then re-run the yarn refresh packages command (`yarn --check-files`) to update the `yarn.lock`
- [ ] 3. please remember to test the entire system is still functional with it

HamzaIbnFarooq:
@umarmughal824 I have updated the PR according to the 3 steps you mentioned.

umarmughal824 (approved):
👍 , please don't forget to squash and rebase first.

PR #1058, opened 2020-10-15
Changed NovoEd API to update the sync date if a NovoEd enrollment already exists

### LINKED ISSUE #1056: NovoEd sync date should be updated if it's empty and the enrollment already exists
We have some API logic that attempts to create an enrollment for some course in NovoEd (`novoed.api.enroll_in_novoed_course`). If the enrollment already exists on NovoEd's side, it's basically a no-op on our end. If the `BootcampRunEnrollment` in question has a blank `novoed_sync_date` value, we should update it in those cases. That value is used to determine whether or not we can show a NovoEd link in the user's dashboard

gsidebo:
Closes #1056 

#### What's this PR do?
Updates the `novoed_sync_date` property for an enrollment if we get a response from NovoEd indicating the enrollment already exists on their end, and that property is currently blank

This involves a bunch of interaction with NovoEd that might make it too complicated for the small size/impact of the PR. We can get away with testing this on RC and leaving this review as code review only

That `novoed_sync_date` property determines whether or not we show a NovoEd link in a user's dashboard. In cases where the user has a NovoEd enrollment but their local `BootcampRunEnrollment` record does not reflect that, this PR corrects for that state and makes sure the user will see the NovoEd link in their dashboard

### DIFF
--- a/novoed/api_test.py
+++ b/novoed/api_test.py
@@ -5,7 +5,9 @@ import pytest
 from rest_framework import status
 from requests.exceptions import HTTPError
 
+from klasses.factories import BootcampRunEnrollmentFactory
 from main.test_utils import MockResponse
+from main.utils import now_in_utc
 from profiles.factories import UserFactory
 from novoed.api import enroll_in_novoed_course, unenroll_from_novoed_course
 from novoed.constants import REGISTER_USER_URL_STUB, UNENROLL_USER_URL_STUB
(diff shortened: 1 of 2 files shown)


novoed/api_test.py
```diff
@@ -41,15 +42,29 @@ def patched_post(mocker):
...
+    exp_existing,
+    exp_update_sync_date,
+):  # pylint:disable=too-many-arguments
     """
     enroll_in_novoed_course should make a request to enroll a user in NovoEd and return flags indicating the results
     """
+    enrollment = BootcampRunEnrollmentFactory.create(
+        user=novoed_user,
+        bootcamp_run__novoed_course_stub=FAKE_COURSE_STUB,
+        novoed_sync_date=None,
```
mbertrand, line 66: Is it worth adding a test where `novoed_sync_date` is not `None`, to make sure it doesn't get updated?

mbertrand (changes requested):
Looks great, suggested another test but if it's not worth the effort then that's ok with me.

novoed/api_test.py
gsidebo: Added (sorry, forgot about this PR!)

mbertrand (approved):
👍

PR #1060, opened 2020-10-19
Added cms-login to bootcamp-login redirection

### LINKED ISSUE #1059: CMS: redirect the wagtail login to the site login
CMS users are confused by the wagtaiil login screen, when they should be logging in through /signin

We set up a similar redirect on xPRO like this: https://github.com/mitodl/mitxpro/pull/642

arslanashraf7:
Fixes #1059 

#### What's this PR do?
This PR is an attempt to fix the ambiguity issue where we had two different login pages for site and cms. Now whenever users would open CMS's login page he will they will be redirected to the login page of the site instead of cms, hence creating a single entry point for login.

Open login page of cms. e.g. log out of bootcamps and then hit `/cms` URL. you should be redirected to the site's login page instead.

### DIFF
--- a/main/views.py
+++ b/main/views.py
@@ -3,10 +3,11 @@ bootcamp views
 """
 import json
 
+from django.contrib.auth.views import redirect_to_login
 from django.shortcuts import render, redirect, reverse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
-
+from main import settings
 from main.utils import has_all_keys
 
 
@@ -96,3 +97,10 @@ def page_500(request, *args, **kwargs):  # pylint: disable=unused-argument
     Overridden handler for the 404 error pages.
     """
     return standard_error_page(request, 500, "500.html")
+
+
+def cms_login_redirect_view(request):
+    """
+    Redirects wagtain's login page to site's login page
+    """
+    return redirect_to_login(reverse("wagtailadmin_home"), login_url=settings.LOGIN_URL)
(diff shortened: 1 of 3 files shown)


umarmughal824 (changes requested):

main/views.py
```diff
@@ -3,10 +3,11 @@
 """
 import json
 
+from django.contrib.auth.views import redirect_to_login
 from django.shortcuts import render, redirect, reverse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
-
+from django.conf import settings
```
umarmughal824, line 10: Please remember to keep following the sequence of imports in the order. So it should come before the `redirect_to_login` import.

main/views_test.py
```diff
@@ -61,3 +61,12 @@ def test_cybersource_context(client, user):
         "bootcamp_run_purchased": fake_title,
         "purchase_date_utc": fake_datetime,
     }
+
+
+@pytest.mark.django_db
```
umarmughal824, line 66: That is not required as well.

umarmughal824 (changes requested):

main/views_test.py
arslanashraf7: @umarmughal824  The test fails without this annotation. Since client redirection is unsuccessful while ensuring connection because of database access failure giving `Failed: Database access not allowed, use the "django_db" mark` error. 
Can you please suggest another approach that should be followed here?

umarmughal824: If that error is occurred, keep it intact.

umarmughal824 (approved):
:+1:, pls don't forget to squash your commits.

PR #1064, opened 2020-10-21
add copy bootcamp feature

HamzaIbnFarooq:
fixes https://github.com/mitodl/bootcamp-ecommerce/issues/1057
#### What's this PR do?
It allows the creation of a `new Bootcamp Page` without requiring `BootcampRun`.

* Copy a BootcampRunPage through cms.
* You will observe that a new run page has been created and it has no bootcamp_run selected.

Following are the steps of the new use-case:
* Copy an existing Bootcamp Run Page
image
* Change its Slug and Title according to your need and click `Copy this Page`
image
* You will see a new Bootcamp Run Page entry
image
* Open the new Bootcamp Run Page and select the required Bootcamp run using the dropdown.
image


pdpinch:
This approach seems perfect to me. @gsidebo any concerns?

cms/models.py
```diff
@@ -244,10 +244,15 @@ class BootcampRunPage(BootcampPage):
     bootcamp_run = models.OneToOneField(
         "klasses.BootcampRun",
         null=True,
+        blank=True,
```
asadiqbal08, line 247: @HamzaIbnFarooq do not forget to create the migration file for this change.

asadiqbal08 (changes requested):

asadiqbal08:
Otherwise that approach also seems me good.

cms/models.py
HamzaIbnFarooq: Thanks for highlighting that, I just added it.

cms/models.py
```diff
@@ -244,10 +244,15 @@ class BootcampRunPage(BootcampPage):
...
         on_delete=models.SET_NULL,
         help_text="The bootcamp run for this page",
     )
 
+    def copy(self, *args, **kwargs):  # pylint: disable=signature-differs
+        kwargs["exclude_fields"] = kwargs.get("exclude_fields", []) + ["bootcamp_run"]
```
gsidebo, line 253: This works well, but copied bootcamp run pages should be unpublished by default since (a) they don't have a run associated with them when they're first copied, and (b) they still have text/images/etc. from an existing bootcamp run page. I tinkered with this a little and I couldn't find an easy way to validate these things before publishing, but we can pass in an extra kwarg here that will default the page to be unpublished. That will make it less likely for us to have a published bootcamp run page that isn't yet ready to be viewed.

```suggestion
        kwargs["exclude_fields"] = kwargs.get("exclude_fields", []) + ["bootcamp_run"]
        kwargs["keep_live"] = False
```

gsidebo (changes requested):

cms/models.py
HamzaIbnFarooq: Thanks @gsidebo for the detailed review. I have added the required change.

Just a question though: do we need to ask the `course authoring team` about the `Pages` they are usually copying so we can override their copy method as well to ensure that user experience never gets affected on prod, by this at least?

gsidebo: @HamzaIbnFarooq We haven't had any issues yet with the existing workflow, so I think we can wait until an issue comes up before making any changes to other copy methods

gsidebo (approved):
Looks great! 👍

PR #1077, opened 2020-11-16
Added command to migrate applications from one run to another

### LINKED ISSUE #1073: Add management command to migrate applications
We have a request from the bootcamp admin team to migrate users' completed applications from the first run of a bootcamp to a new run. The goal is to make it so these users don't have to fill out a brand new application, and the newly-created application for the new run would be in a state where they could pay right away.

gsidebo:
Closes #1073 

#### What's this PR do?
Adds a management command to migrate applications from one bootcamp run to another

1. Find or create a bootcamp that has at least 2 runs
2. Create an application for an earlier run of that bootcamp and set it to complete
3. Run the `migrate_applications` command, specifying the correct `--from-run` and `--to-run` (you can use the run id or title)

This should create a new application in the 'to' run for your user. It should be in the `AWAITING_PAYMENT` state, it should have the same resume info, the same submissions, and those submissions should be approved. The command should be idempotent, so running it again should not change any data.

You can also test the `--users` parameter by adding a comma-separated string of emails to limit the migration to specific users


applications/management/utils.py
```diff
@@ -0,0 +1,122 @@
...
+        to_run_application.save()
+
+        # Copy submissions (video interview, etc.)
+        from_app_step_submissions = ApplicationStepSubmission.objects.filter(
+            bootcamp_application=from_run_application
+        ).order_by("run_application_step__application_step__step_order")
+        for from_app_step_submission in from_app_step_submissions:
+            ApplicationStepSubmission.objects.update_or_create(
+                bootcamp_application=to_run_application,
+                run_application_step=from_app_step_submission.run_application_step,
```
mbertrand, line 108: This needs to be linked to the new run's application step instead.  Otherwise the new submission ends up pointing to the original run:

Screen Shot 2020-11-16 at 2 11 42 PM

and although the application has a status of 'AWAITING_PAYMENT', the UI expects a new Video interview:

Screen Shot 2020-11-16 at 2 10 42 PM

mbertrand (changes requested):

applications/management/utils.py
gsidebo: Great catch

gsidebo: @mbertrand Your comment made me realize that it would be a good idea to handle cases where the 'from' and 'to' bootcamps have repeated steps and/or differently-ordered steps. I added some logic to handle those cases correctly

applications/api_test.py
```diff
@@ -66,6 +66,13 @@ def test_derive_application_state():
     app.refresh_from_db()
     assert derive_application_state(app) == AppStates.AWAITING_USER_SUBMISSIONS.value
 
+    # The resume requirement should be considered fulfilled if the user uploads a resume *or* provides a LinkedIn URL
+    app.resume_file = None
+    app.linkedin_url = "http://example.com/linkedin"
```
mbertrand, line 71: The `linkedin_url` will be a non-null value by default for `BootcampApplicationFactory` (`linkedin_url = Faker("url")`) - so that is currently causing the test condition on line 62 to fail.  Either changing the factory linked_in url to be None by default or modifying these tests is fine with me.

mbertrand (changes requested):

applications/api_test.py
gsidebo: Hmph... don't know how I missed this. Fixed now!

mbertrand (approved):
👍

PR #1079, opened 2020-11-18
Added 'state' param to migrate_applications command

gsidebo:
No ticket - adds a param to #1073 

#### What's this PR do?
Adds a parameter to the `migrate_applications` command so applications can be filtered by the `state` value

Same steps as #1073. Just add a `--state` param to the command and make sure it filters the applications correctly


mbertrand (approved):
👍

PR #927, opened 2020-07-16
Added seed data scripts and commands

gsidebo:
#### What are the relevant tickets?
Closes #490 

#### What's this PR do?
Adds seed data script for creating/deleting database objects

#### How should this be manually tested?
Run the `seed_data` and `delete_seed_data` management commands. Check the app and Django admin to see what was created

#### Any background context you want to provide?
I had a few goals with the json structure for describing the seed data objects
1. Make it readable enough that you can easily understand what objects are going to be created
1. Avoid requiring field values to be specifically provided when they can be set dynamically according to some other values/attributes or given sensible defaults 
1. Where appropriate, set dates dynamically to avoid situations where we define static dates that become stale over time

I can add the the documentation for this feature and enable custom seed data json files, but I wanted to put this up first.

#### Screenshots (if appropriate)
ss 2020-07-16 at 10 20 50

PR #1084, opened 2020-11-20
Added seed data command for setting application state

gsidebo:
No ticket - adds some seed data functionality alongside the command added in #927 

#### What's this PR do?
Adds a management command to set an application to a desired state

Create an application for some user and bootcamp run, then run the management command to set the state - ex:`set_application_state -i <your application id> --state AWAITING_PAYMENT`. Check that the application fields have the expected data, including related objects like submissions. From there you can use different states for the command and confirm accordingly.

ss 2020-11-20 at 15 52 25


mbertrand (commented):
LGTM.   Noticed that reversing an application's state works too - nice!   One suggestion, take it or leave it - what do you think of adding a check on `settings.ENVIRONMENT` to make sure this isn't run on production - especially for reverting complete applications and deleting orders?

gsidebo:
@mbertrand Great call. Will do

mbertrand (approved):
👍

PR #1085, opened 2020-11-20
Added flag to run python tests only without pylint/cov/warnings

gsidebo:
No ticket
#### What's this PR do?
Adds a pytest flag that allows us to run Python tests only without pylint and coverage, and with warnings suppressed

Run bash in a web container, then run `pytest --simple`. You should see the test suite run without pylint, without a coverage report, and with no warnings output

This was done for xpro in this PR: https://github.com/mitodl/mitxpro/pull/1983

This flag will be particularly helpful while hacking on tests in a local branch. Most of the time you just need the tests to run and to have the console output limited to which tests passed and failed.

ss 2020-11-20 at 16 33 01

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -8,9 +8,37 @@ from fixtures.cybersource import *
 TEST_MEDIA_ROOT = "/var/media/test_media_root"
 
 
-def pytest_configure():
+def pytest_addoption(parser):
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

PR #1086, opened 2020-11-23
Replacing Github Actions with Travis

HamzaIbnFarooq:
#### What are the relevant tickets?
https://github.com/orgs/mitodl/projects/2

#### What's this PR do?
Uses Github Actions instead of Travis

#### How should this be manually tested?
Anyone can simply push a branch to see if tests are running in GitHub actions and test coverage is being uploaded or not.

**Side Note**: This PR also includes some work regarding **Remove tox, use pytest instead** as done in the following PR https://github.com/mitodl/bootcamp-ecommerce/pull/1072

PR #1087, opened 2020-11-24
Release 0.62.0

odlbot:
## Gavin Sidebottom
  - [x] Added flag to run python tests only without pylint/cov/warnings (#1085) ([da38f9ae](../commit/da38f9ae77084e88ad913014a5c903e16be52d82))
  - [x] Added seed data command for setting application state (#1084) ([9393d9c6](../commit/9393d9c609edd12e61927271e43a07b356c8fa21))

## Hamza Farooq
  - [x] Replacing Travis with Github actions and using pytest instead of tox (#1086) ([aa9779ad](../commit/aa9779ade7c75f6924b28d9b378270d153a8ae11))


gsidebo:
There's an issue with saving resume files in the seed data command, but that doesn't affect the user experience at all, and can be fixed in a separate release.

PR #1088, opened 2020-11-25
Fixed file handling for set_application_state command

gsidebo:
No ticket - fixes issue with #1084 

#### What's this PR do?
Changes the way that the example resume file is handled so it works correctly with S3 as the storage backen

See #1084. You'll just need to run the command to set the state to `AWAITING_USER_SUBMISSIONS` and check that your bootcamp application has a resume file attached


mbertrand (approved):
👍

Issue #1082, opened 2020-11-19
Better error message when profile fails validation 


pdpinch:
### Steps to Reproduce

1. Create a new account
2. You should received a verification email
2. Click on verification link 
3. You should get the profile form 
4. In the profile form, enter a period, `.` as your last name

### Expected Behavior

Assuming that a last name of `.` is unacceptable, I expect an error message that tells me to enter a correct last name. It would be nice if it could tell me what's wrong with the last name I enterred, but that may be too hard. 

### Actual Behavior

Generic error message "There was a problem validating your name and address: Please try again.
If this happens more than once, please contact customer support."

<img width="767" alt="image" src="https://user-images.githubusercontent.com/430126/99704776-d5630800-2a66-11eb-91e9-1ff9c1e9acb4.png">

arslanashraf7:
@pdpinch I was able to reproduce it on RC and looking further it looks like the issue is related to our `Export compliance checks` through `Cybersource`, which is breaking maybe because of `.` sent as the last name for a person.

I tried to look into the details of the `ReasonCodes` being sent by compliance API and I'm not sure if we will specifically be able to identify if the issue was with the name there as there doesn't seem to be a related reason-code specified by  [Compliance API reason codes](https://developer.cybersource.com/library/documentation/dev_guides/Verification_Svcs_SO_API/Verification_Svcs_SO_API.pdf#page=42&zoom=100,96,56)

**One option** seems to be going with field validation e.g. something like (I have put up a branch for these changes as well https://github.com/mitodl/bootcamp-ecommerce/tree/arslan/1082-proper-error-message). Currently, I've checked for `.`'s for `First Name` and `Last Name` field validations, Maybe we could define a `regex` for better validation.

<img width="1109" alt="Screenshot 2020-11-27 at 3 16 06 PM" src="https://user-images.githubusercontent.com/34372316/100439929-f4e0dd00-30c5-11eb-8c1a-7c4648261469.png">


**Other Option** could be to dig in further and try to identify or maybe refactor compliance API consumption. I wasn't able to run the compliance API locally to dig in further.

Let me know of your thoughts.

pdpinch:
@arslanashraf7 I think this is a reasonable approach, given that we know that a last name of `.` won't pass export compliance. 

@rhysyngsun do you have any thoughts?

arslanashraf7:
[...]

PR #1091, opened 2020-12-02
Added transformer for proper compliance error messages on register

arslanashraf7:
Fixes https://github.com/mitodl/bootcamp-ecommerce/issues/1082

#### What's this PR do?
It adds enhancement to show specified proper error messages on the `/create-account` page with respect to the type of error that occurred specifically in export compliance. Previously we were showing a generic message in case export compliance would fail in during register pipeline, Now we try to identify the actual reason and show a proper specified message.

This could be tested following the steps [here](https://github.com/mitodl/bootcamp-ecommerce/issues/1082#issue-746804013).

Start with `/create-account` in bootcamps and follow the steps for account creation.

Previously we were showing the generic error message if any of the register page's field data validation failed in export compliance.

**Below is how different type of errors will populate**
**Note:** Red outline and case messages above are just to highlight the changes for screenshots.

***Single field invalid error message , e.g. First Name**
bootcamp-register-invalid-single-field

***Multiple field invalid error message, e.g. First Name, Last Name**
bootcamp-register-multiple-errors

***We couldn't identify the field but it was surely due to some invalid data**
bootcamp-register-geenric-invalid

***We couldn't identify the field not the type of error, But we are sure that it was during export compliance**
bootcamp-register-generic-message


arslanashraf7:
@pdpinch Can you please have a look at the screenshots attached with details on how the error messages will appear now. Please let me know about the things/texts that might need changes.

pdpinch:
Nice. Can you take a look at how we're handling this situation in xPRO? It would be best to have similar error messages.

rhysyngsun:
I haven't tried to run the code, but a quick read tells me that while this will fix the issue so the user sees an error message, because the social auth pipeline has proceeded past the profile editing action, any updated values the user tries to submit won't be processed so they won't be able to proceed through the registration flow. So ultimately without a bigger change to how we're doing this, their registration is dead-ended either way whether they see a better error message or not.

I think this requires some more thought into how we need to readjust our pipeline design and how the frontend integrates with it. Also there's the partial effort towards [refactoring the compliance app into a modular app](https://github.com/mitodl/ol-django/issues/4) which is where we really should make this kind of change (the compliance pipeline actions should really live in the compliance app itself).

In the meantime, we'd probably be better off adding some frontend validation if there's specific input cases we're aware of. Other than that, we just need to put in the effort to fix this rather than patch over it.

arslanashraf7:
@rhysyngsun Thanks for such valuable insights,

> I haven't tried to run the code, but a quick read tells me that while this will fix the issue so the user sees an error message, because the social auth pipeline has proceeded past the profile editing action, any updated values the user tries to submit...

I tried to follow the same pattern of retry as the existing one https://github.com/mitodl/bootcamp-ecommerce/pull/969, which worked for me for proceeding through the registration flow, But I'm not sure of the direction though.

> I think this requires some more thought into how we need to readjust our pipeline design and how the frontend integrates with it. Also there's the partial effort towards refactoring the compliance app into a modular app which is where we really should make this kind of change (the compliance pipeline actions should really live in the compliance app itself).

This makes sense.

> In the meantime, we'd probably be better off adding some frontend validation if there's specific input cases we're aware of. Other than that, we just need to put in the effort to fix this rather than patch over it.

@pdpinch should I go for field validation as mentioned in **Option one** in this comment https://github.com/mitodl/bootcamp-ecommerce/issues/1082#issuecomment-734769569. If you agree I will put up a PR for field validations on the frontend for now.

pdpinch:
@arslanashraf7 I couldn't quite follow Option 1 that you linked to, but if your plan is to add front-end validation as @rhysyngsun suggested, then I would say please proceed. 

@rhysyngsun could you possibly open an issue (in the ol-django repo?) with what you think would be a more robust solution? I don't think we'd work on it soon, but I'd like to capture the think you all have already done.

arslanashraf7:
Closing in the light of the above conversation and alternate implementation PR https://github.com/mitodl/bootcamp-ecommerce/pull/1096

PR #1094, opened 2020-12-04
Fixed buggy personal price adjustment behavior

### LINKED ISSUE #1013: Cannot make more payments after application status is 'COMPLETE'
If a user's balance due is increased after the original balance is paid (the application status is `COMPLETE`), the user will not be able to make additional payments unless the status is manually changed back to `AWAITING_PAYMENT`.  

If the user tries to make a payment, the API responds with an `invalid status` error message that is not properly handled by the UI, so the user does not get any helpful message to indicate what went wrong.

### LINKED ISSUE #1036: Personal Price: check balance when updating personal price
As an admin, when I update a bootcamper's personal price, I expect their application status to update appropriately. 

For example, if the user has paid $1000 out of $2000 balance due and I update their personal price to $1000, I expect their applicationstatus to be updated to `COMPLETE` and an enrollment to be created for them. 

Another example, if the has paid $2000 out of a $2000 balance due and I update their personal price _up_ to $3000, I expect their application status to be updated to `AWAITING_PAYMENT` so the user can continue making payments. 

See also: #1013

gsidebo:
Fixes #1013
Fixes #1036

#### What's this PR do?
Fixes bug where adjusting a user's personal price did not change the state of their bootcamp application when appropriate 

1. Start out with a user with an application that is awaiting payment (use the `set_application_state` command)
1. Add an `Order`/`Line` to represent a partial payment toward the total amount owed for that run
1. Create/update a personal price for that user/run so that the amount is less than the amount your user has already paid

This is signals-based, so your application should be automatically set to "complete"

Next, do the same, only start off with a fully-paid application in the "complete" state, and set the personal price to an amount that is greater than the amount already paid. Your application should be automatically set to "awaiting payment"

This should also work if you have a personal price in the database then delete it. The original bootcamp run price will be used in this case.

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -30,8 +30,13 @@ from applications.constants import INTEGRATION_PREFIX
 from applications.utils import validate_file_extension
 from ecommerce.models import Order
 from jobma.models import Interview
+from klasses.api import deactivate_run_enrollment
+from klasses.constants import ENROLL_CHANGE_STATUS_REFUNDED
+from klasses.models import BootcampRunEnrollment
+from main import features
 from main.models import TimestampedModel, ValidateOnSaveMixin
 from main.utils import now_in_utc
+from novoed import tasks as novoed_tasks
 
 
 class ApplicationStep(models.Model):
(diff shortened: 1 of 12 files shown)


mbertrand (changes requested):
One thing that seems to be missing when a personal price change causes an `AWAITING_PAYMENT` application to change to `COMPLETE` - no `BootcampRunEnrollment` is created and `novoed_tasks.enroll_users_in_novoed_course` is not called (see `ecommerce.api.complete_successful_order`).  Also not sure what should happen to an existing `BootcampRunEnrollment` if a price change makes it switch from `COMPLETE` to `AWAITING_PAYMENT` - nothing?  Deactivated?

gsidebo:
@mbertrand Great catch. I'll address that now

mbertrand (approved):
👍

PR #1096, opened 2020-12-08
Added field validation on first and last name

### LINKED ISSUE #1082: Better error message when profile fails validation

arslanashraf7:
Fixed #1082 

#### What's this PR do?
It adds validation on the first and last name of the user in the registration form.

**Note:**
While the associated ticket shows Period `.` as invalid input for the first name but in this PR I tried to generate a regex to handle possible invalid values for both first and last name fields and added validation on the front end so that error-prone input doesn't pass through compliance API.

Reference details mentioned by cybersource could be found [here](https://developer.cybersource.com/library/documentation/dev_guides/EChecks_SCMP_API/html/Topics/app_fields_SCMP.htm#XREF_53480_SCMP_API_Fields) & [here](https://developer.cybersource.com/library/documentation/dev_guides/Verification_Svcs_SO_API/Verification_Svcs_SO_API.pdf) (Formatting restrictions part)

Since documentation wasn't very clear about the formatting details so I tested it manually,
**Some cases** were identified during the search and testing of compliance API.

Acceptable Special characters were ```~!@&)(+:'.?/,`-```

Non Acceptable special characters were: ```/^$#*=[\]`%_;<>{}"|```

1. API doesn't accept names containing only special characters such as `.`, `...`, `.@~` (Failing case in associated ticket)
2. API doesn't accept names starting with special characters e.g. `@Name` or `.Name`
3. Some of the special characters that ware not accepted If present anywhere in first/last name string were: `#,$,&,^,_`
4. API accepts names starting at least with an `alphanumeric` and then containing any `sequence of alphanumeric or special characters`.

The `regex` used in the PR is created based on the above scenarios.

**Solution Applied:**
To be on the safe side and not to block users, It will only reject those sequences/values about which we will be sure that are invalid based on the above 4 points, For all other values we'll keep it open-ended, we'll let them pass to compliance API and let it decide their validity (Only in this case we'll now see a general error message).

Testing instructions are mentioned in the associated ticket above.

Register account for bootcamps app.

So far, we were handling validation errors from compliance API as generic since we knew we are only validating the user's name and address through compliance during the register so we'd show an error regarding name & address whenever there is a validation error from compliance API.

Screenshot 2020-12-08 at 8 55 36 PM
Screenshot 2020-12-08 at 8 55 54 PM

### DIFF
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -19,7 +19,13 @@ log = logging.getLogger()
 US_POSTAL_RE = re.compile(r"[0-9]{5}(-[0-9]{4}){0,1}")
 CA_POSTAL_RE = re.compile(r"[A-Z]\d[A-Z] \d[A-Z]\d$", flags=re.I)
 USER_NAME_RE = re.compile(
-    r"^(?![~!@&)(+:'.?/,`-]+)([^/^$#*=\[\]`%_;<>{}\"|]+)$", flags=re.I
+    r"""
+    ^                               # Start of string
+    (?![~!@&)(+:'.?/,`-]+)          # String should not start from character(s) in this set - They can exist in elsewhere
+    ([^/^$#*=\[\]`%_;<>{}\"|]+)     # String should not contain characters(s) from this set - All invalid characters
+    $                               # End of string
+    """,
+    flags=re.I | re.VERBOSE | re.MULTILINE,
 )
 
 User = get_user_model()
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -18,6 +18,9 @@ log = logging.getLogger()
 
 US_POSTAL_RE = re.compile(r"[0-9]{5}(-[0-9]{4}){0,1}")
 CA_POSTAL_RE = re.compile(r"[A-Z]\d[A-Z] \d[A-Z]\d$", flags=re.I)
+USER_NAME_RE = re.compile(
+    r"^(?![~!@&)(+:'.?/,`-]+)([^/^$#*=\[\]`%_;<>{}\"|]+)$", flags=re.I
+)
 
 User = get_user_model()
 
@@ -37,6 +40,18 @@ class LegalAddressSerializer(serializers.ModelSerializer):
     # only required in the US/CA
     state_or_territory = serializers.CharField(max_length=255, allow_blank=True)
     postal_code = serializers.CharField(max_length=10, allow_blank=True)
+
+    def validate_first_name(self, value):
+        """Validates first name of the user"""
+        if value and not USER_NAME_RE.match(value):
+            raise serializers.ValidationError("First name is not valid")
+        return value
+
+    def validate_last_name(self, value):
+        """Validates last name of the user"""
+        if value and not USER_NAME_RE.match(value):
+            raise serializers.ValidationError("Last name is not valid")
+        return value
 
     def validate_street_address(self, value):
         """Validates an incoming street address list"""
--- a/static/js/lib/validation.js
+++ b/static/js/lib/validation.js
@@ -7,6 +7,7 @@ import {
   CA_POSTAL_CODE_REGEX,
   COUNTRIES_REQUIRING_STATE,
   NAME_REGEX,
+  NAME_REGEX_FAIL_MESSAGE,
   US_ALPHA_2,
   US_POSTAL_CODE_REGEX
 } from "../constants"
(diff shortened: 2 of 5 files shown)


profiles/serializers.py
```diff
@@ -18,6 +18,9 @@
 
 US_POSTAL_RE = re.compile(r"[0-9]{5}(-[0-9]{4}){0,1}")
 CA_POSTAL_RE = re.compile(r"[A-Z]\d[A-Z] \d[A-Z]\d$", flags=re.I)
+USER_NAME_RE = re.compile(
+    r"^(?![~!@&)(+:'.?/,`-]+)([^/^$#*=\[\]`%_;<>{}\"|]+)$", flags=re.I
```
rhysyngsun, line 22: Could you document this regex using [verbose regular expressions](https://diveintopython3.problemsolving.io/regular-expressions.html#verbosere) so it's easy to understand what's going on here at a glance?

static/js/lib/validation.js
```diff
@@ -78,11 +79,13 @@ export const legalAddressValidation = yup.object().shape({
       .string()
       .label("First Name")
       .trim()
+      .matches(NAME_REGEX, "Invalid First Name")
       .required(),
     last_name: yup
       .string()
       .label("Last Name")
       .trim()
+      .matches(NAME_REGEX, "Invalid Last Name")
```
rhysyngsun, line 88: Could we provide some feedback as to how the user might fix this so we can save some effort on the user having to contact support for something they could probably troubleshoot themselves?

profiles/serializers.py
```diff
@@ -38,6 +41,18 @@ class LegalAddressSerializer(serializers.ModelSerializer):
...
+        """Validates first name of the user"""
+        if value and not USER_NAME_RE.match(value):
+            raise serializers.ValidationError("First name is not valid")
+        return value
+
+    def validate_last_name(self, value):
+        """Validates last name of the user"""
+        if value and not USER_NAME_RE.match(value):
+            raise serializers.ValidationError("Last name is not valid")
+        return value
```
rhysyngsun, line 54: Could you add some tests for this?

static/js/lib/validation.js
```diff
@@ -78,11 +79,13 @@ export const legalAddressValidation = yup.object().shape({
       .string()
       .label("First Name")
       .trim()
+      .matches(NAME_REGEX, "Invalid First Name")
```
rhysyngsun, line 82: We should probably have tests for all of the stuff in here, but could you add some at least for this regex validation?

rhysyngsun (changes requested):
Code review, i'll try to functionally test this later today.

profiles/serializers.py
arslanashraf7: Incorporated.

static/js/lib/validation.js
arslanashraf7: Trying to keep it generic and esthetic I have used ```Name cannot start with a special character, And it cannot contain any character from {/^$#*=[]`%_;<>{}}```. Please let me know if this prompt is good enough.

profiles/serializers.py
arslanashraf7: Tests added

static/js/lib/validation.js
arslanashraf7: Test Added

rhysyngsun (approved):

static/js/lib/validation.js
arslanashraf7: @pdpinch FYI, the Above comment, contains the message that will be shown to users on invalid names, I'll merge this PR if you are also good with the above validation error message.

rhysyngsun: @arslanashraf7 could you lowercase the "And", good to go after that

arslanashraf7: @rhysyngsun Sure, I'll update it.

Issue #887, opened 2020-07-07
Add cache control headers to responses


odlbot:
When Fastly caches API responses it can lead to unexpected and unpredictable behavior on the client side. We have currently excluded all API routes from caching in the Fastly configuration, but it would be best to add the appropriate cache control headers in the application to ensure that we are enforcing the intended behavior everywhere.

HamzaIbnFarooq:
Can we elaborate on this ticket a bit more in the following regards:
* Are we willing to use fastly again? if not, do we have any specific header to add?
* What issues were we getting previously and on which specific APIs/Requests?
* Do we want to put any kind of classification on APIs and Views while adding specific headers to them?

pdpinch:
> Are we willing to use fastly again? 

We are using Fastly now, to cache the site in production. When we started seeing issues with the application dashboard, we made a quick fix by adding a setting in Fastly so that all routes starting with `/api` would not be cached. This achieved the desired behavior, but the caching is inefficient and doesn't take into account the nature of particular APIs. 

> if not, do we have any specific header to add?

Take a look at the Fastly guide on caching at https://docs.fastly.com/en/guides/implementing-api-cache-control

We should also have @mitodl/devops review this PR in addition to an engineer. 

> What issues were we getting previously and on which specific APIs/Requests?

See https://mitodl.slack.com/archives/G49GL0CVA/p1593728537385800

In particular, a user could start an application and then reload the page and their application would be lost. This isn't reproducible in a local deployment because you don't have Fastly doing caching. 

> Do we want to put any kind of classification on APIs and Views while adding specific headers to them?

I don't have an opinion about this, but it might be worth doing some research to see if there is best practice for doing this in django.

HamzaIbnFarooq:
While writing a django middleware to change headers, I found that this issue is a duplicate of https://github.com/mitodl/bootcamp-ecommerce/issues/952 and has been solved by https://github.com/mitodl/bootcamp-ecommerce/pull/944

Currently, https://github.com/mitodl/bootcamp-ecommerce/pull/944 checks for all the incoming requests which start with `/api/` path and assigns a `cache-control header to stop caching` that response on browsers and fastly.

So the primary intention of this ticket has been addressed. Although, we can enhance that middleware by categorizing all the urls, present in the platform, into cacheable and cache-restricted, and then we can assign appropriate headers to both of them. But to govern that, we need to analyze each url/endpoint which may or may not worth the effort. 

cc: @pdpinch

[...]

PR #1099, opened 2020-12-15
add dynamic cache controlling via env variable

HamzaIbnFarooq:
Fixes: https://github.com/mitodl/bootcamp-ecommerce/issues/887

#### What's this PR do?
Allows changing cache-control headers via an environment variable CACHEABLE_ENDPOINTS

Add a `List of Endpoints` to a variable named `CACHEABLE_ENDPOINTS` in your local `.env`.
You will observe that cache-control headers are changed in the network tab of the browser.

#### How to configure it on PROD?
Simply add `All the Endpoints` you want to cache to `CACHEABLE_ENDPOINTS` variable as a list in Prod's environment.

https://docs.fastly.com/en/guides/configuring-caching
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

### DIFF
--- a/main/envs.py
+++ b/main/envs.py
@@ -72,7 +72,6 @@ def var_parser(parser_func):
 
         if name in configured_envs:
             raise ValueError(f"Environment variable '{name}' was used more than once")
-
         value = environ.get(name, default)
 
         # attempt to parse the value before we store it in configured_envs
@@ -186,9 +185,10 @@ def parse_list(name, value, default):  # pylint: disable=unused-argument
             parsed_value = [
                 item.strip(" ").lstrip('"').rstrip('"')
                 for item in parsed_value.lstrip("[").rstrip("]").split(",")
+                if item
             ]
         else:
-            parsed_value = [item.strip(" ") for item in parsed_value.split(",")]
+            parsed_value = [item.strip(" ") for item in parsed_value.split(",") if item]
     return parsed_value
 
 
--- a/main/middleware.py
+++ b/main/middleware.py
@@ -9,7 +9,7 @@ class CachelessAPIMiddleware(MiddlewareMixin):
     def process_response(self, request, response):
         """ Add a Cache-Control header to an API response """
         if request.path.startswith(settings.CACHEABLE_ENDPOINTS):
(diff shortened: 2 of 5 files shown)


main/envs.py
```diff
@@ -186,9 +185,10 @@ def parse_list(name, value, default):  # pylint: disable=unused-argument
             parsed_value = [
                 item.strip(" ").lstrip('"').rstrip('"')
                 for item in parsed_value.lstrip("[").rstrip("]").split(",")
+                if item
```
HamzaIbnFarooq, line 188: Previously, if we add a `list` in the `.env` file `e.g: VALUE_LIST=[first_value,]` it will be rendered as `['first_value', '']`. After applying this fix, that anomaly will be gone.

main/middleware.py
```diff
@@ -7,6 +8,8 @@ class CachelessAPIMiddleware(MiddlewareMixin):
 
     def process_response(self, request, response):
         """ Add a Cache-Control header to an API response """
-        if request.path.startswith("/api/"):
+        if request.path.startswith(settings.CACHEABLE_ENDPOINTS):
+            response["Cache-Control"] = "max-age=3600, public"
```
mbertrand, line 12: Should the max-age be configurable via .env variables too?

HamzaIbnFarooq: Rather than creating a variable for max-age, I have introduced `CACHEABLE_ENDPOINTS_CACHE_VALUE` so the whole cache-control value can be changed. Kindly have a look now.

mbertrand (changes requested):
Looks great! Can you also add `CACHEABLE_ENDPOINTS` and `CACHEABLE_ENDPOINTS_CACHE_VALUE` to the `app.json` file?

HamzaIbnFarooq:
> Looks great! Can you also add `CACHEABLE_ENDPOINTS` and `CACHEABLE_ENDPOINTS_CACHE_VALUE` to the `app.json` file?

@mbertrand, that has already been done.

mbertrand:
Sorry, missed it, I need to drink my morning coffee first!

mbertrand (approved):
👍

Issue #1104, opened 2020-12-15
Certificates: management command to create certificates


pdpinch:
As an bootcamp admin, I need a way to create certificates for all bootcampers who have earned a certificate. 

In most cases, this is every bootcamper who is enrolled. There is no grading criteria for bootcamps. 

Part of #1095 Certificates epic

### acceptance criteria

- [x] simple management command that takes a bootcamp run title of id as an input
- [x] created certificate records for all users with an enrollment in the bootcamp run

### related issues

- Management command for revoking an existing certificate for a specific user and bootcamp run
- Management command for blocking the creation of certificate for specific user and bootcamp run

HamzaIbnFarooq:
@pdpinch  Acceptance criteria for this ticket has been covered by https://github.com/mitodl/bootcamp-ecommerce/pull/1111

Now we are only left with `Management command for blocking the creation of certificate for specific user and bootcamp run` we might need to create a separate ticket for that.

PR #1111, opened 2020-12-21
add management commands for certificates

HamzaIbnFarooq:
Fixes https://github.com/mitodl/bootcamp-ecommerce/issues/1104

#### What's this PR do?
add management commands for certificates

You can test these commands via executing management commands in the web container, it supports 4 processes:
1) "./manage.py manage_certificates `--user <USER-ID>` `--run <BOOTCAMP-RUN-ID>` `--revoke`"  # To revoke certificates
2) "./manage.py manage_certificates `--user <USER-ID>` `--run <BOOTCAMP-RUN-ID>` `--unrevoke`"  # To unrevoke certificates
3) "./manage.py manage_certificates `--user <USER-ID>` `--run <BOOTCAMP-RUN-ID>` `--generate`"  # To generate certificates
4) "./manage.py manage_certificates `--run <BOOTCAMP-RUN-ID>` `--generate-batch`" # To generate certificates for everyone whose enrollment is active

### DIFF
--- a/klasses/management/commands/manage_certificates.py
+++ b/klasses/management/commands/manage_certificates.py
@@ -1,0 +1,114 @@
+"""
+Management command to manage certificates.
+"""
+from django.core.management.base import BaseCommand, CommandError
+
+from klasses.api import fetch_bootcamp_run
+from klasses.utils import (
+    generate_single_certificate,
+    generate_batch_certificates,
+    revoke_certificate,
+    unrevoke_certificate,
+)
+from profiles.api import fetch_user
+
+
+class Command(BaseCommand):
+    """
+    Command to manage certificates.
+    """
+
+    help = "Creates, assigns and revokes certificates"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--user",
+            type=str,
+            help="The id, email or username of the enrolled User",
+            required=False,
+        )
+        parser.add_argument(
+            "--run",
+            type=str,
+            help="The 'bootcamprun_id' value for a bootcamprun",
+            required=True,
+        )
+        group = parser.add_mutually_exclusive_group()
+        group.add_argument(
+            "--revoke",
+            help="Flag to revoke single certificate",
+            dest="revoke",
+            action="store_true",
+            required=False,
+        )
+        group.add_argument(
+            "--unrevoke",
+            help="Flag to unrevoke single certificate",
+            dest="unrevoke",
+            action="store_true",
+            required=False,
+        )
+        group.add_argument(
+            "--generate",
+            help="Flag for single certificate generation",
+            dest="generate",
+            action="store_true",
+            required=False,
+        )
+        group.add_argument(
+            "--generate-batch",
+            help="Flag for batch certificate generation",
+            dest="generate_batch",
+            action="store_true",
+            required=False,
+        )
+
+        super().add_arguments(parser)
+
+    def handle(self, *args, **options):  # pylint: disable=too-many-locals
+        """Handle command execution"""
+        try:
+            user = fetch_user(options["user"]) if options["user"] else None
+            generate_single = options.get("generate")
+            generate_batch = options.get("generate_batch")
+            bootcamp_run = fetch_bootcamp_run(options.get("run"))
+            revoke = options.get("revoke")
+            unrevoke = options.get("unrevoke")
+        except:
+            raise CommandError("Provided values are not valid.")
+
+        if generate_single and (not user or not bootcamp_run):
+            raise CommandError(
+                "A valid 'user' and 'run' must be provided with 'generate'."
+            )
+        if generate_batch and not bootcamp_run:
+            raise CommandError("A valid 'run' must be provided with 'generate-batch'.")
+        if (revoke or unrevoke) and (not user or not bootcamp_run):
+            raise CommandError(
+                "A valid 'user' and 'run' must be provided with 'revoke' or 'unrevoke'"
+            )
+
+        if revoke and user and bootcamp_run:
+            result = revoke_certificate(user, bootcamp_run)
+            self.show_message(**result)
+        elif unrevoke and user and bootcamp_run:
+            result = unrevoke_certificate(user, bootcamp_run)
+            self.show_message(**result)
+        elif generate_single and bootcamp_run and user:
+            result = generate_single_certificate(user, bootcamp_run)
+            self.show_message(**result)
+        elif generate_batch and bootcamp_run:
+            result = generate_batch_certificates(bootcamp_run)
+            self.show_message(**result)
+        else:
+            raise CommandError(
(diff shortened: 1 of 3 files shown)


klasses/management/commands/manage_certificates.py
```diff
@@ -0,0 +1,130 @@
...
+                message="Certificate has been un-revoked for user:{} and bootcamp-run:{}".format(
+                    user, bootcamp_run
+                ),
+            )
+        elif generate_single and bootcamp_run and user:
+            updated = generate_single_certificate(user, bootcamp_run)
+            self.show_message(
+                updated,
+                message="Certificate has been created for user:{} and bootcamp-run:{}".format(
+                    user, bootcamp_run
```
pdpinch, line 112: How will `bootcamp_run` be rendered? Will it include the title of the bootcamp run?

HamzaIbnFarooq: Yes it will show something like `Certificate has been created for user:hamza and bootcamp-run:Seed Single Variable Calculus, Nov 18, 2020 - Feb 16, 2021`

HamzaIbnFarooq:
@asadiqbal08 I couldn't find any appropriate model regarding payment of boot-camps so I used BootcampRunEnrollment model instead to create certificates, so kindly let me know if you have any other approach in mind.

asadiqbal08:
@HamzaIbnFarooq  For the MVP: 
- I am thinking to keep this management command as simple as. For that, I suggest if we should keep out the `batch generation feature` for certificates here. Just generate certificate for the given `BootcampRun` and `User`. 

- secondly, I have a question for @pdpinch, As we are not yet implementing the auto generation feature/task here so Is it really required (by this command) to get the payment information of the user before generating the certificate?  Actually, I guess, the purpose of this command is to generate certificate for a `user` for a given `BootcampRun` that will be triggered by the `admin`.

klasses/management/commands/manage_certificates.py
pdpinch: Can we print the email address instead of the user name?

pdpinch:
>Is it really required (by this command) to generate certificates for the paid users only ? In other words, Do we need to get the payment information of the user before generating the certificate?

No, you don't need to get the payment information, which is quite complex. The criteria for granting a certificate is whether the user is enrolled, since there may be multiple ways of handling payments (not to mention a variety of payment amounts).

pdpinch:
> For that, I suggest if we should keep out the batch generation feature for certificates here.

To be clear, I will need to create a few hundred certificates in January for the two bootcamp runs that just ended. If I don't do that with this management command, I'll still need a management command.

klasses/management/commands/manage_certificates.py
HamzaIbnFarooq: I have updated it to user's email address.

asadiqbal08:
Ok then make sense to have it in it.

HamzaIbnFarooq:
@asadiqbal08 @pdpinch can you please take a final look into it and review, so I can proceed with merging?

klasses/management/commands/manage_certificates.py
```diff
@@ -0,0 +1,130 @@
...
+                ),
+            )
+        elif generate_batch and bootcamp_run:
+            updated = generate_batch_certificates(bootcamp_run)
+            self.show_message(
+                updated,
+                message="{} new certificates have been created for bootcamp-run:{}".format(
+                    len(updated), bootcamp_run
+                ),
+            )
```
pdpinch, line 122: Can we get to this point and none of the conditions have been met? Should we have a generic error?

klasses/utils.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+
+def revoke_certificate(user, bootcamp_run):
+    """Revokes certificate for the given user and bootcamp-run"""
+    if user and bootcamp_run:
+        certificate = BootcampRunCertificate.all_objects.filter(
+            user=user, bootcamp_run=bootcamp_run
+        ).first()
+        return certificate.revoke() if certificate else None
```
pdpinch, line 44: What happens if no certificate is found?

klasses/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def unrevoke_certificate(user, bootcamp_run):
+    """Unrevokes certificate for the given user and bootcamp-run"""
+    if user and bootcamp_run:
+        certificate = BootcampRunCertificate.all_objects.filter(
+            user=user, bootcamp_run=bootcamp_run
+        ).first()
```
pdpinch, line 52: What happens if no certificate is found?

pdpinch (commented):
I left a few questions about edge cases. If you think they are handled OK, then this is a 👍 for merging from me.

klasses/utils.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+        elif BootcampRunEnrollment.objects.filter(
+            user=user, bootcamp_run=bootcamp_run, active=True
+        ).exists():
+            certificate = BootcampRunCertificate.objects.create(
+                user=user, bootcamp_run=bootcamp_run
+            )
+            result.update(
+                {
+                    "updated": True,
```
asadiqbal08, line 30: will it not be added instead updated ?

klasses/utils.py
```diff
@@ -0,0 +1,152 @@
...
+            ] = "A {} certificate:{} already exists for user:{} in bootcamp-run:{}".format(
+                "revoked" if certificate.is_revoked else "",
+                certificate.link,
+                user.email,
+                bootcamp_run,
+            )
+        elif BootcampRunEnrollment.objects.filter(
+            user=user, bootcamp_run=bootcamp_run, active=True
+        ).exists():
+            certificate = BootcampRunCertificate.objects.create(
```
asadiqbal08, line 25: will it not create multiple entries for same `user` and `bootcamp_run`. Not better to use `get_or_create/update_or_create ` there

asadiqbal08:
Overall it lgtm 👍 at that point. just 2 things.
- we should add unit test for the command as well. 
- A comment over BootcampRunCertificate creation flow.

klasses/utils.py
HamzaIbnFarooq: I needed a `generic name`, as it has been used in case of `revoke, unrevoke and generate`, so chosen `updated` meaning if something gets updated by this command or not.

klasses/utils.py
HamzaIbnFarooq: On line 10, we have already checked if a certificate already exists or not for this user and bootcamp-run.

klasses/utils.py
HamzaIbnFarooq: I have added a lot of logging now, hope it will help. You may have a look.

asadiqbal08 (approved):
:shipit:

Issue #1102, opened 2020-12-15
Certificates: add link to certificate from application dashboard


pdpinch:
As a user, I need to have a link from my dashboard to my certificate

### Design 

https://invis.io/PGXCR7EMTKD#/417192809_Bootcamp_Dashboard_Dashboard_Collapsed

### Acceptance criteria

- [ ] Only display the link if there is both a certificate record for the user AND a certificate template
- [ ] This feature should be behind a feature flag so we can control when it is deployed

HamzaIbnFarooq:
@pdpinch can you please provide the material-icon shown in the invisio link:
![image](https://user-images.githubusercontent.com/42243411/102994689-579c8b80-4541-11eb-811e-7f50c1a9fd48.png)

pdpinch:
Took me a while to find it, but I believe this is it: https://material.io/resources/icons/?icon=history_edu&style=baseline

Issue #1103, opened 2020-12-15
Certificates: user view


pdpinch:
As a bootcamper, I need to be able to see my certificate, and share it with employers and social networks. 

Part of #1095 Certificates epic

### acceptance criteria

- [x] **IMPORTANT** the user view should be behind a feature flag
- [x]  unique URL based on UUID, same as xPRO
- [x] when the bootcamper is logged in, and viewing their own certificate, include sharing UI (same as xPRO)
- [x] print CSS omits header and footer and **prints on one page horizontally**
- [x] return a 404 when the certificate has been revoked

HamzaIbnFarooq:
Closed By https://github.com/mitodl/bootcamp-ecommerce/pull/1116

PR #1114, opened 2020-12-23
add certificate link to dashboard

HamzaIbnFarooq:
Fixes https://github.com/mitodl/bootcamp-ecommerce/issues/1102

#### What's this PR do?
Shows certificate link to user dashboard if the user has an active certificate and a certificate template as well.

Set environment variable in your .env file `FEATURE_ENABLE_CERTIFICATE_USER_VIEW=True`
Create a certificate for a user and then go to dashboard `/applications` to view it.

image

Without "View Statement"
image


app.json
```diff
@@ -211,6 +211,10 @@
       "description": "The OAuth client secret configured in the edX instance",
       "required": true
     },
+    "ENABLE_CERTIFICATE_DASHBAORD_LINKS": {
```
pdpinch, line 214: Since this a feature flag that will be short-lived, can you name it FEATURE_ENABLE_CERTIFICATE_DASHBAORD_LINKS ?

Also, does it make sense to use one feature flag for both the dashboard links and the user view #1103 ?

HamzaIbnFarooq: Yeah sure, I will refactor that variable name.

As far as the usage of `FEATURE_ENABLE_CERTIFICATE_DASHBAORD_LINKS` is concerned, I don't exactly know the difference between dashboard-links and user view, are not both the same? 
According to the current implementation, if this flag is turned off, user will not be able to view his certificate link/button on the dashboard/application page.

main/settings.py
```diff
@@ -219,6 +219,13 @@
     required=True,
 )
 
+FEATURE_ENABLE_CERTIFICATES_FOR_USERS = get_bool(
```
umarmughal824, line 222: Could we use `FEATURE_ENABLE_CERTIFICATES` as these certificates are only generated for the enrolled users so, in my opinion, no need to mention users?

applications/serializers.py
```diff
@@ -155,6 +156,21 @@ class BootcampApplicationSerializer(serializers.ModelSerializer):
     bootcamp_run = BootcampRunSerializer()
     has_payments = serializers.SerializerMethodField()
     enrollment = serializers.SerializerMethodField()
+    certificate_link = serializers.SerializerMethodField()
+
+    def get_certificate_link(self, application):
+        """Returns certificate link if both certificate and certificate template are presentand enabled"""
+        if getattr(settings, "FEATURE_ENABLE_CERTIFICATES_FOR_USERS", False):
```
umarmughal824, line 163: If I try to test it by setting `True` as a default value here rather than `False` without adding the environment variable that doesn't work either as you have set the default value `True`.

umarmughal824 (changes requested):
I am unable to test because of that error by setting the environment variable in `.env`. I don't how is it running for you?

```
web_1     | Traceback (most recent call last):
web_1     |   File "manage.py", line 14, in <module>
web_1     |     execute_from_command_line(sys.argv)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
web_1     |     utility.execute()
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 325, in execute
web_1     |     settings.INSTALLED_APPS
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 79, in __getattr__
web_1     |     self._setup(name)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 66, in _setup
web_1     |     self._wrapped = Settings(settings_module)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 157, in __init__
web_1     |     mod = importlib.import_module(self.SETTINGS_MODULE)
web_1     |   File "/usr/local/lib/python3.6/importlib/__init__.py", line 126, in import_module
web_1     |     return _bootstrap._gcd_import(name[level:], package, level)
web_1     |   File "<frozen importlib._bootstrap>", line 994, in _gcd_import
web_1     |   File "<frozen importlib._bootstrap>", line 971, in _find_and_load
web_1     |   File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
web_1     |   File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
web_1     |   File "<frozen importlib._bootstrap_external>", line 678, in exec_module
web_1     |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
web_1     |   File "/src/main/settings.py", line 803, in <module>
web_1     |     for key in get_all_config_keys()
web_1     |   File "/src/main/settings.py", line 804, in <dictcomp>
web_1     |     if key.startswith(BOOTCAMP_FEATURES_PREFIX)
web_1     |   File "/src/main/envs.py", line 74, in wrapper
web_1     |     raise ValueError(f"Environment variable '{name}' was used more than once")
web_1     | ValueError: Environment variable 'FEATURE_ENABLE_CERTIFICATES_FOR_USERS' was used more than once
```

umarmughal824 (commented):
It should be something like this.

Remove that line FEATURE line from the settings file as it's already getting imported and set the environment variable that be will be picked up and in your serialize use that statement 
`if settings.FEATURES.get("ENABLE_CERTIFICATES_FOR_USERS", False):`, then you will get rid-off that error.

applications/serializers.py
```diff
@@ -155,6 +156,21 @@ class BootcampApplicationSerializer(serializers.ModelSerializer):
     bootcamp_run = BootcampRunSerializer()
     has_payments = serializers.SerializerMethodField()
     enrollment = serializers.SerializerMethodField()
+    certificate_link = serializers.SerializerMethodField()
+
+    def get_certificate_link(self, application):
+        """Returns certificate link if both certificate and certificate template are presentand enabled"""
```
umarmughal824, line 162:
```suggestion
        """Returns certificate link if both certificate and certificate template are present and enabled"""
```

applications/serializers.py
umarmughal824:
```suggestion
        if settings.FEATURES.get("ENABLE_CERTIFICATES_FOR_USERS", False):
```

applications/serializers.py
```diff
@@ -155,6 +156,21 @@ class BootcampApplicationSerializer(serializers.ModelSerializer):
...
+        if getattr(settings, "FEATURE_ENABLE_CERTIFICATES_FOR_USERS", False):
+            user = application.user
+            certificate = application.bootcamp_run.certificates.filter(
+                user=user
+            ).first()
```
umarmughal824, line 167: Normally first is used to obtain if they could be more than one object but in our case, one user can have one certificate.
It should be something like this
```
from django.core.exceptions import ObjectDoesNotExist
try:
     // return the link
except ObjectDoesNotExist:
     return None
```

main/settings.py
HamzaIbnFarooq: This name sounds good, but my intention of using `FOR_USERS` was that the whole certificate feature doesn't get blocked by this. Admin can still create certificates and certificate pages. Only the user's view will be affected. Does this change your mind?

HamzaIbnFarooq:
> It should be something like this.
> 
> Remove that line FEATURE line from the settings file as it's already getting imported and set the environment variable that be will be picked up and in your serialize use that statement
> `if settings.FEATURES.get("ENABLE_CERTIFICATES_FOR_USERS", False):`, then you will get rid-off that error.

I am still thinking, how the tests are getting passed 😬

umarmughal824 (changes requested):

static/scss/application.scss
```diff
@@ -183,3 +183,43 @@ $border-style: 1px solid $black;
...
+    margin-left: 100px;
+  }
+
+  .view-statement {
+    color: #4e9eff;
+    font-weight: 300;
+  }
+
+  .btn-view-certificate {
+    background-color: #007ee5;
```
umarmughal824, line 210: don't use the hex value. we need to create a variable against that hex in `_variable.scss` and use that variable here.

main/settings.py
umarmughal824: In that case, I would go with the @pdpinch suggestion `FEATURE_ENABLE_CERTIFICATE_DASHBAORD_LINKS `.

HamzaIbnFarooq: What about `ENABLE_CERTIFICATE_USER_VIEW`?

applications/serializers.py
HamzaIbnFarooq: Don't you think `try-except` is an overhead in this case as we know that the filter will not return more than 1 record ever?

static/scss/application.scss
HamzaIbnFarooq: Updated.

applications/serializers.py
HamzaIbnFarooq: Updated to `FEATURE_ENABLE_CERTIFICATE_USER_VIEW` for now.

applications/serializers.py
HamzaIbnFarooq: updated.

applications/serializers.py
umarmughal824: no, it's not an overhead.

HamzaIbnFarooq: Updated.

umarmughal824 (approved):
👍 , Just mention the flag name in the PR Description so that we could remember the exact flag.

umarmughal824:
@HamzaIbnFarooq please add this in your PR Description
Set environment variable in your `.env` file `FEATURE_ENABLE_CERTIFICATE_USER_VIEW=True`

Issue #1115, opened 2020-12-24
Certificates UnitTest Integration


asadiqbal08:
As we merged our certificate changes to central repository to complete the feature functionality so it is better to consider to implement some unit tests for certificate modules. 

#### Unit tests for: 

- [x] Wagtail certificate changes under `cms/models.py`
- [x]  Management command 'klasses/management/commands/manage_certificates.py'
- [x]  /klasses/models.py

PR #1117, opened 2020-12-28
certificates unittest for cms and klass models

umarmughal824:
#1115 

#### What's this PR do?
Certificates unittest for cms and klass models

### DIFF
--- a/cms/factories.py
+++ b/cms/factories.py
@@ -289,3 +289,39 @@ class AdmissionSectionFactory(wagtail_factories.PageFactory):
 
     class Meta:
         model = models.AdmissionsSection
+
+
+class SignatoryPageFactory(wagtail_factories.PageFactory):
+    """SignatoryPage factory class"""
+
+    name = factory.fuzzy.FuzzyText(prefix="Name")
+    title_1 = factory.fuzzy.FuzzyText(prefix="Title_1")
+    title_2 = factory.fuzzy.FuzzyText(prefix="Title_2")
+    organization = factory.fuzzy.FuzzyText(prefix="Organization")
+    signature_image = factory.SubFactory(wagtail_factories.ImageFactory)
+
+    class Meta:
+        model = models.SignatoryPage
+
+
+class CertificateIndexPageFactory(wagtail_factories.PageFactory):
+    """CertificateIndexPage factory class"""
+
+    slug = factory.fuzzy.FuzzyText()
+
+    class Meta:
+        model = models.CertificateIndexPage
+
+
+class CertificatePageFactory(wagtail_factories.PageFactory):
+    """CertificatePage factory class"""
+
+    bootcamp_run_name = factory.fuzzy.FuzzyText(prefix="bootcamp_run_")
+    certificate_name = factory.fuzzy.FuzzyText()
+    location = factory.fuzzy.FuzzyText()
+    signatories = wagtail_factories.StreamFieldFactory(
+        {"signatory": SignatoryPageFactory}
+    )
+
+    class Meta:
+        model = models.CertificatePage
--- a/cms/models_test.py
+++ b/cms/models_test.py
@@ -3,6 +3,8 @@
 import json
 import pytest
 
+from django.http.response import Http404
+
 from cms.factories import (
     SiteNotificationFactory,
     BootcampRunPageFactory,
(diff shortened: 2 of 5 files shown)


klasses/models_test.py
```diff
@@ -171,3 +193,29 @@ def test_total_due_by_next_deadline():
...
+    certificate = BootcampRunCertificateFactory.create(
+        user=user, bootcamp_run=bootcamp_run
+    )
+
+    assert certificate.is_revoked is False
+    assert certificate.start_end_dates == (
+        bootcamp_run.start_date,
+        bootcamp_run.end_date,
+    )
+    assert certificate.link == "/certificate/{}/".format(str(certificate.uuid))
```
HamzaIbnFarooq, line 212: nit: we should stick to one pattern for `format()` either use `{}` or `{uuid}`. secondly, we should also stay consistent with double `"` and single quotes `'`.

klasses/models_test.py
```diff
@@ -171,3 +193,29 @@ def test_total_due_by_next_deadline():
...
+    assert str(
+        certificate
+    ) == 'BootcampRunCertificate for user={user}, run={bootcamp_run} ({uuid})"'.format(
```
HamzaIbnFarooq, line 215:
```suggestion
    ) == 'BootcampRunCertificate for user={user}, run={bootcamp_run} ({uuid})'.format(
```
Also remove that double-quotes from `BootcampRunCertificate.__str__` method

cms/models_test.py
```diff
@@ -265,3 +270,64 @@ def test_home_catalog():
...
+    certificate = BootcampRunCertificateFactory.create(
+        bootcamp_run=bootcamp_run_page.bootcamp_run
+    )
+    certificate_page = CertificatePageFactory.create(parent=bootcamp_run_page)
+    request = rf.get(certificate_page.get_url())
+    assert (
+        certifcate_index_page.bootcamp_certificate(
+            request, certificate.uuid
+        ).status_code
+        == 200
```
HamzaIbnFarooq, line 326: we can also use `status.HTTP_200_OK`

cms/factories.py
```diff
@@ -289,3 +289,39 @@ class AdmissionSectionFactory(wagtail_factories.PageFactory):
...
+    slug = factory.fuzzy.FuzzyText()
+
+    class Meta:
+        model = models.CertificateIndexPage
+
+
+class CertificatePageFactory(wagtail_factories.PageFactory):
+    """CertificatePage factory class"""
+
+    bootcamp_run_name = factory.fuzzy.FuzzyText(prefix="bootcamp_run")
```
HamzaIbnFarooq, line 319:
```suggestion
    bootcamp_run_name = factory.fuzzy.FuzzyText(prefix="bootcamp_run_")
```

cms/models_test.py
```diff
@@ -265,3 +270,64 @@ def test_home_catalog():
...
+    certificate_page = CertificatePageFactory.create(
+        parent=bootcamp_run_page,
+        bootcamp_run_name="bootcamp_run",
+        certificate_name="certificate_name",
+        location="location",
+        signatories__0__signatory__name="Name",
+        signatories__0__signatory__title_1="Title_1",
+        signatories__0__signatory__title_2="Title_2",
+        signatories__0__signatory__organization="Organization",
+        signatories__0__signatory__signature_image__title="Image",
```
HamzaIbnFarooq, line 292: shouldn't we use SignatoryPageFactory here?

HamzaIbnFarooq (changes requested):
commented some minor changes

cms/models_test.py
umarmughal824: See how it's being done in xPRO, you will get the idea of it.

cms/models_test.py
umarmughal824: We can use it but it's not necessary. As that status is a part of `rest_framework` but we are not using it anywhere in the `cms` application.

klasses/models_test.py
umarmughal824: that `"` is also placed in the view so I kept it the same there. I did not notice that.
https://github.com/mitodl/bootcamp-ecommerce/blob/3d1df795f9a1360748e6f31b34e94c1723143b65/klasses/models.py#L343

klasses/models_test.py
umarmughal824: that `"` is also placed in the view so I kept it the same there. I did not notice that. By the way, I have removed it in both 
https://github.com/mitodl/bootcamp-ecommerce/blob/3d1df795f9a1360748e6f31b34e94c1723143b65/klasses/models.py#L343

HamzaIbnFarooq (approved):
LGTM

Issue #1127, opened 2021-01-06
Certificate Preview Error


HamzaIbnFarooq:
When a staff memeber tries to preview certificate, it gives following error:
```
web_1     | Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 829, in _resolve_lookup
web_1     |     current = current[bit]
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/template/context.py", line 83, in __getitem__
web_1     |     raise KeyError(key)
web_1     | KeyError: 'certificate_user'
web_1     |
web_1     | During handling of the above exception, another exception occurred:
web_1     |
web_1     | Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 835, in _resolve_lookup
web_1     |     if isinstance(current, BaseContext) and getattr(type(current), bit):
web_1     | AttributeError: type object 'RequestContext' has no attribute 'certificate_user'
web_1     |
web_1     | During handling of the above exception, another exception occurred:
web_1     |
web_1     | Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 843, in _resolve_lookup
web_1     |     current = current[int(bit)]
web_1     | ValueError: invalid literal for int() with base 10: 'certificate_user'
web_1     |
web_1     | During handling of the above exception, another exception occurred:
web_1     |
web_1     | Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 850, in _resolve_lookup
```

PR #1128, opened 2021-01-06
small fixes for certificates template rendering

HamzaIbnFarooq:
Fixes https://github.com/mitodl/bootcamp-ecommerce/issues/1127

#### What's this PR do?
This PR solves three small bugs regarding Certificate Template rendering:
1) `certificate_user was not present in the certificate's preview` context which caused rendering errors, this PR fixes it.
2) certificate page doesn't have any field named `product_name`, this PR replaces it with `bootcamp_run_name`
3) it adds `share_text` for the LinkedIn certificate sharing feature in the certificate context. 

Just open certificates in different modes and see if any error occurs or not

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -1035,6 +1035,7 @@ class CertificatePage(BootcampRunChildPage):
                 if parent.bootcamp_run
                 else datetime.now(),
                 "location": self.location if self.location else "Brisbane, Australia",
+                "certificate_user": None,
             }
         elif self.certificate:
             # Verify that the certificate in fact is for this same course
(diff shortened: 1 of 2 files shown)


pdpinch:
Hamza, please be more specific about what this PR does. The commit message and the PR need more detail.

HamzaIbnFarooq:
> Hamza, please be more specific about what this PR does. The commit message and the PR need more detail.

Updated.

cms/models.py
```diff
@@ -1059,6 +1060,7 @@ def get_context(self, request, *args, **kwargs):
                 request.build_absolute_uri("///"),
                 static("images/certificates/share-image.png"),
             ),
+            "share_text": "I just earned a certificate",
```
arslanashraf7, line 1063:
```suggestion
            "share_text": "I just earned a certificate in {}".format(
                self.bootcamp_run_name
            ),
```

I think we should include `Bootcamp Run Name`, Contextually, similar to what we are doing in other application https://github.com/mitodl/mitxpro/blob/b4eed13024fc9438a25b05d47a2292a20e55dbf7/cms/models.py#L1597.

arslanashraf7 (approved):
Just a small suggestion, The rest looks good 👍

cms/models.py
HamzaIbnFarooq: updated.

PR #1133, opened 2021-01-08
Fixed application state for personal prices set to zero

### LINKED ISSUE #1112: Fix personal price behavior when price is set to 0
Related to #1094 

Steps to reproduce:
1. Get a bootcamp application into `AWAITING_PAYMENT` state (you can use the `set_application_state` mgmt command)
2. Create a personal price record and set the amount to 0, indicating that the user does not need to pay anything to complete the application

Expected behavior: The application should be automatically set to `COMPLETE`

Actual behavior: The application remains at `AWAITING_PAYMENT`

_Note:_ This does work if the user has made any payment at all (i.e.: the user has a least one `Order` and `Line` record for a payment towards the total amount owed for the given bootcamp run). This should be an easy fix

gsidebo:
Fixes #1112 

#### What's this PR do?
- Fixes personal price adjustment behavior such that if a personal price is set to 0, and a user has an application for that bootcamp run in the "awaiting payment" state, that application is then set to "complete"
- Streamlines some of the `klasses` app admin class definitions

1. Start out with a user with an application that is awaiting payment (use the `set_application_state` command)
1. Ensure that the given bootcamp run has an installment with a non-zero amount
1. Create/update a personal price for that user/run with the amount set to zero

This is signals-based, so your application should be automatically set to "complete"

Initial implementation for this personal price adjustment behavior was in this PR: https://github.com/mitodl/bootcamp-ecommerce/pull/1094

### DIFF
--- a/klasses/admin.py
+++ b/klasses/admin.py
@@ -45,7 +45,7 @@ class BootcampRunEnrollmentAdmin(TimestampedModelAdmin):
 
     model = models.BootcampRunEnrollment
     include_created_on_in_list = True
-    list_display = ("id", "bootcamp_run", "user", "change_status", "active")
+    list_display = ("id", "get_user_email", "bootcamp_run", "change_status", "active")
     list_filter = ("change_status", "active", "bootcamp_run__bootcamp")
     raw_id_fields = ("bootcamp_run", "user")
     search_fields = (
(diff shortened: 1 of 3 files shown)


klasses/admin.py
```diff
@@ -104,17 +126,24 @@ class BootcampRunCertificateAdmin(TimestampedModelAdmin):
     ]
     raw_id_fields = ("user",)
 
+    def get_queryset(self, request):
+        return self.model.all_objects.get_queryset().select_related(
```
arslanashraf7, line 130: Should we consider adding a doc-string?

arslanashraf7 (approved):
Looks good 👍 , Just a small question added.

klasses/admin.py
gsidebo: It's an overridden method and I think it's clear enough what its used for

PR #1134, opened 2021-01-08
Ignore submit, do not throw error, if partial token is null on password page

### LINKED ISSUE #1130: Error: Invalid state: password page with no partialToken
Sentry Issue: [BOOTCAMPS-X6](https://sentry.io/organizations/mit-office-of-digital-learning/issues/2083616856/?referrer=github_integration)

```
Error: Invalid state: password page with no partialToken
  at wrap (./static/js/pages/login/LoginPasswordPage.js:68:13)
  at rf (None:44:11)
  at a (None:40:1)
  at None (None:40:1)
  at Promise ([native code])
...
(7 additional frame(s) were not displayed)
```

mbertrand:
Closes #1130 

#### What's this PR do?
Instead of throwing an error,  the submit button does nothing if pressed while the auth partial_token is null.  After talking with Nathan, best guess is that this scenario should only happen on a double submission.  If the token is null on page load, the user would be redirected back to the signin page.

Tests should pass.

### DIFF
--- a/static/js/pages/login/LoginPasswordPage.js
+++ b/static/js/pages/login/LoginPasswordPage.js
@@ -64,17 +64,15 @@ export class LoginPasswordPage extends React.Component<Props> {
       auth: { partial_token }
     } = this.props
 
-    if (!partial_token) {
-      throw Error("Invalid state: password page with no partialToken")
-    }
-
     try {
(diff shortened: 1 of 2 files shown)


gsidebo (approved):
Looks great! 👍

Issue #1135, opened 2021-01-11
Certificates: management command to block certificate creation for specified bootcampers


pdpinch:
As a bootcamp admin, I often have a list of enrolled bootcampers who shouldn't get a certificate (typically because they dropped out before the bootcamp finished). 

I'd like to be able to block from these bootcampers from getting a certificate before I run ```python manage.py manage_certificates --generate-batch```

asadiqbal08:
@pdpinch are you thinking to pass the list of enrolled bootcampers who shouldn't get a certificate along with the command ? or are you maintain the list elsewhere in database model ?

pdpinch:
I'd like to run management commands something like:

```
python manage.py manage_certificates --block camper1@example.com camper2@example.com camper3@example.com
python manage.py manage_certificates --generate-batch
```

It's up to you where to store the data -- please make a suggestion.

asadiqbal08:
@pdpinch I guess, you will run the command in that format then. 
```
python manage.py manage_certificates --generate-batch --block camper1@example.com camper2@example.com camper3@example.com
```

asadiqbal08:
What about, If we define a boolean field in `BootcampRunEnrollment` as `user_blocked`. 
```
python manage.py manage_certificates --block camper1@example.com camper2@example.com camper3@example.com
``` 
the following format of command will just **update** the `BootcampRunEnrollment` model for the provided enrolled user list (if present in model). 

and then the command 
``` python manage.py manage_certificates --generate-batch``` will ignore the blocked user while filtering `BootcampRunEnrollment` model [here](https://github.com/mitodl/bootcamp-ecommerce/blob/master/klasses/utils.py#L51-L60). 

thought ?

pdpinch:
That makes sense to me, but the field name should make it clear that the bootcamper is blocked from getting a certificate — perhaps `certificate_is_blocked` ? 

----------------
I am mobile.

PR #1142, opened 2021-01-20
Block/UnBlock list of users for getting certificates. 

asadiqbal08:
fixes: #1135 

#### What's this PR do?
As a bootcamp admin, I often have a list of enrolled bootcampers who shouldn't get a certificate (typically because they dropped out before the bootcamp finished).

Execute the command in following format. 
```
python manage.py manage_certificates --run={RUN} --block camper1@example.com camper2@example.com camper3@example.com
python manage.py manage_certificates --generate-batch --run={RUN}
python manage.py manage_certificates --generate --run={RUN} --user={USER}
```
Certificate should not be generated for the blocked users in ~any~ provided enrollment run. 
But after, 
```
python manage.py manage_certificates --run={RUN} --unblock camper1@example.com camper2@example.com camper3@example.com
python manage.py manage_certificates --generate-batch --run={RUN}
python manage.py manage_certificates --generate --run={RUN} --user={USER}
```
Certificate should not be generated for the users. 

~#### NOTE: Currently executing the command with option `--block [USER LIST]` will block the users for getting certificate in ~any~ specific enrollment run. We are not targeting to block user for a specific enrollment run in the command.~


klasses/admin.py
```diff
@@ -45,8 +45,20 @@ class BootcampRunEnrollmentAdmin(TimestampedModelAdmin):
...
     include_created_on_in_list = True
-    list_display = ("id", "get_user_email", "bootcamp_run", "change_status", "active")
-    list_filter = ("change_status", "active", "bootcamp_run__bootcamp")
+    list_display = (
+        "id",
+        "get_user_email",
+        "bootcamp_run",
+        "change_status",
+        "active",
+        "user_certificate_is_blocked",
```
HamzaIbnFarooq, line 54: kindly have a look at this comment: https://github.com/mitodl/bootcamp-ecommerce/issues/1135#issuecomment-762510159

HamzaIbnFarooq (commented):
I have three questions regarding this PR (although they don't seem to be in the current scope):
1) What happens if a blocked user enrolls in a new course? according to the current implementation, certificates will start getting created for new enrollments.
2) If a certificate has already been created and then the user gets blocked, should we show that certificate to user?
3) Would there be any scenario where admin would want to block user certificates only for a specific bootcamp-run?

klasses/admin.py
asadiqbal08: Yes, I changed to add a mark `user` here for better understanding.

klasses/utils.py
```diff
@@ -150,3 +155,26 @@ def unrevoke_certificate(user, bootcamp_run):
         result["msg"] = "Valid user and bootcamp_run must be provided."
 
     return result
+
+
+def manage_user_certificate_blocking(users, block_state):
+    """Block users for getting certificates in all bootcamp run enrollments"""
+    result = {"updated": False}
+    if users:
+        BootcampRunEnrollment.objects.filter(user__email__in=users, active=True).update(
```
HamzaIbnFarooq, line 164:
```suggestion
        BootcampRunEnrollment.objects.filter(user__email__in=users).update(
```
might not create much impact but I guess we shouldn't leave any loophole.

pdpinch:
Good questions @HamzaIbnFarooq 

> 1. What happens if a blocked user enrolls in a new course? according to the current implementation, certificates will start getting created for new enrollments.

The block is specific to an enrollment, not a user. So creating certificates for new enrollments is the correct behavior. 

> 2. If a certificate has already been created and then the user gets blocked, should we show that certificate to user?

We have a separate management command for revoking certificates that's designed for hiding an already created certificate. I think it makes more sense for an admin to use the revoke command, instead of adding more code logic to hide a blocked certificate.

pdpinch:
@asadiqbal08 I would prefer that the admin user specify the run that the block applies to, instead of having it apply to all runs. Is that difficult?

> NOTE: Currently executing the command with option --block [USER LIST] will block the users for getting certificate in any enrollment. We are not targeting to block user for a specific enrollment run in the command.

asadiqbal08:
> What happens if a blocked user enrolls in a new course? according to the current implementation, certificates will start getting created for new enrollments.


Good catch @HamzaIbnFarooq then I think adding the option of `--run` along with `--block` make sense.

klasses/utils.py
```diff
@@ -150,3 +155,26 @@ def unrevoke_certificate(user, bootcamp_run):
...
+        state = "blocked"
+        if not block_state:
+            state = "unblocked"
+
+        result.update(
+            {
+                "updated": True,
+                "msg": "Users:{} are ".format(users)
+                + state
+                + " for getting certificate for any bootcamp enrollment.",
```
HamzaIbnFarooq, line 177: we should use bootcamp title here.

klasses/utils.py
```diff
@@ -150,3 +155,26 @@ def unrevoke_certificate(user, bootcamp_run):
         result["msg"] = "Valid user and bootcamp_run must be provided."
 
     return result
+
+
+def manage_user_certificate_blocking(users, block_state, bootcamp_run):
+    """Block users for getting certificates in all bootcamp run enrollments"""
+    result = {"updated": False}
+    if users:
+        BootcampRunEnrollment.objects.filter(
```
HamzaIbnFarooq, line 164: We can improve logging for the case where no matching enrollment is found.

HamzaIbnFarooq (commented):
LG, just need some extra logging (even in certificate generation for blocked users) and test cases.

klasses/utils.py
```diff
@@ -150,3 +155,37 @@ def unrevoke_certificate(user, bootcamp_run):
...
+
+        state = "blocked"
+        if not block_state:
+            state = "unblocked"
+
+        blocked_users_emails = [run.user.email for run in run_enrollments]
+        result.update(
+            {
+                "updated": True,
+                "msg": "User(s) with email in {} ".format(blocked_users_emails)
```
HamzaIbnFarooq, line 186:
```suggestion
                "msg": "User(s) with email in {} are ".format(blocked_users_emails)
```

HamzaIbnFarooq (approved):
LG 👍

klasses/utils.py
asadiqbal08: It can be a single user in list or multiple so I thought it is better to drop `is/are` from the message.

PR #1148, opened 2021-01-26
Wiretransfer: after data is imported, update the application state

### LINKED ISSUE #1032: Wiretransfer: after data is imported, update the application state
### Steps to reproduce

1. Create a user and an application in the state `AWAITING_PAYMENT`
2. Run the management comment `import_wiretransfers` with a payment sufficient to complete the payment for the bootcamp

### Expected Behavior

1. Since the payment is complete, the application state should advance to `COMPLETE`
2. Since the state is `COMPLETE` an enrollment object should be created

### Actual Behavior

- application state remains in `AWAITING PAYMENT`
- no enrollment is created

umarmughal824:
fixes #1032 

#### What's this PR do?
Wiretransfer: after data is imported, update the application state

Run the `import_wire_transfers` management command for `AWAITING_PAYMENT` application

Screenshot 2021-01-26 at 17 11 32

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -559,7 +559,7 @@ def import_wire_transfer(wire_transfer, header_row):
 
     with transaction.atomic():
         order = Order.objects.create(
-            status=Order.FULFILLED,
+            status=Order.CREATED,
             total_price_paid=wire_transfer.amount,
             application=application,
             user=user,
@@ -571,7 +571,6 @@ def import_wire_transfer(wire_transfer, header_row):
             description=f"Wire transfer payment for {bootcamp_run}",
             price=wire_transfer.amount,
         )
-        order.save_and_log(None)  # save record in audit table
 
         WireTransferReceipt.objects.create(
             wire_transfer_id=wire_transfer.id,
@@ -580,6 +579,8 @@ def import_wire_transfer(wire_transfer, header_row):
             },
             order=order,
         )
+        complete_successful_order(order, send_receipt=False)
+
     log.info("Wire transfer %d successfully imported", wire_transfer.id)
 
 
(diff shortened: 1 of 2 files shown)


ecommerce/api.py
```diff
@@ -580,6 +581,22 @@ def import_wire_transfer(wire_transfer, header_row):
...
             order=order,
         )
+
+        if application.is_paid_in_full:
+            application.state = AppStates.COMPLETE.value
+            application.save()
+
+            _, created = BootcampRunEnrollment.objects.get_or_create(
+                user=user, bootcamp_run=bootcamp_run
+            )
```
mbertrand, line 591: Might be better to replace the above with 

```python
complete_successful_order(order, send_receipt=????)
``` 

Not sure if we'd want to send a receipt to the user or not, Peter probably knows.
The above function would also run `novoed_tasks.enroll_users_in_novoed_course.delay` if appropriate.

mbertrand (changes requested):

pdpinch:
> Not sure if we'd want to send a receipt to the user or not, Peter probably knows.
> 
Yes, we want to send a receipt when the wiretransfer is imported — same as if we had completed a Cybersource transaction. 
> The above function would also run novoed_tasks.enroll_users_in_novoed_course.delay if appropriate.
> 
We also want bootcampers to be enrolled in novoed, if appropriate (i.e. if they have completed paying their personal price, and there is a novoed stub set up).

mbertrand (approved):
👍

ecommerce/api.py
```diff
@@ -580,6 +579,8 @@ def import_wire_transfer(wire_transfer, header_row):
             },
             order=order,
         )
+        complete_successful_order(order, send_receipt=False)
```
mbertrand, line 582: Based on Peter's comment, can you drop the `send_receipt=False`?

mbertrand (changes requested):

mbertrand (approved):
👍

PR #1156, opened 2021-02-09
#1155 display user's country in export compliance admin

### LINKED ISSUE #1155: display user's country in export compliance admin
As the customer support manager, it would be helpful to be able to look up the country users' have submitted when they fail export compliance. 

This information is currently shared by creating a ZenDesk ticket, but that is less convenient than using the compliance admin where I already have the interface to look up and override the export compliance status. 

This could be as simple as adding profile.country to the admin list view.

umarmughal824:
fixes #1155 

#### What's this PR do?
display user's country in export compliance admin

Just see the country in the list display of the admin

Screenshot 2021-02-09 at 17 35 38

### DIFF
--- a/compliance/admin.py
+++ b/compliance/admin.py
@@ -2,6 +2,7 @@
 Admin site bindings for compliance
 """
 
+import pycountry
 from django.conf import settings
 from django.contrib import admin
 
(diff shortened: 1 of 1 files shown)


compliance/admin.py
```diff
@@ -17,11 +18,21 @@ class ExportsInquiryLogAdmin(admin.ModelAdmin):
...
+        "user__legal_address__country",
+    ]
+    list_display = ["user", "computed_result", "info_code", "reason_code", "country"]
+    list_select_related = ["user__legal_address"]
     readonly_fields = [] if settings.DEBUG else get_field_names(ExportsInquiryLog)
     actions = ["manually_approve_inquiry"]
 
+    def country(self, instance):
+        """ generate country name from pycountry """
+        return pycountry.countries.get(alpha_2=instance.user.legal_address.country).name
```
HamzaIbnFarooq, line 34: If a country's abbreviation doesn't match any country in pycountry, it will give an error. I think we should do some exception handling here.

umarmughal824: @HamzaIbnFarooq that's not going to happen because it's coming from the user profile legal_address which is already validating that. 

Note: We are not putting anything manually this is what ever we are getting from the user profile data which is already validated.

https://github.com/mitodl/bootcamp-ecommerce/blob/2eea9ef8a62dd9c6b992598893072f2d179676d7/profiles/models.py#L108-L110

https://github.com/mitodl/bootcamp-ecommerce/blob/2eea9ef8a62dd9c6b992598893072f2d179676d7/profiles/models.py#L57-L71

HamzaIbnFarooq (approved):
👍

umarmughal824:
@pdpinch could you take a look at the attached screenshot?

pdpinch:
Looks great, thanks.

pdpinch (approved):

PR #1162, opened 2021-02-22
Send compliance emails to ADMIN_EMAIL instead of EMAIL_SUPPORT

### LINKED ISSUE #1161: Send compliance emails to ADMIN_EMAIL instead of EMAIL_SUPPORT
Sending export compliance failure emails to ZenDesk isn't useful, and creates more work managing tickets. Customer support only needs to know these details when a user contacts us to ask for help, and then always have to ask the user to verify their address anyway. 

The only useful information for customer support turns out to be the user's country, which we add to the export compliance admin in PR #1155.

Instead of sending the emails to ZenDesk, we can send them to the ADMIN_EMAIL account, so the engineering team can monitor any unusual bursts in activity, but otherwise ignore them. 

The email code is here: 
https://github.com/mitodl/bootcamp-ecommerce/blob/e9d7ef194c62614533eb1ae9cec1092807a8dd3c/authentication/pipeline/compliance.py#L102-L108

umarmughal824:
fixes #1161 

#### What's this PR do?
Send compliance emails to ADMIN_EMAIL instead of EMAIL_SUPPORT

Try to generate the Compliance email and check you are receiving it on ADMIN_EMAIL


mbertrand (approved):
Now receiving emails from the support email to the admin email, instead of vice versa.  👍

PR #1171, opened 2021-03-11
import bootcamp alumni

### LINKED ISSUE #1167: Import Bootcamp Alums
### Goal: 
Sometimes, we would like to treat bootcamp users who have already attended a bootcamp differently from other users of the site. We need a way to keep track of who has completed a bootcamp. (For example #1166 )

### Background

For bootcampers who have been admitted since the site redesign in July 2020, we can use the enrollment table as a sign of alumni status. If the user was enrolled in the bootcamp, we can assume that they are alumni (since they paid and enrolled). For any alumni who paid and enrolled but did not complete the bootcamp, they should have the flag `user_certificate_is_blocked` set to True.

### Possibile implementation

For bootcampers from prior runs, we have a spreadsheet from the bootcamp team consisting of email addresses and bootcamp(s). Most of these emails, and most of these historical bootcamps will already be in our database. We can simply import the data, and create an entry in the enrollments table for each of them. 

If there are any missing bootcamp runs, we can just create them. 

For missing user accounts, we can create accounts for users, but without password and incomplete profiles. When the user attempts to login, they will have to use the “forgot password” feature to set a password, and then will need to complete their profile.

umarmughal824:
fixes #1167

#### What's this PR do?
import Bootcamp alumni and create the enrollment without creating any application

Run the management command 
`docker-compose exec web python manage.py import_alumni profiles/testdata/example_alumni.csv`

(Optional)

Just check that for those users who were not existing in our system. They should try to forget password and login that should ask them to fill out their profile first.

### DIFF
--- a/profiles/api.py
+++ b/profiles/api.py
@@ -1,19 +1,49 @@
 """Users api"""
-import re
+from collections import namedtuple
+import csv
+from datetime import datetime, timedelta
 from functools import reduce
+import logging
+import re
 import operator
+import pytz
 
+from django.db import transaction
 from django.db.models import Q
 from django.core.exceptions import ValidationError
 from django.core.validators import validate_email
 from django.contrib.auth import get_user_model
+from django.utils.timezone import make_aware
 
+from klasses.models import BootcampRun, BootcampRunEnrollment
 from main.utils import first_or_none, unique, unique_ignore_case, max_or_none
-from profiles.constants import USERNAME_MAX_LEN
+from profiles.constants import (
+    ALUM_BOOTCAMP_END_DATE,
+    ALUM_BOOTCAMP_NAME,
+    ALUM_BOOTCAMP_RUN_TITLE,
+    ALUM_BOOTCAMP_START_DATE,
+    ALUM_HEADER_FIELDS,
+    ALUM_LEARNER_EMAIL,
+    USERNAME_MAX_LEN,
+)
+from profiles.exceptions import AlumImportException
+from profiles.utils import usernameify
 
 User = get_user_model()
 
 CASE_INSENSITIVE_SEARCHABLE_FIELDS = {"email"}
+
+Alum = namedtuple(
+    "Alum",
+    [
+        "learner_email",
+        "bootcamp_name",
+        "bootcamp_run_title",
+        "bootcamp_start_date",
+        "bootcamp_end_date",
+    ],
+)
+log = logging.getLogger(__name__)
 
 
 def get_user_by_id(user_id):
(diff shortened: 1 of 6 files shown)


umarmughal824:
@pdpinch could you review that template for it?
https://github.com/mitodl/bootcamp-ecommerce/blob/de4e82308a7c047a0a616c9c2e369ee9b7853882/profiles/testdata/example_alumni.csv

pdpinch:
> @pdpinch could you review that template for it?

Can the values for dates be just dates instead of timestamps?

profiles/api.py
```diff
@@ -256,3 +285,125 @@ def is_user_info_complete(user):
...
+    Read CSV file and convert to Alumns object
+
+    Args:
+        csv_path(str): Path to the CSV file
+
+    Returns:
+        (list of Alumns, list):
+    """
+    with open(csv_path) as csv_file:
+        rows = list(csv.reader(csv_file))
```
noisecapella, line 357: I think you can use `csv.DictReader` here. That will save you the logic of storing the header row and looking up indexes to columns there.

noisecapella (changes requested):

profiles/api.py
```diff
@@ -256,3 +285,119 @@ def is_user_info_complete(user):
...
+
+def parse_alumni_csv(csv_path):
+    """
+    Read CSV file and convert to Alumns object
+
+    Args:
+        csv_path(str): Path to the CSV file
+
+    Returns:
+        (list of Alumns, list):
```
noisecapella, line 357: This should be `list of Alums:`, not a tuple

profiles/api.py
```diff
@@ -1,20 +1,49 @@
...
+    USERNAME_MAX_LEN,
+)
+from profiles.exceptions import AlumsImportException
+from profiles.utils import usernameify
 
 User = get_user_model()
 
 CASE_INSENSITIVE_SEARCHABLE_FIELDS = {"email"}
 
+Alums = namedtuple(
```
noisecapella, line 35: Since this is referring to one person, I think this should be `Alum` instead here and elsewhere

profiles/api.py
```diff
@@ -256,3 +285,119 @@ def is_user_info_complete(user):
...
         and user.legal_address.is_complete
     )
+
+
+def import_alums(alums):
+    """
+    Import alums
+
+    Args:
+        alums (Alums namedtuple)
```
noisecapella, line 295: This should be something like `alum (Alum): ` per my comment above

profiles/api.py
```diff
@@ -256,3 +285,119 @@ def is_user_info_complete(user):
...
+        user.username = usernameify("", user.email)
+        user.save()
+
+        log.info(
+            "User does not exist against that email=%s but created without password and incomplete profile",
+            alums.learner_email,
+        )
+    bootcamp_start_date = datetime.strptime(alums.bootcamp_start_date, "%Y-%m-%d")
+    bootcamp_end_date = datetime.strptime(alums.bootcamp_end_date, "%Y-%m-%d")
+    if is_naive(bootcamp_start_date):
```
noisecapella, line 317: I think this is always true since there is no time zone information from the start or end date

profiles/api.py
```diff
@@ -256,3 +285,119 @@ def is_user_info_complete(user):
...
+        bootcamp_start_date = make_aware(bootcamp_start_date)
```
noisecapella, line 318: Since we don't know the timezone, and since the date is the only important part anyway, we should set the timezone to UTC here and in the next line: `make_aware(bootcamp_start_date, tz=pytz.UTC)`.

profiles/api.py
```diff
@@ -256,3 +285,119 @@ def is_user_info_complete(user):
...
+            alums.bootcamp_run_title,
+        )
+
+
+def parse_alumni_csv(csv_path):
+    """
+    Read CSV file and convert to Alumns object
+
+    Args:
+        csv_path(str): Path to the CSV file
```
noisecapella, line 354: Can we say `str or Path` here?

profiles/api.py
```diff
@@ -256,3 +285,119 @@ def is_user_info_complete(user):
...
+
+        return alumni
+
+
+def import_alumni(csv_path):
+    """
+    Import alums from the csv file
+
+    Args:
+        csv_path (str): Path to a csv file
```
noisecapella, line 389: Can we say `str or Path` here?

noisecapella (changes requested):

noisecapella (approved):

Issue #1182, opened 2021-03-28
Imported alums should not be blocked from getting a certificate


pdpinch:
In issue #1167 / PR #1171 we added a management command to import alums and create enrollments for them. By definition, the alums we are importing have all earned a certificate for the bootcamp they enrolled in (if they did not earn a certificate, we would not be importing them). Therefore, the value of `user_certificate_is_blocked` should be False when they are imported. 

https://github.com/mitodl/bootcamp-ecommerce/blob/9840936e4c12a750b72117a41b5436bacc9686d9/profiles/api.py#L334

umarmughal824:
@pdpinch after that removal, we have no other way to differentiate between the alumni and the normal user. So we will have to wait for https://github.com/mitodl/bootcamp-ecommerce/issues/1166 to get merged.

pdpinch:
Ok, in that case, let's get #1166 functional and merged first before working on this.

PR #1187, opened 2021-03-31
Allow alumni to skip application steps

### LINKED ISSUE #1166: Allow alumni to skip application steps
_Related to https://github.com/mitodl/bootcamp-ecommerce/issues/1010. This issue may end up closing that one. After this issue is done, we'll want to check the requirements of #1010 and if they're totally satisfied, we can close that issue too._

We want users who have been able to skip every application step after profile completion and before payment if they're alumni (meaning they have enrolled in a bootcamp some time in the past). We have a `BootcampRunCertificate` model, but this would only contain records for users that have (a) completed a bootcamp run, which may be fewer people than have successfully enrolled, and (b) have enrolled after our "v2" update of this application (summer 2020). We have many learners who enrolled before then whom we would still want to be able to skip to payment for a new bootcamp.

To summarize: if a user has enrolled in a bootcamp at any time in the past, we want them to be able to skip the resume and interview steps and go straight to payment (as long they have filled out their profile info)

**Technical details**

First thing we need to figure out is if we would want users to be able to skip the application steps for _all_ bootcamp runs, or only for specific ones. 

Once we figure that out, this is roughly what it will take to implement:

- Create a flag for indicating whether or not a user can skip application steps. We could call this `can_skip_application` or something like that (we could do something like `is_alumni`, but that's not as specific to the purpose this flag will be used for). It would be ideal to add this to the `User` model, but we don't have a custom User model in this project and it probably wouldn't make sense to add one for just this purpose. We can get away with adding it to `Profile`. Profile info is serialized in the application detail API, so it will be available when we need it.
- Add a flag to the `BootcampRun` model - something like `allows_skipped_steps`. Only when this flag is set to `True` _and_ the user's `can_skip_application` flag is set to `True` can a user skip the application steps
- Add a state transition method in the `BootcampApplication` model which can take a user from awaiting resume, awaiting interview, or awaiting review to awaiting payment. If a user has the flag above set to `True`, then they complete the profile completion step or start an application with the profile already complete, they should be transitioned straight to awaiting payment.
- `renderApplicationDetailSection` in `static/js/pages/applications/ApplicationDashboardPage.js` is the method that renders the application "progress meter" when you expand a bootcamp application on the dashboard page. This long function basically renders each application step and the "progress meter" state for each step. For users that have the `can_skip_application` flag set to `True`, we'll basically want to add some logic render `null` for the skipped steps. That should remove the steps that the user doesn't need to fill out from the UI.
- If we go with the approach where we use a flag to indicate whether a user can skip application steps (see first bullet), we will need to add some logic for setting that flag to `True` where appropriate. We'll definitely want to do that whenever a `BootcampRunEnrollment` record is created for the first time. There may be other scenarios as well

At some point we're going to want to "import" users that have enrolled in a legacy bootcamp run before we updated this app in the summer of 2020. For those users, a `Profile` object will be created for them one way or another, and we'll just need to make sure the `can_skip_application` flag is set to `True` for them

Developers can use the `set_application_state` management command to manipulate an application into a certain state. Run with `--help` to see the options.

umarmughal824:
fixes #1166

#### What's this PR do?
Allow alumni to skip application steps 

Need to confirm that for the alumni it skips application steps and goes direc to `AWAITING_PAYMENT`

- run the migrations
- set `can_skip_application_steps` to `True` in the user profile
- set `allows_skip_steps` to `True` in the bootcamp run 
- apply for that specific bootcamp run

(Optional)

(Optional)

**Alumni**

Screenshot 2021-03-31 at 14 45 00
Screenshot 2021-03-31 at 14 45 55

**Non-Alumni**

Screenshot 2021-03-31 at 18 22 28

**ADMIN PANEL**

Screenshot 2021-04-06 at 13 34 23
Screenshot 2021-04-06 at 13 34 57


pdpinch:
It would be preferable not to offer the application to users who aren't alumi or don't have an enrollment. Can we do that in this PR?

umarmughal824:
> It would be preferable not to offer the application to users who aren't alumi or don't have an enrollment. Can we do that in this PR?

@pdpinch I did not understand that what do you mean?

profiles/models.py
```diff
@@ -183,6 +183,7 @@ class Profile(TimestampedModel):
     highest_education = models.CharField(
         max_length=60, blank=True, default="", choices=HIGHEST_EDUCATION_CHOICES
     )
+    can_skip_application_steps = models.BooleanField(default=False)
```
gsidebo, line 186: Can you add this to the admin class too? (`ProfileAdmin`)

applications/models.py
```diff
@@ -299,6 +299,25 @@ def refund(self):
             change_status=ENROLL_CHANGE_STATUS_REFUNDED,
         )
 
+    @property
+    def is_eligible_to_skip_steps(self):
```
gsidebo, line 303: I think we can just get rid of this and use `check_eligibility_to_skip_steps` directly in `get_or_create_bootcamp_application`

static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -418,10 +418,20 @@ export class ApplicationDashboardPage extends React.Component<Props, State> {
...
       paymentReady = false
     } else {
-      // If there are no submissions required for this application, payment should be ready after the resume
-      // requirement is fulfilled. Otherwise, it should only be ready if the last submission was approved.
-      paymentReady =
-        submissionApproved === undefined ? resumeFulfilled : submissionApproved
+      if (
+        currentUser.profile &&
+        currentUser.profile.can_skip_application_steps === true &&
+        application.bootcamp_run.allows_skipped_steps === true
```
gsidebo, line 424: We should create a helper function for this in `static/js/lib/applicationApi.js`. We can call it something like `isEligibleToSkipSteps`

static/js/pages/applications/ApplicationDashboardPage.js
```diff
@@ -445,9 +455,15 @@ export class ApplicationDashboardPage extends React.Component<Props, State> {
...
-        {submissionStepRows}
+        {currentUser.profile &&
+        currentUser.profile.can_skip_application_steps === true &&
+        application.bootcamp_run.allows_skipped_steps === true ? null : (
+            <Fragment>
+              {profileRow}
+              {resumeRow}
+              {submissionStepRows}
+            </Fragment>
+          )}
```
gsidebo, line 466: If you implement the helper function I recommended above, this can be refactored... 

```suggestion
        {!isEligibleToSkipSteps(currentUser, application) && (
            <Fragment>
              {profileRow}
              {resumeRow}
              {submissionStepRows}
            </Fragment>
          )}
```

static/js/pages/applications/ApplicationDashboardPage_test.js
```diff
@@ -385,6 +387,50 @@ describe("ApplicationDashboardPage", () => {
...
+          assert.isFalse(profileDetail.exists())
+          assert.isFalse(resumeDetail.exists())
+          assert.isFalse(videoInterviewDetail.exists())
+          assert.isFalse(reviewDetail.exists())
+        } else {
+          assert.isTrue(profileDetail.exists())
+          assert.isTrue(resumeDetail.exists())
+          assert.isTrue(videoInterviewDetail.exists())
+          assert.isTrue(reviewDetail.exists())
+        }
```
gsidebo, line 429: This can be edited down

```suggestion
        assert.equal(profileDetail.exists(), isAlumni)
        assert.equal(resumeDetail.exists(), isAlumni)
        assert.equal(videoInterviewDetail.exists(), isAlumni)
        assert.equal(reviewDetail.exists(), isAlumni)
```

gsidebo (changes requested):
This looks great! Just a few code style/refactoring suggestions

profiles/models.py
umarmughal824: There is no such class in our project and if you could see the attached screenshots in the PR. This field is already showing in the profile admin section.

applications/api.py
```diff
@@ -43,7 +43,11 @@ def get_or_create_bootcamp_application(user, bootcamp_run_id):
             bootcamp_app.save()
 
     if created:
-        tasks.populate_interviews_in_jobma.delay(bootcamp_app.id)
+        if bootcamp_app.is_eligible_to_skip_steps:
+            bootcamp_app.skip_application_steps()
+            bootcamp_app.save()
+        else:
+            tasks.populate_interviews_in_jobma.delay(bootcamp_app.id)
```
umarmughal824, line 50: @gsidebo we still need to check the eligibility or so here because we don't want to populate the interviews for alumni type application.

applications/models.py
umarmughal824: In my opinion, we need to keep it as we need some checks to stop populating the jobma interview.

umarmughal824: Let my know what's your final opinion on that.

gsidebo (changes requested):
Just one last comment

profiles/models.py
gsidebo: Whoops! Sorry about that. Looks great

pdpinch:
> It would be preferable not to offer the application to users who aren't alumi or don't have an enrollment. Can we do that in this PR?

On the application dashboard, when the user clicks "Select Bootcamp" they get a dropdown list of available bootcamps. 

image

We should hide the bootcamp run in the dropdown when `allows_skip_steps` is True and `can_skip_application_steps` is False, preventing users who are ineligible from applying for the bootcamp. This could go in a follow-up PR if you prefer.

gsidebo (approved):
Looks great! 👍

umarmughal824:
@pdpinch As #1182 is blocked by it, so I will merge it after the approval and will put up a follow-up PR for it.

PR #1192, opened 2021-04-02
Move the django logging level from DEBUG to INFO

### LINKED ISSUE #1180: Move the logging level to INFO on RC
It would be better to move the log level from `DEBUG` to `INFO`. As on debug level, it shows too many details which are not really needed as we are creating a replica of our production.

umarmughal824:
fixes #1180 

#### What's this PR do?
Move the Django logging level from DEBUG to INFO

(Required)

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -552,7 +552,7 @@ LOG_LEVEL = get_string(
     "BOOTCAMP_LOG_LEVEL", "INFO", description="The logging level for the application"
 )
 DJANGO_LOG_LEVEL = get_string(
-    "DJANGO_LOG_LEVEL", "DEBUG", description="The log level for Django"
+    "DJANGO_LOG_LEVEL", "INFO", description="The log level for Django"
 )
 ES_LOG_LEVEL = get_string(
     "ES_LOG_LEVEL", "INFO", description="The log level for elasticsearch"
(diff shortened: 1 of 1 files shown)


main/settings.py
```diff
@@ -552,7 +552,7 @@
     "BOOTCAMP_LOG_LEVEL", "INFO", description="The logging level for the application"
 )
 DJANGO_LOG_LEVEL = get_string(
-    "DJANGO_LOG_LEVEL", "DEBUG", description="The log level for Django"
+    "DJANGO_LOG_LEVEL", "INFO", description="The log level for Django"
```
arslanashraf7, line 555: Looks fine but I think it can be more configurable with the use of `DEBUG` flag if that is set specifically e.g. in the case of RC and prod the `DEBUG` flag will be false and I think keeping that in mind we can do something like below. What do you think?

```suggestion
    "DJANGO_LOG_LEVEL", "DEBUG" if DEBUG else "INFO", description="The log level for Django"
```

umarmughal824: I added it to keep it consistent with the other settings `LOG_LEVEL` and `ES_LOG_LEVEL`

umarmughal824: ok, that makes more sense to follow the `DEBUG`.

arslanashraf7 (approved):
👍

arslanashraf7:
@umarmughal824, i missed it but you might want to take a look at the CI checks.

PR #1202, opened 2021-04-14
make import alumni case insensitive

### LINKED ISSUE #1201: Bug: import_alums shouldn't be case sensitive
### Steps to Reproduce

1. Find the email address for an existing use, for example bootcamper@example.com
2. Use the `import_alums` management command to import an alum with a capitalized version of the email address, for example BOOTCAMPER@EXAMPLE.COM

### Expected results

- No user account created
- An enrollment record should be created, with a foreign key to the user with email bootcamper@example.com

### Actual results

- New user account created with email BOOTCAMPER@EXAMPLE.COM
- Enrollment record created, with a foreign key to the user with email BOOTCAMPER@EXAMPLE.COM

umarmughal824:
fixes #1201 

#### What's this PR do?
make import alumni (`email`) case insensitive

Please perform these following steps

1. Convert the emails to uppercase in the testdata
for example
`bootcamper@example.com` -> `BOOTCAMPER@EXAMPLE.COM`

2. Just run the import alumni command after updating the
`docker-compose exec web python manage.py import_alumni profiles/testdata/example_alumni.csv`

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/profiles/api.py
+++ b/profiles/api.py
@@ -296,7 +296,7 @@ def import_alum(alum):
         alum (Alum namedtuple)
     """
     try:
-        user = User.objects.get(email=alum.learner_email)
+        user = User.objects.get(email__iexact=alum.learner_email)
     except User.DoesNotExist:
         user = User.objects.create(
             email=alum.learner_email, username=alum.learner_email


profiles/api.py
```diff
@@ -367,7 +367,7 @@ def parse_alumni_csv(csv_path):
 
         alumni = [
             Alum(
-                learner_email=row[ALUM_LEARNER_EMAIL],
+                learner_email=str(row[ALUM_LEARNER_EMAIL]).lower(),
```
pdpinch, line 370: Will this handle the case where the existing user's email is uppercase and the imported alum email is lowercase?

profiles/api_test.py
```diff
@@ -235,6 +242,35 @@ def test_parse_import_alumni_csv():
...
 
+def test_parse_import_alumni_csv_case_insensitive(tmp_path):
+    """parse_alumni_csv should import the user email case insensitive"""
+    path = tmp_path / "test.csv"
+    with open(path, "w") as f:  # create file
+        f.write(
+            f"{ALUM_LEARNER_EMAIL},{ALUM_BOOTCAMP_NAME},{ALUM_BOOTCAMP_RUN_TITLE},{ALUM_BOOTCAMP_START_DATE},{ALUM_BOOTCAMP_END_DATE}\n"
+        )
+        f.write("HDOOF@ODL.MIT.EDU,,,,\n")
+        f.write("EXAMPLE@ODL.MIT.EDU,,,,\n")
```
pdpinch, line 253: It would be better to use non-sendable email addresses:
```suggestion
        f.write("HDOOF@EXAMPLE.COM,,,,\n")
        f.write("EXAMPLE@EXAMPLE.COM,,,,\n")
```

profiles/api.py
umarmughal824: @pdpinch No, because normally we only the email addresses in lowercase.

umarmughal824: Normally we have that format `abc@example.com`. It considered a standard format.

HamzaIbnFarooq: @umarmughal824 what if we do a case insensitive search at https://github.com/mitodl/bootcamp-ecommerce/blob/master/profiles/api.py#L299 as well? and maybe convert email to lower while creating a user https://github.com/mitodl/bootcamp-ecommerce/blob/master/profiles/api.py#L301?

umarmughal824: Yeah in that case it these could be the possible combinations in my opinion like
'ABC@example.com'
'abc@EXAMPLE.COM'
'ABC@EXAMPLE.COM'

HamzaIbnFarooq (approved):
LGTM but you can wait for Peter's feedback.

umarmughal824:
@pdpinch do you wanna add something?

pdpinch (approved):
I think this should work. Let's get it deployed today!

umarmughal824:
> I think this should work. Let's get it deployed today!

Sure

Issue #1185, opened 2021-03-29
Create management command to retire users


briangrossman:
As an admin, I'd like a management command that can be used to retire an account, similar to the ones in [xPRO](https://github.com/mitodl/mitxpro/blob/master/users/management/commands/retire_users.py), [MicroMasters](https://github.com/mitodl/micromasters/blob/master/profiles/management/commands/retire_users.py) and [Open](https://github.com/mitodl/open-discussions/blob/master/profiles/management/commands/retire_user.py).

**Acceptance Criteria**
- [ ] set `is_active` on the account to False
- [ ] obfuscate the email address 
- [ ] obfuscate the password
- [ ] Question: Consider if there is any other data that needs to be altered
- [ ] Delete any social auth records

pdpinch:
@rhysyngsun is this something that should be a part of https://github.com/mitodl/ol-django ?

rhysyngsun:
I think probably no, our user/profile models differ a bit between applications and there are almost always other associated records we want to delete so it's best to just inline the command in each project.

pdpinch:
I added a checkbox to delete social auth records, which I believe will remove any vestigial edX social auths. We should also remove any authentication for NovoEd -- but I'm not sure if that's stored in the social auth table or not.

pdpinch:
We should also remove the user from Jobma as well, although I don't know how to do that.

asadiqbal08:
Do we need to add in acceptance criteria this ? 
- Delete user bootcamp run enrollments

briangrossman:
@asadiqbal08 

> Do we need to add in acceptance criteria this ?
> - Delete user bootcamp run enrollments

I don't think this is necessary for now. In xPRO, we don't remove enrollments. The user won't be able to access the bootcamp run enrollments... and keeping the enrollment will help ensure we have complete data.

PR #1205, opened 2021-04-16
Management command to retire users

asadiqbal08:
fixes: #1185 

#### What's this PR do?
A management command for retiring a user or list of users.

Retire one or multiple users. Username or email can be used to identify a user.

For single user use:\n
`./manage.py retire_users --user=foo` or do \n
`./manage.py retire_users -u foo` or do \n
`./manage.py retire_users -u foo@email.com` \n

For multiple users, add arg `--user` for each user i.e:\n
`./manage.py retire_users --user=foo --user=bar --user=baz` or do \n
`./manage.py retire_users --user=foo@email.com --user=bar@email.com --user=baz` or do \n
`./manage.py retire_users -u foo -u bar -u baz`


profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,129 @@
...
+            self.display_messages(f"Email changed from '{email}' to '{user.email}' and password is not useable now", log_messages)
+
+            # reset user social
+            auth_delete_count, _ = UserSocialAuth.objects.filter(user=user).delete()
+            self.display_messages(
+                f"For user {user}: {auth_delete_count} SocialAuth rows deleted",
+                log_messages
+            )
+
+            Interview.objects.filter(applicant=user).update(status='expired')
```
asadiqbal08, line 121: @pdpinch , I just found the user association in the Interview model of `Jobma ` application. Either deleting the records would be required here or keeping the interview status to `expired` would be good ?

pdpinch: In general I would prefer changing a status instead of deleting a record. But I don't know what the meaning of the statuses are. 

Can you ask @noisecapella or @mbertrand next week?

asadiqbal08:
Do we need to add in acceptance criteria this ? 
- Delete user bootcamp run enrollments

HamzaIbnFarooq (commented):
@asadiqbal08 should we add test cases for this command as well?

asadiqbal08:
Not sure because I did not notice tests for any management command in repository but it is good practice to have tests. Let me think about it.

HamzaIbnFarooq:
> Not sure because I did not notice tests for any management command in repository but it is good practice to have tests. Let me think about it.

If you plan to write them you can pick use https://github.com/mitodl/mitxpro/blob/master/users/management/commands/retire_users_test.py as well, at a glimpse I feel like it covers 90% of this PR's scenarios.

profiles/management/commands/retire_users.py
mbertrand: I think setting the status to expired should be adequate (but not sure if it's really necessary).  I'd suggest using a constant in the code if it's going to be changed:

```python
from jobma.contants import EXPIRED
.
.
.

    Interview.objects.filter(applicant=user).update(status=EXPIRED)
```

pdpinch:
What happens to the data in hubspot? (@briangrossman do you know?)

mbertrand (approved):
👍

briangrossman:
> What happens to the data in hubspot? (@briangrossman do you know?)

Looking into it, @pdpinch

briangrossman:
@asadiqbal08 

FYI: I did a quick test on RC and it seemed to work as expected. 
(I created a user and then retired them)

PR #1218, opened 2021-04-28
defer enrollment

### LINKED ISSUE #1208: Defer enrollment
Sometimes a bootcamper who has been admitted to and paid for a bootcamp run wants to defer to another bootcamp run.

We've done this manually so far (for example, https://odl.zendesk.com/agent/tickets/93651) but it would be best to have make this process repeatable, with a clear audit trail, similar to what we did in [xPRO](https://github.com/mitodl/mitxpro/blob/master/courses/management/commands/defer_enrollment.py).

### (suggested) acceptance criteria
- add a `change_status` status field to the enrollment model, with possible values of `deferred` and `refunded` (I don't think there are transfers in bootcamps)
- add an enrollment_audit model
- create a `defer_enrollment` management command similar to xPro that:
   - updates the status of the existing enrollment
   - creates a new enrollment with the same user and the new bootcamp run
   - update the order appropriately (I'm not exactly sure how we do this in xPRO, and whether it's appropriate for bootcamps) 

### Questions
- **Q:** bootcamp_application has a foreign key to the bootcamp_run. Should we update that as well?
 
### out of scope

- In xPRO, we have a google sheet integration that makes it possible for customer support to initiate an automated deferral without engineering intervention. Until we have more than a few of these, it's not worth the maintenance overhead of setting up the Google integration.

umarmughal824:
fixes #1208 

#### What's this PR do?
defer enrollment

just run the management command and check for the defer enrollment

`docker-compose exec web python manage.py defer_enrollment --user=admin --from-run=bootcamp-v1:private+MCR-f2f+R2 --to-run=bootcamp-v1:public+DER-f2f+R2 --order=2 -f`

(Optional)

(Optional)

Screenshot 2021-04-29 at 15 37 32
Screenshot 2021-04-29 at 15 37 41
Screenshot 2021-04-29 at 15 44 23
Screenshot 2021-04-29 at 15 49 59
Screenshot 2021-04-29 at 16 08 44
Screenshot 2021-04-29 at 16 15 56
Screenshot 2021-04-29 at 16 16 43
Screenshot 2021-04-29 at 16 33 42
Screenshot 2021-04-29 at 16 34 17
Screenshot 2021-04-29 at 16 53 55
Screenshot 2021-04-29 at 17 10 30
Screenshot 2021-04-29 at 17 10 45
Screenshot 2021-04-29 at 17 15 51
Screenshot 2021-04-29 at 17 21 35


klasses/api.py
```diff
@@ -210,3 +212,121 @@ def fetch_bootcamp_run(run_property):
...
+                application = order.application
+                application.bootcamp_run = run
+                application.save()
+
+            if (
+                features.is_enabled(features.NOVOED_INTEGRATION)
+                and enrollment.bootcamp_run.novoed_course_stub
+            ):
+                novoed_tasks.enroll_users_in_novoed_course.delay(
+                    user_id=enrollment.user.id,
```
gsidebo, line 250: This code is functionally very similar to `applications.models.BootcampApplication.complete`, which makes sense. Having 2 copies of this logic opens us up to bugs in a very critical area of the application. We should move this logic to a separate method so there's just one place where we handle a user completing enrollment in a bootcamp run. Also, we should create a separate method for enrolling a user in a single run (`create_run_enrollments` just performs the enrollment actions and aggregates the results).

Changes would look something like this:

```python
# models.py
class BootcampApplication(TimestampedModel):
  def complete(self):
    create_run_enrollment(self.user, self.bootcamp_run)

# klasses/api.py
def defer_enrollment(
    user, from_bootcamp_run_id, to_bootcamp_run_id, order_id, force=False
):
  # ...
  to_enrollments = create_run_enrollment(user, to_run, order=order)

def create_run_enrollment(user, run, order=None):
    enrollment, _ = BootcampRunEnrollment.objects.update_or_create(
        user=user,
        bootcamp_run=run,
        defaults={"active": True, "change_status": None},
    )
    if order:
        application = order.application
        application.bootcamp_run = run
        application.save()
    if (
        features.is_enabled(features.NOVOED_INTEGRATION)
        and run.novoed_course_stub
    ):
        novoed_tasks.enroll_users_in_novoed_course.delay(
            user_id=user.id,
            novoed_course_stub=run.novoed_course_stub,
        )
    return enrollment

def create_run_enrollment(user, run, order=None):
    successful_enrollments = []

    for run in runs:
        try:
            enrollment = create_run_enrollment(user, run, order=None)
    # ...
    return successful_enrollments
```

gsidebo (changes requested):
Sorry for the long delay on this review!

klasses/api_test.py
```diff
@@ -177,3 +187,216 @@ def test_fetch_bootcamp_run_dates(end_date_params):
...
+        "klasses.api.BootcampRunEnrollment.objects.get_or_create",
+        side_effect=[Exception()],
+    )
+    patched_novoed_enroll = mocker.patch(
+        "klasses.api.novoed_tasks.enroll_users_in_novoed_course.delay"
+    )
+
+    with pytest.raises(Exception):
+        successful_enrollments = create_run_enrollments(user, [run], order=None)
+        patched_novoed_enroll.assert_not_called()
```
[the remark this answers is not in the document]
umarmughal824: It does not raise any exception if we move it out of that.

klasses/conftest.py
```diff
@@ -16,3 +27,39 @@ def mocked_requests_get(mocker):
...
+
+
+@pytest.fixture()
+def patched_deactivate_run_enrollment(mocker):
+    """patched deactivate_run_enrollment"""
+    return mocker.patch("klasses.api.deactivate_run_enrollment", return_value=[])
+
+
+@pytest.fixture()
+def enrollment_data(user):
```
gsidebo, line 45: If any of these new fixtures are only used in `klasses/api_test.py`, we should define them in that module instead of here

klasses/api.py
```diff
@@ -210,3 +212,144 @@ def fetch_bootcamp_run(run_property):
...
+        order_id (int): The order_id value for an user's order ID
+        force (bool): If True, the deferral will be completed even if the current enrollment is inactive
+            or the desired enrollment is in a different bootcamp
+
+    Returns:
+        (BootcampRunEnrollment, BootcampRunEnrollment): The deactivated enrollment paired with the
+            new enrollment that was the target of the deferral
+    """
+    from_enrollment = BootcampRunEnrollment.objects.get(
+        user=user, bootcamp_run__bootcamp_run_id=from_bootcamp_run_id
```
gsidebo, line 307: `BootcampRun.bootcamp_run_id` is a nullable field. I don't think we can use it for this purpose. We should pass in two `BootcampRun` objects for the 'from' and 'to' bootcamp runs, and use a different function for fetching those runs in the management command. If we want to make it user-friendly, we can allow users to specify any one of multiple fields of the `BootcampRun`s they want, like `id`, `title`, or `bootcamp_run_id`

klasses/api_test.py
```diff
@@ -177,3 +185,249 @@ def test_fetch_bootcamp_run_dates(end_date_params):
     assert fetch_bootcamp_run(run.display_title) == run
     with pytest.raises(BootcampRun.DoesNotExist):
         fetch_bootcamp_run("invalid")
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("novoed_integration", [True, False])
+def test_create_run_enrollments(mocker, user, settings, novoed_integration):
```
gsidebo, line 192: `create_run_enrollment` is now the function that has the critical functionality, and `create_run_enrollments` is just a helper that calls that function and aggregates results. We should change this test case to test `create_run_enrollment` directly, then add a test case for `create_run_enrollments` which patches `create_run_enrollment`

gsidebo (changes requested):

klasses/conftest.py
umarmughal824: @gsidebo I have used it in the same file but it was generating that so decided to move it in `conftest`

```
___________________________________ [pylint] klasses/api_test.py ____________________________________
W:311, 4: Redefining name 'patched_create_run_enrollments' from outer scope (line 303) (redefined-outer-name)
W:332, 4: Redefining name 'patched_create_run_enrollments' from outer scope (line 303) (redefined-outer-name)
W:353, 4: Redefining name 'patched_create_run_enrollments' from outer scope (line 303) (redefined-outer-name)
W:374, 4: Redefining name 'patched_create_run_enrollments' from outer scope (line 303) (redefined-outer-name)
W:395, 4: Redefining name 'patched_create_run_enrollments' from outer scope (line 303) (redefined-outer-name)
W:416, 4: Redefining name 'patched_create_run_enrollments' from outer scope (line 303) (redefined-outer-name)```

File: klasses/api.py [hunk 6]
```diff
@@ -210,3 +212,144 @@ def fetch_bootcamp_run(run_property):
...
+    """
+    enrollment, _ = BootcampRunEnrollment.objects.update_or_create(
+        user=user,
+        bootcamp_run=run,
+        defaults={"active": True, "change_status": None},
+    )
+
+    try:
+        enrollment.user.profile.can_skip_application_steps = True
+        enrollment.user.profile.save()
```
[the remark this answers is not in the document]
umarmughal824: I can't trust it while running the management command it would be either username or id but not the user instance so that's why preferring to use the `enrollment.user.profile`

klasses/conftest.py
gsidebo: We typically ignore those linting errors (`redefined-outer-name`) in Python test files. You can grep the codebase for the many places where we ignore it. It has to do with the fundamental difference between the way that function arguments in Python are normally used and the way that they're used in pytest test cases.

klasses/api.py
umarmughal824: @pdpinch what's your opinion on that, should we use `bootcamp_run_id` or we should also add support for (id, title)?

pdpinch: Thanks for the discussion. `bootcamp_run_id` is nullable because we have legacy (and test) bootcamp runs that may never get an appropriate `bootcamp_run_id`.

However, any bootcamp run that we can actually defer from or to, **_will_** have a bootcamp_run_id. As an admin, if the field is null, that means I haven't done my job. 

We don't need the added complexity of supporting `id` or `title` inputs.

umarmughal824: While considering Peter's comment, that's ready for review again.

klasses/api_test.py
```diff
@@ -289,41 +320,44 @@ def test_defer_enrollment(mocker, user, force):
...
-        # Deferring to the same bootcamp run should raise a validation error
-        defer_enrollment(
-            user,
-            enrollment_data.enrollments[0].bootcamp_run.bootcamp_run_id,
-            enrollment_data.enrollments[0].bootcamp_run.bootcamp_run_id,
-            enrollment_data.order.id,
-        )
-    patched_create_run_enrollments.assert_not_called()
+@pytest.fixture()
+def enrollment_data(user):
```
gsidebo, line 334: Nit: our current pattern is to define fixtures before all test cases in the file. This includes `patched_create_run_enrollment` etc.

klasses/api_test.py
```diff
@@ -187,11 +192,35 @@ def test_fetch_bootcamp_run_dates(end_date_params):
...
+    expected_calls = []
+    for run in bootcamp_runs:
+        expected_calls.append(mocker.call(user, run, None))
+    patched_create_run_enrollment.assert_has_calls(expected_calls)
+
+
 @pytest.mark.django_db
 @pytest.mark.parametrize("novoed_integration", [True, False])
-def test_create_run_enrollments(mocker, user, settings, novoed_integration):
+def test_create_run_enrollment(mocker, user, settings, novoed_integration):
```
gsidebo, line 221: This test case should check to see that `can_skip_application_steps` is set to `True`. If you can't guarantee that `can_skip_application_steps` is set to `False` when you're using that `user` fixture, you'll need to use `ProfileFactory` in this test case to create a new user

gsidebo (changes requested):
Last comment!

gsidebo (approved):
Looks great! 👍  Thanks for all your attention to this

umarmughal824:
> Looks great! 👍 Thanks for all your attention to this

My pleasure.

PR #1224, opened 2021-05-07
asadiqbal08/Update the merge package

asadiqbal08:
https://github.com/mitodl/bootcamp-ecommerce/security/dependabot/yarn.lock/merge/open

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "jsdom-global": "^3.0.2",
     "keycode": "^2.2.0",
     "lodash": "^4.17.19",
+    "merge": "^2.1.1",
     "mini-css-extract-plugin": "^0.9.0",
     "minimist": "^1.2.5",
     "mocha": "^7.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "jsdom-global": "^3.0.2",
     "keycode": "^2.2.0",
     "lodash": "^4.17.19",
+    "merge": "^2.1.1",
```
arslanashraf7, line 61: I checked that the `merge` package is only used by `sass-lint` which is already upgraded to maximum. In this case pinning this dependency would just resolve dependabot alert but not actually do anything else since `sass-lint` will still use `merge(1.2.0)` inside.

Instead of pinning we should go for dependency resolution and check that it doesn't breaking any existing thing You can take a look at this link for reference https://classic.yarnpkg.com/en/docs/selective-version-resolutions.

arslanashraf7 (changes requested):

package.json
asadiqbal08: Do you mean merge should be in `resolutions ` ?

```
"resolutions": {
    "merge": "^2.1.1"
  },
```

arslanashraf7 (approved):
👍

PR #1225, opened 2021-05-07
asadiqbal08/version bump property expr

asadiqbal08:
https://github.com/mitodl/bootcamp-ecommerce/security/dependabot/yarn.lock/property-expr/open

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "postcss-loader": "^3.0.0",
     "prettier-eslint": "^9.0.1",
     "prettier-eslint-cli": "^5.0.0",
+    "property-expr": "^2.0.3",
     "query-string": "^6.4.0",
     "raf": "^3.4.1",
     "ramda": "^0.27.0",
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "postcss-loader": "^3.0.0",
     "prettier-eslint": "^9.0.1",
     "prettier-eslint-cli": "^5.0.0",
-    "property-expr": "^2.0.3",
     "query-string": "^6.4.0",
     "raf": "^3.4.1",
     "ramda": "^0.27.0",
@@ -117,7 +116,7 @@
     "webpack-cli": "^3.3.12",
     "webpack-dev-middleware": "^3.7.2",
     "webpack-hot-middleware": "^2.25.0",
-    "yup": "^0.27.0"
+    "yup": "^0.30.0"
   },
   "engines": {
     "node": "13.13.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -71,6 +71,7 @@
     "postcss-loader": "^3.0.0",
     "prettier-eslint": "^9.0.1",
     "prettier-eslint-cli": "^5.0.0",
+    "property-expr": "^2.0.3",
```
arslanashraf7, line 74: It would be a good idea to upgrade this dependency through `yup`, which uses this package instead of pinning it.
I checked a bit and it looks like `yup(0.30.0)` https://github.com/jquense/yup/blob/v0.30.0/package.json supports `property-expr(2.0.4)`. We might need to upgrade tests just in case.

arslanashraf7 (changes requested):

arslanashraf7:
@asadiqbal08 Can you please take a look at the failing JS tests, We might need to make some updates.

asadiqbal08:
@gsidebo I tried it out today again to understand the enzyme js test behavior but no luck to get about the test failure. It seems to me, `findFormikErrorByName` is working fine to get the element but somehow `formik` errors are not rendering either the `simulate("change")` is not working ...

gsidebo:
Just FYI the problems with this upgrade go beyond the test suite. Form validation is broken for at least a few of these forms. If you create a new account and start filling out the first registration form, there is are no error messages when you focus and blur a required field, and clicking "Continue" with empty fields just does nothing. If you do the same in RC, error messages show up as soon as you blur a required field (like First Name) and don't enter anything for the value.

static/js/components/forms/EditProfileForm_test.js
```diff
@@ -61,7 +61,7 @@ describe("EditProfileForm", () => {
     wrapper.update()
     assert.deepEqual(
       findFormikErrorByName(wrapper, "legal_address.street_address").text(),
-      "Street address is a required field"
+      "Street address field must have at least 1 items"
```
gsidebo, line 64: This is a less helpful error message than what we had before. Can you look into changing it?

gsidebo (changes requested):

static/js/components/forms/EditProfileForm_test.js
asadiqbal08: sure.

asadiqbal08: Sorry, I re requested you the change, let me update it then tagged you again.

asadiqbal08:
are we good to merge ? @arslanashraf7 @gsidebo

asadiqbal08:
@gsidebo any thought ?

PR #1251, opened 2021-06-07
upgrade django to 2.2.21

umarmughal824:
NA, fix security alert

#### What's this PR do?
upgrade django to 2.2.21

Just test that everything is up


arslanashraf7 (approved):
👍

PR #1260, opened 2021-06-14
bump urllib3=1.26.5 through boto3 and requests

arslanashraf7:

Alert: https://github.com/mitodl/bootcamp-ecommerce/security/dependabot/requirements.txt/urllib3/open

#### What's this PR do?
- Updates a transitive dependency `urllib3` to version `1.26.5` as suggested by a security alert
- In other to do the above we had to update `boto3` and `requests` as well. (So, here they are upgraded as well)

Check basic file uploads or any functionality dependent on `boto3 and requests`

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,6 +1,6 @@
 Django==2.2.24
 celery==4.3.0
-boto3==1.12.42
+boto3==1.16.63
 dj-database-url==0.3.0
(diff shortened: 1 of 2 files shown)


HamzaIbnFarooq (approved):
LGTM, might need to test on RC with old data present there.

Issue #1262, opened 2021-06-16
Fix warning message when running management commands


briangrossman:
When you run management commands, you get the warning below. Please make the necessary changes to fix this warning.

```
System check identified some issues:

WARNINGS:
cms.SignatoryPage.signature_image: (wagtailcore.W001) Field hasn't specified on_delete action
	HINT: Set on_delete=models.SET_NULL and make sure the field is nullable or set on_delete=models.PROTECT. Wagtail does not allow simple database CASCADE because it will corrupt its tree storage.
jobma.Job.run: (fields.W342) Setting unique=True on a ForeignKey has the same effect as using a OneToOneField.
	HINT: ForeignKey(unique=True) is usually better served by a OneToOneField.
```

Note that you can get this to happen on RC using `retire_users`:
```
~ $ python manage.py retire_users --user=[username]
```

asadiqbal08:
@briangrossman can you confirm the warning showing up --> `cms.SignatoryPage.signature_image`, that is related to wagtail, is coming up by this command? 
I mean, I did not notice any connection in `retire_users` command with wagtail CMS models.

Locally, I run:
```
./manage.py retire_users --user=admin2
```

asadiqbal08:
But I find out that these warnings are related to models in CMS and Jobma application. I update the models and create migrations to resolve these warnings.

briangrossman:
@asadiqbal08 

It looks like you might have found a solution. Regarding your question... the error was coming up with many (if not all) management commands. It doesn't necessarily need to be the `retire_users` command. That was just an example of a command that showed the error. 

Let me know if you have other questions.

pdpinch:
Can this be closed?

PR #1274, opened 2021-06-28
asadiqbal08/Fix for Django model warning messages.

asadiqbal08:
fixes: #1262 

#### What's this PR do?
Over the console, some of the warnings were rendered. Did the necessary changes to fix this warning.

```
System check identified some issues:

WARNINGS:
cms.SignatoryPage.signature_image: (wagtailcore.W001) Field hasn't specified on_delete action
	HINT: Set on_delete=models.SET_NULL and make sure the field is nullable or set on_delete=models.PROTECT. Wagtail does not allow simple database CASCADE because it will corrupt its tree storage.
jobma.Job.run: (fields.W342) Setting unique=True on a ForeignKey has the same effect as using a OneToOneField.
	HINT: ForeignKey(unique=True) is usually better served by a OneToOneField.
```

Test the application behavior and also noticed any console warnings.


jobma/models.py
```diff
@@ -15,7 +15,7 @@ class Job(AuditableModel):
     job_code = models.TextField()
     job_title = models.TextField()
     interview_template_id = models.IntegerField()
-    run = models.ForeignKey(BootcampRun, on_delete=models.CASCADE, unique=True)
+    run = models.ForeignKey(BootcampRun, on_delete=models.CASCADE)
```
HamzaIbnFarooq, line 18: It conflicts with https://github.com/mitodl/bootcamp-ecommerce/pull/738. Secondly, when the error says:
```
jobma.Job.run: (fields.W342) Setting unique=True on a ForeignKey has the same effect as using a OneToOneField.
	HINT: ForeignKey(unique=True) is usually better served by a OneToOneField.
```
It asks to make it an OneToOneField rather than removing the unique constraint, corrupting the whole logic.

HamzaIbnFarooq (changes requested):

jobma/models.py
asadiqbal08: Actually, putting a unique constraint is not recommended on a ForeignKey so either we need to convert field to 1-1 or remove the unique constraint. 
I am not sure, what are the actual requirements for making a unique foreign key earlier. @noisecapella can you please suggest about the reasons in the [PR](https://github.com/mitodl/bootcamp-ecommerce/pull/738)?

pdpinch:
The title and description of PR #738 suggests to me that the goal was to ensure that:

> there's only one Job per BootcampRun

Are there any downsides to changing it from a ForeignKey to a OneToOneField?

asadiqbal08:
Apparently, Not seems any effect but I suggest a separate story for this model relation change. thought ?

jobma/models.py
noisecapella: I would keep the unique constraint and use a `OneToOneField` as @HamzaIbnFarooq suggested. I think a `OneToOneField` would have an identical database schema as a `ForeignKey` with a unique constraint so it should only require code changes, hopefully minor ones

pdpinch:
@asadiqbal08 what's the status of this?

jobma/models.py
umarmughal824: OneToOneField always carries a "unique" constraint with it so no need to specify it

umarmughal824 (approved):
looks good 👍

PR #1276, opened 2021-07-01
add validation for linkedin url

arslanashraf7:
#1246 

#### What's this PR do?
Adds two types of validation on the LinkedIn URLs:
- The length of the URL is not more than 200
- The URL format is correct for a profile

- Run the app
- Add a Linkedin URL with more than 200 characters (You should see an error message)
- Add a non-LinkedIn or ill-formatted LinkedIn URL (You should see an error message)
- You should be able to verify the LinkedIn URL pattern introduced in this PR.

- On Home page click `Apply Now`
- Next step, Click `View/Edit Resume or LinkedIn Profile`
- Next step, Add URLs in the linked in the field

Some users were adding URLs having a length of more than 200 characters.

**When URL is not valid (If its is not a linked in or URL or an invalid linkedin URL)**
Screenshot 2021-07-07 at 5 49 35 PM

**When URL length is more then 200**
Screenshot 2021-07-07 at 5 50 29 PM

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -233,7 +233,15 @@ class UploadResumeView(GenericAPIView):
                 {"errors": "The URL should be less than 200 characters."}
             )
 
-        regex = re.compile("^https://[a-z]{2,3}[.]linkedin[.]com/.*$", re.I)
+        regex = re.compile(
+            "^(http|https)://"  # Support for both http and https
+            "([a-zA-Z]{2,3}[.]|)"  # Support for global or localized prefix
+            "linkedin[.]"  # Contains the linkedin domain
+            "([a-zA-Z]{2,3})/"  # Support for .com or localized postfix
+            "+([a-zA-Z0-9-_])"  # Support for /<in or org>
+            "+/+([a-zA-Z0-9-_])+.*$",  # Any type of username
+            re.I,
+        )
         if not regex.match(str(linkedin_url)):
             raise ValidationError({"errors": "Please enter a valid LinkedIn URL"})
 
--- a/applications/views.py
+++ b/applications/views.py
@@ -230,7 +230,7 @@ class UploadResumeView(GenericAPIView):
         """
         if len(linkedin_url) > 200:
             raise ValidationError(
-                {"errors": "The URL should be of 200 characters at max"}
+                {"errors": "The URL should be less than 200 characters."}
             )
 
         regex = re.compile("^https://[a-z]{2,3}[.]linkedin[.]com/.*$", re.I)
--- a/applications/views.py
+++ b/applications/views.py
@@ -1,6 +1,7 @@
 """Views for bootcamp applications"""
 from collections import OrderedDict
 
+import re
 from django.db.models import Count, Subquery, OuterRef, IntegerField, Prefetch, Q
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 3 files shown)


applications/views.py
```diff
@@ -218,6 +221,22 @@ def post(self, request, *args, **kwargs):
...
+
+        Args:
+            linkedin_url (string): LinkedIn URL of a user
+        """
+        if len(linkedin_url) > 200:
+            raise ValidationError(
+                {"errors": "The URL should be of 200 characters at max"}
+            )
+
+        regex = re.compile("^https://[a-z]{2,3}[.]linkedin[.]com/.*$", re.I)
```
pdpinch, line 236: I think it makes sense to check the URL for a linkedin domain name, but are you sure that they use a shortened or localized domain? 

We should at least take a look at the data we've already collected from users.

applications/views.py
```diff
@@ -218,6 +221,22 @@ def post(self, request, *args, **kwargs):
...
+    def validate_linkedin_url(self, linkedin_url):
+        """
+        Validate that a LinkedIn URL has the right format and length
+
+        Args:
+            linkedin_url (string): LinkedIn URL of a user
+        """
+        if len(linkedin_url) > 200:
+            raise ValidationError(
+                {"errors": "The URL should be of 200 characters at max"}
```
pdpinch, line 233:
```suggestion
                {"errors": "The URL should be less than 200 characters."}
```

Trying to keep the message as simple as possible.

pdpinch (changes requested):

applications/views.py
arslanashraf7: Correct, We can take a look at the collected data and base this logic on that. I can verify that data on RC would that be enough? or would it be a good idea to take a look at a more realistic data set?

briangrossman:
@arslanashraf7 
After reviewing the data on prod, discussing the option of loosening the restrictions, and talking with Mariah, let's proceed with the path you were on of checking for a valid LinkedIn URL. In Slack you said we could use this logic:

`http<s>://<www or country_code or empty>.linkedin.<com or country_code>/in/<any_value(e.g. Username)>`

This seems pretty good, but check out rows 4627 through 4734 which contain /mwlite/ in the URL (and seem to work). Also, it looks like you can use a /m/ as well:

For example:
- https://www.linkedin.com/mwlite/in/brian-grossman-705289
- https://www.linkedin.com/m/in/brian-grossman-705289

Thanks!

arslanashraf7:
@briangrossman Thanks for giving a thorough look at the data and identifying one other valid the usecase. 

>This seems pretty good, but check out rows 4627 through 4734 which contain /mwlite/ in the URL (and seem to work). Also, it looks like you can use a /m/ as well:

I've updated the PR with an updated LinkedIn URL validation pattern and I've also added some screenshots. Furthermore, I'm setting up the CI build for this PR so that you can take a look flexibly if you'd like.

briangrossman:
[Brian Grossman](https://trello.com/c/fG9vk9pv/6241-brian-grossman)

briangrossman:
@arslanashraf7 
I did a bunch of testing on the CI build using real data. It worked great. Functionally, I'm giving it a 👍

HamzaIbnFarooq (approved):
👍

Issue #1273, opened 2021-06-28
Migrate from node-sass to sass


arslanashraf7:
We should migrate to `sass` from `node-sass` since `node-sass` is now deprecated, and the pure-JS sass package is preferred. We have seen `node-sass` as a hurdle in upgrading our dependencies in our projects from time to time.

- This is [a reference](https://github.com/mitodl/micromasters/pull/4965) to how we did it on another project

arslanashraf7:
Closing, It's deployed on prod.

PR #1279, opened 2021-07-05
migrate from node-sass to sass

arslanashraf7:
#1273 

#### What's this PR do?
- Migrates from `node-sass` to `sass`. 
- Removes `node-sass/**/trim-newlines` resolution since it is no more required as we removed `node-sass`.
- Pins `sass` to `1.32.13` to get rid of unnecessary warnings from `bootstrap` which are currently not fixed even in the latest bootstrap 5. This should not affect us and we will update `sass` to 2.0.0 eventually when those warnings are fixed in a bootstrap update.

- Check CI build is passing
- Check that CSS is working fine without error

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "moment": "^2.24.0",
     "mutationobserver-shim": "^0.3.5",
     "named-urls": "^2.0.0",
-    "node-sass": "^4.14.1",
     "nyc": "^15.1.0",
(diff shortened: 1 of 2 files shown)


asadiqbal08:
Can you deploy it https://github.com/mitodl/mitxpro/pull/2273 on heroku to look css ?

asadiqbal08 (approved):

PR #1271, opened 2021-06-25
Update to use the common libraries from mitol-django

umarmughal824:
#### What are the relevant tickets?
fixes #1220 

#### What's this PR do?
Update to use the common libraries from mitol-django

#### How should this be manually tested?
Just verify that everything is functional

PR #1288, opened 2021-07-19
Release 0.83.1

odlbot:
## Umar Asghar
  - [x] Update to use the common libraries from mitol-django (#1271) ([da55c13f](../commit/da55c13f478d3559bfda66b8def3c16d41c0b40d))


pdpinch:
@umarmughal824 this can wait until after the holiday, but I don't want it to wait for too long next week.

PR #1370, opened 2022-11-01
Upgrade celery, redis

mbertrand:
N/A

#### What's this PR do?
Upgrades celery in a way that [doesn't break things](https://github.com/mitodl/bootcamp-ecommerce/pull/1341).

- Run `docker-compose build` then `docker-compose up`.  Celery should not throw any errors in the logs on startup
- Start a django shell and run the following
```python
from applications.tasks import *
refresh_pending_interview_links.delay()
```
- An async task should be returned and logs should indicate that the task completes successfully.


annagav (approved):
👍

PR #1371, opened 2022-11-01
Remove an obsolete github workflow

mbertrand:
N/A

#### What's this PR do?
Removes an obsolete workflow that was throwing errors in github actions

You should not see any new errors[ like this one ](https://github.com/mitodl/bootcamp-ecommerce/actions/runs/3371909880) under the "Actions" tab above for this branch.


collinpreston (commented):
This might be a silly question but it's not clear to me; what makes this workflow obsolete?

mbertrand:
I asked Tobias if we needed this anymore and he said it's okay to just get rid of it.

PR #1394, opened 2023-01-03
Remove email validation (not necessary) from a hubspot sync function

### LINKED ISSUE #1393: Email validation is used incorrectly and not needed in get_hubspot_id_for_object function
The `hubspot_sync.api.get_hubspot_id_for_object` function calls `validate_email` incorrectly in a way that will prevent `find_contact()` from being called when it should, causing some deals to not be associated with contacts in hubspot.

It should be removed altogether, because if the email is invalid or non-existent in hubspot, then the hubspot api call will return a 404 either way, which is fine.

mbertrand:
Closes #1393 

#### What's this PR do?
Removes email validation from a function where it is not really needed and is used incorrectly anyway

- With hubspot integration disabled (no value for `MITOL_HUBSPOT_API_PRIVATE_TOKEN`), create a new user.
- Follow readme instructions for setting up hubspot integration (I can give you access to hubspot if needed), you can use the same value as RC for `MITOL_HUBSPOT_API_PRIVATE_TOKEN`
- Run the following:
```
from users.models import User
from hubspot_sync.api import get_hubspot_id_for_object

user = User.objects.get(id=<your_new_user_id>)
contact = get_hubspot_id_for_object(user)
assert contact is not None
```

### DIFF
--- a/hubspot_sync/api.py
+++ b/hubspot_sync/api.py
@@ -7,7 +7,7 @@ from decimal import Decimal
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
-from django.core.validators import validate_email
+
 from hubspot.crm.objects import SimplePublicObject, SimplePublicObjectInput
 from mitol.common.utils.collections import replace_null_values
 from mitol.hubspot_api.api import (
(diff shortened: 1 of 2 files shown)


arslanashraf7 (approved):
Looks good 👍 

I was also able to test as per mentioned in PR description. Didn't have access to the portal though, but I used keys from RC to perform steps through Django shell.

PR #1411, opened 2023-03-20
Call sync_hubspot_application if an order is successful but application is not paid in full

### LINKED ISSUE #1024: Wiretransfer: after order is created, sync with hubspot
Currently, when wiretransfers are imported into the Bootcamps web app, orders are created but they are synced to Hubspot deals.

mbertrand:
Closes https://github.com/mitodl/hq/issues/1072
Closes #1024

#### What's this PR do?
When an order is completed, call the hubspot sync task directly if the order's application is not paid in full.

- Set the `MITOL_HUBSPOT_API_PRIVATE_TOKEN` to the same value as RC, and `MITOL_HUBSPOT_API_ID_PREFIX` to some unique value.
- Seed data if you need to (`manage.py seed_data`)
- Apply for a bootcamp
- Set the state of your application to `AWATING_PAYMENT`:
    ```
    manage.py set_application_state --app-id <application_id> --state AWAITING_PAYMENT 
    ```
 - There is an example wire CSV here: https://gist.github.com/noisecapella/9c093a86825b66f73df95afefad29bc6
 - Adjust the CSV to match the application bootcamp run start date and bootcamp title and the email to match the learner email.  Set the amount paid to half the amount owed.
 - Run `./manage.py import_wire_transfers ./example_wire_transfers.csv`
 - View the order history for the application. You should see the payment listed with payment type "Wire Transfer"
 - Look for the task `hubspot_sync.tasks.sync_deal_with_hubspot` in your celery container logs, and/or check the hubspot dashboard for the bootcampsdev sandbox account and find the deal, make sure the 'total price paid' is half the total owed.
 
 Screenshot 2023-03-21 at 8 52 52 AM

Currently, a signal calls the hubspot sync task whenever an application is saved.  When an order is completed, the application is saved only if it has been paid in full.  But we should sync the application to hubspot regardless because the serialized data will have changed (total paid amount, via a property).

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -42,7 +42,10 @@ from ecommerce.exceptions import (
     WireTransferImportException,
 )
 from ecommerce.models import Line, Order, WireTransferReceipt
-from hubspot_sync.task_helpers import sync_hubspot_application, sync_hubspot_application_from_order
+from hubspot_sync.task_helpers import (
+    sync_hubspot_application,
+    sync_hubspot_application_from_order,
+)
 from klasses.api import deactivate_run_enrollment
 from klasses.constants import ENROLL_CHANGE_STATUS_REFUNDED
 from klasses.models import BootcampRun
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -737,7 +737,7 @@ def test_parse_wire_transfer_csv_missing_header(tmp_path):
 )
 @pytest.mark.parametrize("paid_in_full", [True, False])
 def test_import_wire_transfer(
-    mocker, with_bootcamp_title, with_run_title, paid_in_full
+    settings, mocker, with_bootcamp_title, with_run_title, paid_in_full
 ):
     """import_wire_transfer should store a wire transfer in the database and create an order for it"""
     mock_hubspot_sync = mocker.patch("ecommerce.api.sync_hubspot_application")


pdpinch:
Will this fix #1024 as well?

mbertrand:
It should, sounds like the same problem

ecommerce/views.py
```diff
@@ -133,12 +133,8 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
         if decision != CYBERSOURCE_DECISION_ACCEPT:
             handle_rejected_order(order=order, decision=decision)
         else:
-            # import pdb; pdb.set_trace()
             complete_successful_order(order)
 
-        # Sync order data with hubspot
-        sync_hubspot_application_from_order(order)
```
mbertrand, line 140: Hubspot sync should now happen within `handle_rejected_order` or `complete_successful_order`

pdpinch:
Why does the wiretransfer need special treatment, when all the credit card orders get synced even if they don't pay in full?

mbertrand:
@pdpinch the credit card orders hit [a different section of the code here](https://github.com/mitodl/bootcamp-ecommerce/blob/0e02fb25db72639b5a5f5fca3f80e7714462398d/ecommerce/views.py#L101-L143).

PR #1419, opened 2023-05-23
Handle hubspot syncing for contacts w/multiple conflicting emails

mbertrand:
#### What are the relevant tickets?
Closes #1418 

#### What's this PR do?
- Uses the new version of mitol.hubspot_api which can handle conflicts caused by a Hubspot contact having 2 emails associated with different bootcamps users
- Modifies the batch create/update tasks to try syncing contacts individually (to hit code above) if any API exceptions occur when running batch sync operations.

#### How should this be manually tested?
- Ask for access to the bootcampsdev sandbox account for hubspot if you don't already have it.
- Copy the `HUBSPOT_*` settings values from RC to your .env file

Part 1:
- Create a new user in bootcamps, something like `fake_user+1@gmail.com`
- Find the contact in hubspot (may take a minute or so to show up) and add a secondary email to that contact: `fake_user+2@gmail.com`

Screenshot 2023-05-11 at 10 30 32 AM

- Register another new user with that same email - `fake_user+2@gmail.com`
- Check back on hubspot, that secondary email should be gone from the contact, and there should be a new contact with email `fake_user+2@gmail.com`

Part 2:
- Comment out your hubspot .env settings and restart your containers
- Create another new user, `fake_user+3@gmail.com`
- In hubspot, add that email to one of the existing contacts as a secondary email
- Uncomment your hubspot .env settings and restart your containers
- Run `manage.py sync_db_to_hubspot --users create`.  It should complete successfully
- Check hubspot again, the secondary email from the contact should be gone and assigned to a new contact.

PR #1423, opened 2023-05-25
Fix interaction between hubspot contact creation form and hubspot api contact sync

### LINKED ISSUE #1421: Hubspot user form call conflicts with hubspot api call under certain conditions
## Steps to Reproduce
- Register a new user in bootcamps with email fake_user1@gmail.com
- A hubspot contact is created with that email
- Manually add a second email, fake_user2@gmail.com, to that same hubspot contact
- Register another new user in bootcamps with email fake_user2@gmail.com

## Expected Behavior
The second email is removed from the 1st hubspot contact and added as a new contact.
The first bootcamp user is associated w/the first hubspot contact, the second bootcamp user is associated w/the second contact.

## Current Behavior
The first email is removed from the 1st hubspot contact, the 2nd email becomes the primary email, and no new contact is created.
The first bootcamp user is no longer associated with a hubspot contact, the second user is associated with the 1st contact.

## Possible Solution
The form seems to no longer be necessary.  So as a first step, remove the value for it, and remove the error log statement that is triggered if it is null.  
It may also be a good idea to sync the contact w/hubspot after every login as part of the auth pipeline to ensure everything is up to date.
A separate tech debt issue will be created to remove the form-related code altogether after we are confident that it is no longer necessary.

mbertrand:
Closes #1421 

#### What's this PR do?
Makes HUBSPOT_CREATE_USER_FORM_ID optional (no error logged if null).
Makes an API call to sync the user to hubspot on login, which should ensure it's up to date regardless of whether or not the form above is being used.

- Set `HUBSPOT_CREATE_USER_FORM_ID` to the same value as RC in your .env file
- Follow the instructions for https://github.com/mitodl/bootcamp-ecommerce/pull/1419
- Log out and back in a couple times.
- There should not be any errors in the log.
- Repeat the above but with `HUBSPOT_CREATE_USER_FORM_ID` removed from your .env file

Related: https://github.com/mitodl/salt-ops/pull/1567

### DIFF
--- a/authentication/pipeline/user.py
+++ b/authentication/pipeline/user.py
@@ -281,3 +281,11 @@ def send_user_to_hubspot(request, **kwargs):
     requests.post(url=url, data=data, headers=headers)
 
     return {}
+
+def sync_user_profile_to_hubspot(
+    backend, user, response, is_new, *args, **kwargs
+):  # pylint: disable=unused-argument
+    """
+    Sync the user's latest profile data with hubspot on login
+    """
+    sync_hubspot_user(user)
--- a/authentication/pipeline/user.py
+++ b/authentication/pipeline/user.py
@@ -3,28 +3,26 @@ import json
 import logging
 
 import requests
-from social_core.backends.email import EmailAuth
-from social_core.exceptions import AuthException
-from social_core.pipeline.partial import partial
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


collinpreston (approved):
Tested and no errors.  Looks good

Issue #1440, opened 2023-10-20
Use `poetry` for Python dependency management


arslanashraf7:
Use poetry instead of relying on separate requirements.in, requirements.txt, and test_requirements.txt files

#### Designs and Mockups

See:
 - Open-discussions:
   - https://github.com/mitodl/open-discussions/pull/4151
   - https://github.com/mitodl/open-discussions/pull/4178
   - https://github.com/mitodl/open-discussions/pull/4179
 - OVS:
   - https://github.com/mitodl/odl-video-service/pull/1060
   - https://github.com/mitodl/odl-video-service/pull/1064
- xPRO
    - https://github.com/mitodl/mitxpro/pull/2715
- MicroMasters
    - https://github.com/mitodl/micromasters/pull/5345   

Related to https://github.com/mitodl/ol-infrastructure/issues/1578

arslanashraf7:
Done via #1456

PR #1456, opened 2023-11-01
feat: use poetry dependency management

arslanashraf7:
https://github.com/mitodl/bootcamp-ecommerce/issues/1440

#### What's this PR do?
- Adds support for poetry dependency management
- updates GitHub actions to use poetry
- updates docker to use poetry

- The build should succeed
- The tests should pass
- The `pyproject.toml` dependencies should be accurate as per `^` (For main dependencies only minor and patch versions should be upgraded in this PR)

**Try two build scenarios:**
- Run this PR on an existing build from master, ideally `docker-compose build --no-cache` should work
- Run a fresh instance `docker-compose down`, `docker-compose build --no-cache`, `docker-compose up`

**Try the Functionality test scenarios**
- Try creating new account/Signin
- Try setting up bootcamp in CMS and make sure it displays and works the same as before
- Try running commands like `pytest, pylint` and they should work
- Try running one or more management commands
- It would be a good idea to test If you have Hubspot setup locally


pyproject.toml
```diff
@@ -20,3 +20,96 @@ include_trailing_comma = true
...
+Pillow = "^9.3.0"
+user-util = "^0.1.5"
+
+
+[tool.poetry.group.dev.dependencies]
+astroid = "^2.4.2"
+black = "^22.10.0"
+bpython = "^0.19"
+django-debug-toolbar = "^3.2.4"
+hypothesis = "~4.23.9"
```
arslanashraf7, line 91: Pinning this to update only patch because the minor version upgrade breaks tests with StateMachine

pyproject.toml
```diff
@@ -20,3 +20,96 @@ include_trailing_comma = true
...
+ipdb = "^0.13.2"
+isort = "^4.3.21"
+factory_boy = "^3.2.0"
+faker = "^8.8.2"
+nplusone = "^1.0.0"
+pdbpp = "^0.10.2"
+pip-tools = "^7.1.0"
+pylint = "~2.5.3"
```
arslanashraf7, line 99: Using only patch version upgrade because the minor version upgrades requires changes in files as many as ~80 and doing that would increase the testing scope of this PR huge.

pyproject.toml
```diff
@@ -20,3 +20,96 @@ include_trailing_comma = true
...
+pylint-django = "~2.1.0"
```
arslanashraf7, line 100: Using only patch version upgrade because the minor version upgrades requires changes in files as many as ~80 and doing that would increase the testing scope of this PR huge.

feoh (approved):

mbertrand (approved):
LGTM 👍

PR #1491, opened 2024-03-12
feat(import_wire_transfers)!: use bootcamp run id in import_wire_transfers

asadali145:
Closes https://github.com/mitodl/hq/issues/3632

#### What's this PR do?
Updates the `import_wire_transfers` management command to support the Bootcamp run ID instead of the Name and start date in the CSV.

Name and start date can be present in the CSV but we won't use them. We will just use the Bootcamp Run ID to import the wire transfers.

- Create a sample CSV with these fields. `Id,Learner Email,Amount,Bootcamp Run ID,Bootcamp Start Date,Bootcamp Name`
- Add some sample records.
- Make sure to add the Bootcamp Run ID
- Use `import_wire_transfers` to import the CSV
- It should be successful.


arslanashraf7 (commented):
The code changes look good and it works fine. However, I have some questions:

1. I wonder why we didn't use the bootcamp run Id or something more robust instead of time and title when we implemented that change. I was looking at the history of the code and I couldn't find anything specific. Maybe providing `bootcamp_run_id` in the sheet wasn't possible?
2. We should highlight it properly to the finance team that we will be expecting the `Bootcamp Run ID` column in the sheet moving onwards because without that the code won't work.
3. Should we hard fail it on the `bootcamp_run_id` moving onwards? Or we should keep the title & start_date check as a fallback when `bootcamp_run_id` isn't provided?

@pdpinch

pdpinch:
> I wonder why we didn't use the bootcamp run Id or something more robust instead of time and title when we implemented that change. I was looking at the history of the code and I couldn't find anything specific. Maybe providing bootcamp_run_id in the sheet wasn't possible?

I'm pretty sure the original implementation was written before bootcamp run ids existed. 

> We should highlight it properly to the finance team that we will be expecting the Bootcamp Run ID column in the sheet moving onwards because without that the code won't work.

@cachob please communicate this change to the Bootcamps and finance teams about this when the change is released. 

> Should we hard fail it on the bootcamp_run_id moving onwards? Or we should keep the title & start_date check as a fallback when bootcamp_run_id isn't provided?

I think we should hard fail. There's only 5-10 of these per year and the overheard of making the name and date approach work makes it much harder to maintain.

arslanashraf7 (approved):
Look good 👍 Based on testing and https://github.com/mitodl/bootcamp-ecommerce/pull/1491#issuecomment-2005005741.

cachob:
Sure thing - I will reach out to the respective teams

PR #1507, opened 2024-04-05
add readable_id to bootcamp model and related changes

rachellougee:
https://github.com/mitodl/hq/issues/3905

#### What's this PR do?

- Adding `readable_id` to Bootcamp Model
- Updating Bootcamp serializer,  factories and seed_data to include this new field
~~- Adding backpopulating migration to populate `readable_id` with existing bootcamp run IDs~~
   ~~- If the associated bootcamp run ID is `bootcamp-v1:public+IE-f2f+R2`, then `readable_id` is populated with `bootcamp-v1:public+IE-f2f`~~

- Run `docker-compose run --rm web ./manage.py migrate klasses 0029` to add `readable_id`  to Bootcamp model
- Run `docker-compose run --rm web ./manage.py seed_data` to populate data
~~- Go to admin/klasses/bootcamp/ and blank out Readable id from the above seeded data~~
~~-- Run  `docker-compose run --rm web ./manage.py migrate klasses 0030` to backpopulate `readable_id` for these seed data~~-


rachellougee:
@pdpinch I added the backfill for the existing bootcamps if their associated runs have bootcamp_run_id. The readable ID is back-populated as  bootcamp-v1:TYPE+TOPIC-FORMAT. There are 25 bootcamps on production can be back-populated. But Let me know if you prefer to manually populate them.

pdpinch:
> I added the backfill for the existing bootcamps if their associated runs have bootcamp_run_id. The readable ID is back-populated as bootcamp-v1:TYPE+TOPIC-FORMAT. There are 25 bootcamps on production can be back-populated. But Let me know if you prefer to manually populate them.

Sounds good to me.

klasses/models.py
```diff
@@ -41,6 +41,14 @@ class Bootcamp(models.Model):
 
     title = models.TextField()
     legacy = models.BooleanField(default=False)
+    readable_id = models.CharField(
+        null=True,
+        blank=True,
```
arslanashraf7, line 46: readable_id in other apps is not nullable. Is there a reason we are not forcing it?

One reason I think of is that we don't have data for all of these now. If Yes, I think we should do the following:
1. Create migration for this as is nullable (This is already done in this PR 0029)
2. Create another migration making this non-nullable 
3. This can very well be the migration 0030 that you created but it will now do two tasks in order a) Backfill with appropriate values b) Alter field and make non-nullable ([This is an example from xPRO](https://github.com/mitodl/mitxpro/pull/2786/files#diff-75e20b31ad6576874eb9bee3b7edec5a3a1ed36acc86e408e56093016bd13129))

arslanashraf7: If you would like to do this in a followup PR we would end up with 3 migrations instead of doing it in 2, But I don't have issues with that as well. Just let me know.

arslanashraf7 (changes requested):
I haven't tested the PR yet. Just did the code review. I left a couple of comments.

klasses/factories.py
```diff
@@ -19,6 +19,13 @@ class BootcampFactory(DjangoModelFactory):
     """Factory for Bootcamp"""
 
     title = FuzzyText(prefix="Bootcamp ")
+    readable_id = LazyAttribute(
+        lambda bootcamp: "bootcamp-v1:{type}+{topic}-{format}".format(
+            type=random.choice(["public", "private"]),
+            topic=FAKE.slug(),
+            format=random.choice(["ol", "f2f"]),
```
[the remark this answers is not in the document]
rachellougee: Sorry, which one should to convert to f string? the string "ol", "f2f"?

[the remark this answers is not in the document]
arslanashraf7: Yes

klasses/models.py
arslanashraf7: Yeah, That sounds good.

klasses/factories.py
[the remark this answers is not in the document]
rachellougee: I made the change. It took longer than I anticipated to convert it to f string with lambda.

klasses/models.py
rachellougee: I reverted 0030 migration, the backfill and make it required field will be addressed in https://github.com/mitodl/hq/issues/3975

collinpreston (approved):

PR #1553, opened 2024-05-23
fix: missing migration and migrations detection scripts

arslanashraf7:
None, But it was noticed during testing. It's a followup on https://github.com/mitodl/bootcamp-ecommerce/pull/1507, The migration was missing from that PR.

#### What's this PR do?
- Adds a missing migration
- Fixes why the scripts didn't detect that
- Ports over the xPRO migration detection scripts because they don't use a temporary file.

- Check the tests are passing
- If you want to do more testing, Try testing the auto & missing migration scripts locally

(Optional)

This was left in https://github.com/mitodl/bootcamp-ecommerce/pull/1507


klasses/migrations/0030_alter_bootcamp_readable_id.py
```diff
@@ -0,0 +1,24 @@
...
+        migrations.AlterField(
+            model_name="bootcamp",
+            name="readable_id",
+            field=models.CharField(
+                blank=True,
+                help_text="The unique string to identify this bootcamp. It can be of the form 'bootcamp-v1:TYPE+TOPIC-FORMAT'(example: bootcamp-v1:public+IE-f2f)",
+                max_length=255,
+                null=True,
+                unique=True,
+            ),
```
rachellougee, line 22: Maybe I am missing something, but could you help me understand why this AlterField migration is needed?
The field seems the same as https://github.com/mitodl/bootcamp-ecommerce/blob/master/klasses/migrations/0029_bootcamp_readable_id.py as readable_id is not required field until we do the backfil.
```
   operations = [
        migrations.AddField(
            model_name="bootcamp",
            name="readable_id",
            field=models.CharField(
                blank=True,
                max_length=255,
                null=True,
                unique=True,
                help_text="The unique string to identify this bootcamp. It can be of the form "
                "'bootcamp-v1:TYPE+TOPIC-FORMAT' (example: bootcamp-v1:public+IE-f2f)",
            ),
        ),
    ]
``

arslanashraf7: I originally noticed this when running `makemigrations` was generating new migration everytime on master when it shouldn't.

That's probably because of the reformatting that might have been applied when you merged the PR. The only difference between 0029 & 0030 is the help_text. One has extra `""` probably because of the line break during the formatting.

We might have two ways to solve this extra migration issue:

1. The new migration that I created in this PR because the gets generated everytime on master because the migrations are out of sync with models?
2. We reformat the `help_text` so that it matches the migration in 0029.

rachellougee (approved):
LGTM. Verified that No changes detected in app 'klasses' when running `makemigrations` locally.

PR #1575, opened 2024-07-10
fix: invalid certificate uuid should raise 404

Anas12091101:
### What are the relevant tickets?

Fixes https://github.com/mitodl/hq/issues/4650

### Description (What does it do?)

Using an invalid UUID to get a certificate results in a 500 server error. It should return 404.

Steps to reproduce:

- Checkout master branch
- fetch `/certificate/f5ab714b-a7ee-498f-bbec-/`
- It will raise a validation error

### How can this be tested?

- Create certificate pages in CMS
- Create certificate in the Django Admin for a user
- Fetch the certificate at `/certificate/<uuid>/`. You should see the valid certificates.
- Now fetch the certificate with invalid UUID i.e. `f5ab714b-a7ee-498f-bbec-`. You should see a 404 page.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -8,6 +8,7 @@ import logging
 from urllib.parse import urljoin
 
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.templatetags.static import static
 from django.db import models
 from django.http.response import Http404
@@ -937,7 +938,7 @@ class CertificateIndexPage(RoutablePageMixin, Page):
             and not parent.get_children().type(cls).exists()
         )
 
-    @route(r"^([A-Fa-f0-9-]+)/?$")
+    @route(r"^([A-Fa-f0-9-]{36})/?$")
     def bootcamp_certificate(
         self, request, uuid, *args, **kwargs
     ):  # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


cms/models.py
```diff
@@ -953,7 +954,7 @@ def bootcamp_certificate(
             certificate = BootcampRunCertificate.objects.get(
                 uuid=uuid, is_revoked=False
             )
-        except BootcampRunCertificate.DoesNotExist:
+        except (BootcampRunCertificate.DoesNotExist, ValidationError):
```
arslanashraf7, line 957: Do we need this? When and how is this Validation error going to be thrown? Even if this error is thrown we should let the program break instead of a 404.

cms/models_test.py
```diff
@@ -372,6 +372,18 @@ def test_certificate_index_page(rf):
...
+        certifcate_index_page.bootcamp_certificate(request, "00000000-0000-")
+
+    with pytest.raises(Http404):
+        certifcate_index_page.bootcamp_certificate(request, "")
+
+    with pytest.raises(Http404):
+        certifcate_index_page.bootcamp_certificate(
+            request, "00000000-0000-0000-0000-000000000000000"
+        )
+
```
arslanashraf7, line 386: The test can run on broader data set. Also please use parametrized tests in cases where the code is redundant. You can take a look at https://github.com/mitodl/mitxpro/pull/2990 for reference.

arslanashraf7 (changes requested):

arslanashraf7 (approved):

PR #1598, opened 2024-09-18
fix: pin minimist to solve dependabot security issues

Anas12091101:
### What are the relevant tickets?
Solve https://github.com/mitodl/bootcamp-ecommerce/security/dependabot/100

### Description (What does it do?)
Pinning minimist in resolutions to fix the critical security issue

### How can this be tested?
Make sure everything's is working

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "keycode": "^2.2.0",
     "lodash": "^4.17.21",
     "mini-css-extract-plugin": "^1.1.2",
-    "minimist": "^1.2.6",
+    "minimist": "1.2.8",
     "mocha": "^8.0.0",
     "moment": "^2.29.4",
     "mutationobserver-shim": "^0.3.5",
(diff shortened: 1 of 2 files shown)


asadali145 (changes requested):
yarn.lock contains changes that are not related to this PR. Also, please rebase to resolve conflicts.

package.json
```diff
@@ -125,6 +125,7 @@
     "yup": "^0.32.0"
   },
   "resolutions": {
+    "minimist": "1.2.8",
```
asadali145, line 128: Instead of adding a resolution, we can simply update the minimist version in the dependencies above to `1.2.8`. Is there any issue with that? We only go for the resolutions when we face dependency conflicts and there is no other way but this one seems simpler and worked for me locally/

asadali145 (changes requested):

asadali145 (approved):

